PR #17, opened 2014-12-30
Loop fix

### LINKED ISSUE #4: endless loop in rrule.py
There is a bug in rrule.py (dateutil 2.3). The `while True` loop on line 659 does not always terminate:
```python
elif freq == HOURLY:
    if filtered:
        # Jump to one iteration before next day
        hour += ((23-hour)//interval)*interval
    while True:
        hour += interval
        div, mod = divmod(hour, 24)
        if div:
            hour = mod
            day += div
            fixday = True
        if not byhour or hour in byhour:
            break
```
Here's an example:
If the initial value of `hour` is an odd number, say `23`, and the `byhour` tuple is, e.g., `(0, 6, 12, 18)`, then we have an endless loop if `interval` is even because `hour` will always be odd and, hence, `hour in byhour` is never `True`!

Because dateutil is used by matplotlib, this causes some plotting routines to freeze.

pganssle:
Converted most of the `while True` loops to explicit `for` loops looping over a length after which the pattern repeats, then raise an error if the pattern hasn't broken by then. This should produce a desirable behavior that fixes #4 (though I think that there are certain endless-loop conditions which I'm currently only detecting in the iterator, not the constructor, because checking them in the constructor would involve a much more complicated logic because they involve interactions between the `byhour`, `byminute` and/or `bysecond` rules - either way I think they're likely exceedingly rare in practice).

I did a little refactoring in the looping logic, but not much since I think a lot of it won't end up being used in `rrule2` anyway. I'm a bit curious as to why we're explicitly handling the low-level operations of incrementing the day rather than incrementing a `datetime.datetime` object using `datetime.timedelta` objects - is this for the sake of performance or something? I can't imagine it would add much overhead, and it would be much easier to maintain.

I also converted the `by_` tuples to sets, because it doesn't seem that their order is ever used, but there are frequent calls to `in`, which will be considerably sped up by the use of sets. While I was doing that, I did some refactoring in the constructor which will make it easier to switch back if this change proves undesirable.

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -703,7 +703,7 @@ class rrule(rrulebase):
 
                 valid = False
                 rep_rate = (24*60)
-                for j in range(rep_rate / gcd(interval, rep_rate)):
+                for j in range(rep_rate // gcd(interval, rep_rate)):
                     if byminute:
                         nhours, minute = \
                             self.__mod_distance(value=minute,
@@ -735,7 +735,7 @@ class rrule(rrulebase):
 
                 rep_rate = (24*3600)
                 valid = False
-                for j in range(0, rep_rate / gcd(interval, rep_rate)):
+                for j in range(0, rep_rate // gcd(interval, rep_rate)):
                     if bysecond:
                         nminutes, second = \
                             self.__mod_distance(value=second,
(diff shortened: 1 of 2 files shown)


jarondl:
That's one impressive contribution. Merged,
Thank you very much

pganssle:
@jarondl No problem. Let me know if there's any way I can help with `rrule2` or if there are any improvements I can help with in general - I might as well work on that stuff before I forget all the stuff I learned to fix this problem!

Issue #18, opened 2014-12-31
Provide extended strftime support


pganssle:
In Python 2.x, [`strftime()` is limited to years > 1900](https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior). In Python 3.3, it's limited to years > 1000. My understanding is that at least in the 2.x branch, this behavior is a [WONTFIX](http://bugs.python.org/issue1777412). Is it worth it to provide an alternative version of strftime that works for older versions?

(I bring this up because I've been working on matplotlib's [date-handling module](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/dates.py) and this is one of the things where it's actually doing date calculations that aren't outsourced to library calls.)

jarondl:
@pganssle , sorry for ignoring this for so long.
Yes, we can add this functionality to dateutil if you want.

pganssle:
OK, I can look into the best way to do that. Matplotlib has some way to do it by shifting the date, but I'm a bit worried about the generality of that approach. I don't see a good fit for it in the current set of modules, but I don't know if really deserves to be in its own module. Should it be mapped to a top-level function in the dateutil module itself, or should it be part of a module that might expand later?

mariocj89:
Python3.9 seems to work with any year:

```
>>> datetime.datetime(1,2,3).strftime("%Y/%m/%d")
'1/02/03'
```

Closing given this issue has been open since 2014 and there is not support in Cpython.

PR #20, opened 2015-01-12
Add iso8601 formatter

coldeasy:
Added an `isoformat` function to allow formatting dates with a zulu extension rather than `+00:00` for UTC datetimes.

### DIFF
--- a/dateutil/formatter.py
+++ b/dateutil/formatter.py
@@ -1,0 +1,10 @@
+def isoformat(dt, utc_zulu=True):
+    """
+    Formats the datetime using datetime.isoformat()
+
+    If `utc_zulu` is `True` and `dt` has UTC timezone then the suffix will be
+    `'Z'` instead of `'+00:00'`. Otherwise the output is unchanged.
+    """
+    if dt.tzinfo is not None and dt.tzinfo.tzname(dt) == 'UTC' and utc_zulu:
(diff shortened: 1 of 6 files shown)


jarondl:
Hi @coldeasy 
This is a well prepared PR, but I don't understand the premise.
The solution is a one-liner:

    dt.isoformat().replace('+00:00','Z')

So why would a user import dateutil for that?

coldeasy:
@jarondl The main reason I am adding it here is to remove the need for this one-liner or indeed any 'wrapper' module that will does this due to the lack of extensibility of stdlib's `isoformat`.
Perhaps it's not worth it for such a small change right now but if there turns out to be other formatting features added to this library in future you could consider adding this.

pganssle:
I'm tempted to close this because it's sat idle for over a year and I don't really want to introduce a whole new module just to format iso8601 strings, but if we do decide to go ahead with #18, I can see including this as an add-on module to that.

Either way, it'll need a rebase, but probably best to wait until it's more likely to be included. I don't foresee implementing #18 in 2.6.0.

mariocj89:
Closing per https://github.com/dateutil/dateutil/pull/20#issuecomment-193942192. This is already substantially out of date and does not seem worth adding right now given this is a replacement for a one-liner.

Issue #21, opened 2015-01-21
Garbage input for parse returns a datetime, not an TypeError


tpievila:
A user reported a bug:

When "a" is provided as input str to the parser, TypeError exception is
expected. Instead, the parser returns a datetime object.

    >>> import dateutil.parser as d_parser
    >>> d_parser.parse*('a')*
    datetime.datetime(2014, 11, 10, 0, 0)

pganssle:
I'm nearly 100% sure this shouldn't throw a `TypeError`, if anything it should throw a `ValueError`, since parsers are supposed to parse strings, and this is just a malformed string. That said, the reason this is happening is because `a` is interpreted as "AM", and "AM" is basically ignored when parsing the hours unless the hour specified is 12 (because 12am = 0:00 in 24-hour time). Since an empty string `d_parser.parse('')` returns midnight of the current day, the string `'a'` is treated like `''` with any specified numbers treated as AM.

You'll note that `d_parser.parse('p')` actually *does* throw an error, but not deliberately, it's just thrown because the hour is specified as `None` and it tries to add 12 to the hour. This doesn't seem like a high priority bug, but probably both of these issues can be fixed easily by throwing a `ValueError` any time AM or PM is specified without an hour being specified, since that's a non-sensical situation.

Another issue is that the parser is not checking for non-sensical inputs such as `dparser.parse('14am')`, which is parsed as 2 PM. Again, this is non-sensical because AM and PM are specifiers for a 12-hour clock, so all numbers should be in the range `1 <= hour <= 12`.

Either way, the parser should be explicit about how these edge cases are handled in the documentation. I can prepare a PR for these changes.

PR #22, opened 2015-01-22
Add explicit error checking along with AM/PM flag.

pganssle:
Adds explicit error checking when AM or PM are specified. `dateutil.parser.parse()` will now raise a `ValueError` exception if either AM or PM is specified without specifying an hour, or when the hour specified is outside the range `1 <= hour <= 12`, since AM/PM indicates a 12-hour clock. This resolves issue #21.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -585,7 +585,7 @@ class parser(object):
                         raise ValueError('No hour specified with AM or PM flag.')
 
                     # If AM/PM is found, it's a 12 hour clock, so raise an error for invalid range
-                    if not 1 <= res.hour <= 12:
+                    if not 0 <= res.hour <= 12:
                         raise ValueError('Invalid hour specified for 12-hour clock.')
 
                     if value == 1 and res.hour < 12:
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -580,6 +580,14 @@ class parser(object):
                 # Check am/pm
                 value = info.ampm(l[i])
                 if value is not None:
+                    # If AM/PM is found and hour is not, raise a ValueError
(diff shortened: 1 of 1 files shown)


pganssle:
So I guess this is failing because of [`testRandomFormat24`](https://github.com/dateutil/dateutil/blob/master/dateutil/test/test.py#L3629). @jarondl - Can you comment on how I should resolve that? I had assumed that it was an oversight that the AM/PM flag allowed invalid values for the hours (e.g. null values or values < 1 and > 12), but `testRandomFormat24` explicitly tests to make sure that `'0:00 PM' is parsed as 12:00, so is this a desired flexibility?

If it's a desired flexibility, is it also desired that, when no hour is specified with the AM flag, midnight is assumed and when no hour is specified with the PM flag, noon is assumed? Please advise and I'll change the PR either to add a a test to the testing environment or to more explicitly handle the edge cases.

jarondl:
@pganssle , thank you for your continous contributions.
According to [blame `testRandomFormat24`](https://github.com/dateutil/dateutil/blame/master/dateutil/test/test.py#L3629), it is 11 years old, but it is not set in stone.
However, according to an obscure japanase reference in wikipedia, http://en.wikipedia.org/wiki/12-hour_clock#cite_ref-14 , `00 a.m` is a valid notation.
So I'd say that  `0 A.M` is different from `AM` with no hour at all. Maybe simply change the condition to be `not 0 <= res.hour <= 12`?

pganssle:
Sounds good. Thanks.

pganssle:
Not sure why the travis-CI build is failing. Something about tzinfo, but I didn't make any changes to that section of the code...

Issue #25, opened 2015-01-30
"Feb 2011" parsing fails but Jan, Mar-Dec works


umrashrf:
```python
>>> from dateutil import parser
>>> parser.parse('Jan 2011', fuzzy=True)
datetime.datetime(2011, 1, 30, 0, 0)
>>> parser.parse('Feb 2011', fuzzy=True)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/umair/dev/venvs/nathanartz/local/lib/python2.7/site-packages/dateutil/parser.py", line 743, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/home/umair/dev/venvs/nathanartz/local/lib/python2.7/site-packages/dateutil/parser.py", line 310, in parse
    ret = default.replace(**repl)
ValueError: day is out of range for month
>>> parser.parse('Dec 2011', fuzzy=True)
datetime.datetime(2011, 12, 30, 0, 0)
```

Allactaga:
Actually it is not only problem with February. I believe tomorrow (Jan 31st) would bring even more cases like this: 
```python
>>> from datetime import date
>>> from dateutil import parser
>>> parser.parse("Apr 2011", default=date(2015, 1, 31))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/dateutil/parser.py", line 697, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/dateutil/parser.py", line 310, in parse
    ret = default.replace(**repl)
ValueError: day is out of range for month
```

This is happening because case that there could be less days in the replacing month, than there are days in the default date, should be checked before [this line](https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py#L310).

umrashrf:
Right. If this is coming from default date then I think day should come from **calendar** depending on month.

pganssle:
Presumably the correct solution to this is to interpret `"Feb 2011"` as the first day of February. By extension, is it reasonable to say that `"2011"` should be interpreted as 0:00:00 January 1st, 2011? I'm getting the impression that the parser falls back to modifying the current date rather than falling back to the first instance of a date that can be described by the parsed information.

[...]

PR #30, opened 2015-02-05
Add smart defaults and fallback options to parser.

pganssle:
Addresses Issue #25, but does not change the default behavior of the parser. This adds a `smart_defaults` option to the parser, along with a corresponding `date_in_future` option.

I have also added the option `fallback_on_invalid_day`, which silently corrects the date to the end of the month if a date outside of the valid range is provided. Thus you can use the default automatic-replacement behavior while still allowing no error to be thrown in the event of a conflict in month lengths.

@jarondl, @Allactaga @umrashrf Any issues with this approach?

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -3,6 +3,32 @@
 This module offers a generic date/time string parser which is able to parse
 most known formats to represent a date and/or time.
 
+This module attempts to be forgiving with regards to unlikely input formats,
+returning a datetime object even for dates which are ambiguous. If an element of
+a date/time stamp is omitted, the following rules are applied:
+- If AM or PM is left unspecified, a 24-hour clock is assumed, however, an hour
+  on a 12-hour clock (`0 <= hour <= 12`) *must* be specified if AM or PM is
+  specified.
+- If a time zone is omitted, it is assumed to be UTC.
+
+If any other elements are missing, they are taken from the `datetime.datetime`
+object passed to the parameter `default`. If this results in a day number
+exceeding the valid number of days per month, one can fall back to the last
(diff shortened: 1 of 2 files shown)


pganssle:
By the way, I think that the right thing to do here is to eventually change the default behavior to `smart_defaults=True`.

My instinct says there should be a version with a `FutureWarning` in it if `smart_defaults` isn't specified, but I'm not really sure that's necessary - I don't know how many people are actually relying on the default date specification behavior, and since they always have the option of going back to the old behavior by explicitly specifying `smart_defaults=False`, it's probably better not to have an intermediate version where annoying messages are generated.

I think the middle ground is to have one or two versions where `smart_defaults` defaults to `False`, and the documentation explicitly mentions that this default will be changed in the future, then switch it.

pganssle:
OK, I've given this some more thought, and I don't think this should be merged as is because of the way I have `fallback_on_invalid_day`'s deafults set. Right now, without the smart parser turned on, "Feb 30, 2014" and "Feb 2014" with a default date of "January 30th, 2014" are dealt with the same way, even though the first one is actually an invalid date and the second one should arguably fall back to the end of the month (as I believe `relativedelta` does).

The right solution, I think, is to make `fallback_on_invalid_day` a trinary - if it's unspecified or `None`, then the fallback behavior would be in effect only if the reason for the invalid parse is that the day is actually missing.

I won't worry about invalid dates created because of missing years - for example "Feb 29" with a default date in 2014 probably *should* throw an error by default, since there's reason to believe that the parser is giving you a date other than the one that was specified.

Ramblurr:
Since this issue is closed, I opened a new one #126

Issue #34, opened 2015-02-10
2.4.0 rrule produces wrong result


twiecki:
```python
In [1]: import pytz

In [3]: from datetime import datetime

In [4]: from dateutil import rrule

In [5]: mlk_day = rrule.rrule(
   ...:         rrule.MONTHLY,
   ...:         bymonth=1,
   ...:         byweekday=(rrule.MO(+3)),
   ...:         cache=True,
   ...:         dtstart=datetime(1998, 1, 1, tzinfo=pytz.utc))

In [6]: mlk_day.between(datetime(2000, 1, 1, tzinfo=pytz.utc), datetime(2000, 1, 30, tzinfo=pytz.utc))
Out[6]: 
[datetime.datetime(2000, 1, 3, 0, 0, tzinfo=<UTC>),
 datetime.datetime(2000, 1, 10, 0, 0, tzinfo=<UTC>),
 datetime.datetime(2000, 1, 17, 0, 0, tzinfo=<UTC>),
 datetime.datetime(2000, 1, 24, 0, 0, tzinfo=<UTC>)]
```

The same code with dateutil 2.3.0 correctly produces:
```python
In [9]: mlk_day.between(datetime(2000, 1, 1, tzinfo=pytz.utc), datetime(2000, 1, 30, tzinfo=pytz.utc)) 
Out[9]: [datetime.datetime(2000, 1, 17, 0, 0, tzinfo=<UTC>)]
```

pganssle:
@twiecki Thanks for the catch! The issue came because in 2.4.0 I had refactored the rrule constructor and made an error in how the byweekday parameter is handled. I guess we didn't have a test for this condition, so it didn't get caught by the test suite. PR #35 fixes this issue and adds a test to the test suite.

twiecki:
Thanks for the quick fix! Any chances of a `2.4.1` release after this made it into master? Would help our packaging!

pganssle:
@twiecki No problem! As for a 2.4.1 release, @jarondl handles all that, but I doubt it will be a problem. There's a new update to the timezone database anyway, so it's probably time to update the tz file. Is there a specific time-frame from your release that it would be best if we hit?

jarondl:
Hi @twiecki and @pganssle, a 2.4.1 will be released soon

twiecki:
Friendly nudge. Lots of our users are bumping into this issue.

pganssle:
[...]

PR #35, opened 2015-02-10
RRULE fix for 2.4.0 ByNWeekday issue.

pganssle:
Fixes Issue #34.


jarondl:
Good work @pganssle .

If we were free from API obligations, we could have removed this ambiguity by accepting sequences only..

PR #36, opened 2015-02-14
Doc and test tweaks

pganssle:
Documentation currently uses `parse()` for almost all the rrule tests, and a bunch of others. I think it's better to explicitly specify `datetime` objects, since any problems with the parser will start causing the rrule tests to fail as well, causing the tests to be less well-separated. It also probably speeds up the test process a bit.


pganssle:
Oops, for some reason the readme isn't rendering right - I thought I checked that in a gist...

PR #38, opened 2015-02-17
Add xafter method to rrule

pganssle:
Adds an `xafter` method to `rrule`, which is a generator function generating all recurrences after a specified date.

An equivalent `xbefore` method is not simple to implement, as it is not possible to iterate backwards through an `rrule`. That may be a nice enhancement for `rrule2` if possible.

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -209,7 +209,48 @@ class rrulebase(object):
                     return i
         return None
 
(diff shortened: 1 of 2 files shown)


pganssle:
I'm not seeing any objection to this - doesn't really turn out to solve the other person's problem, but I don't think it's a bad idea anyway, so I'm going to merge.

Issue #27, opened 2015-02-03
Provide version of Olson's database as a constant


skontar:
The documentation claims that:

> Internal up-to-date world timezone information based on Olson's database. 

However, the current used version of the database cannot be found from installed package. Please provide constant with stored version of used database.

pganssle:
@skontar Have you run the updatetzinfo.py?

> To get (and update) the database, run the updatezinfo.py script.

I believe this is done because it is likely better to allow users to update the timezone database independent of updates to the library itself.

skontar:
@pganssle I am not really interested in updating the database myself. I would just like to know version number (e.g. 2014j) of the already included database which was used when the library was built. Something similar to `pytz.OLSEN_VERSION`.

pganssle:
@skontar Oh, sorry, I completely misunderstood what you were asking. I thought you were looking for the *current version* (i.e. most recent) of the *database*, not the version *number* that the library is currently using. I agree that providing the version number is a good idea.

PR #39, opened 2015-02-19
Updated tzinfo version.  Added version feature to zinfo file.

pganssle:
Added a ".version" file to the built tarball to fix Issue #27. This is accessible by `dateutil.zoneinfo.gettz_db_version()`. I believe that the way pytz does it they hard-code the version into the library, but I wanted to preserve the ability for users to use arbitrary zoneinfo_metadata when building dateutil themselves.


jarondl:
Hi @pganssle , good work, but please don't merge yet. I have another idea I want to consider.

pganssle:
@jarondl No problem. I marked this PR and the other two as 2.5.0 anyway, so I wasn't planning on merging them until after the 2.4.1 release, because they all add features rather than fix bugs.

This one especially I wanted to give a bit more time in terms of code review, because there are a lot of ways to do it. I really would have preferred to be able to infer the version from the tzfile itself, but I didn't see any obvious way to do that (there's a version in the MAKEFILE, but I checked old versions of the tzdata file and it wasn't always there, so I don't know if that's stable).

Anyway, I wasn't super happy with the elegance of this, so if you've got a better way to do it, I have zero problem with that.

pganssle:
Superceded by #85

jjrauch32-lab (commented):
k

PR #43, opened 2015-02-26
Boolean evaluation of relativedeltas on python 2.7

nimasmi:
Python 3 implements truth-testing by, among other things, calling the [`__bool__`][1] method of a class instance, whereas Python 2 calls the [`__nonzero__`][2] method. dateutil.relativedelta only implements `__bool__`, which means that for example:

``` python
>>> bool(relativedelta(days=0))
True
```
There's already an open [issue][3] for this on launchpad, with a fix.

This pull-request introduces a failing test.

[1]: https://docs.python.org/3/library/stdtypes.html#truth
[2]: https://docs.python.org/2/library/stdtypes.html#truth
[3]: https://bugs.launchpad.net/dateutil/+bug/1035038

### DIFF
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -375,6 +375,8 @@ Here is the behavior of operations with relativedelta:
                     self.minute is None and
                     self.second is None and
                     self.microsecond is None)
+    # Compatibility with Python 2.x
+    __nonzero__ = __bool__
 
     def __mul__(self, other):
         f = float(other)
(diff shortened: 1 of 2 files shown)


nimasmi:
And 16c999e8c71134401a78d4d46435517b2271d6ac introduces a fix, with credit due to [Cédric Krier].

[1]: https://launchpad.net/~cedk

pganssle:
Bug is confirmed. Patch and test look good to me. I think this can be merged for the 2.4.1 release.

Issue #44, opened 2015-02-27
Make relativedelta more inheritance friendly


pganssle:
Migrating this issue from launchpad issue [#1010199](https://bugs.launchpad.net/dateutil/+bug/1010199).

Quoting [Shai Berger](https://launchpad.net/~shai-platonix):
>Hi,

>I needed to subclass relativedelta. Then, it was disappointing to find that the __add__, __sub__ and __neg__ operations explicitly return a new relativedelta object; it would be much nicer if they could return a self.__class__ object.

>Thanks,
>Shai.

Issue #45, opened 2015-02-27
Add weeks argument to relativedelta


pganssle:
Migrating issue from launchpad [issue #727525](https://bugs.launchpad.net/dateutil/+bug/727525).

You can read the full exchange there, but the gist is that while there's a keyword for "weeks" in relativedelta, you cannot retrieve the number of weeks in a relativedelta object as you can for the other time periods.

The solution would be to offer a `weeks` property that returns `self.days // 7`. Seems reasonable.

PR #49, opened 2015-03-01
Relativedelta enhancements - inheritance and weeks property.

pganssle:
This resolves issues #44 and #45.


shaib:
:+1: (looked at issue #44, seems solved well indeed!)

PR #52, opened 2015-03-03
force TZID

aleontiev:


pganssle:
@aleontiev Can you be a bit more clear about the nature of this request? It seems a bit hacked together. What does it mean when you say that PostgreSQL forces all timezones to be UTC "on storage". What would be the difference between this and having the user make the changes to dtstart *before* passing it to dateutil?

Also, the failure messages on the CI don't seem to be very helpful, but this is failing for python 3. I suspect that it would fail everywhere if a test were added that actually uses this functionality, though.

dateutil/rrule.py
```diff
@@ -343,7 +345,24 @@ def __init__(self, freq, dtstart=None,
...
+
+        # <TIMEZONE HACK>
+
+        # if tzid is set, force the timezone on dtstart
+        # this is to support generating recurrences from events stored in PostgreSQL, 
+        # which converts all timezones to UTC on storage
+        if tzid:
+          tzid = gettz(tzid)
+          if dtstart.tzinfo:
+            dtstart = dtstart.astimezone(tz)
```
pganssle: What is this supposed to do? I don't see `tz` defined anywhere in the file.

aleontiev: My mistake; I meant to merge into the fork rather than back into the main repository. Going to close the PR. For clarity, this is a fix that is meant to force setting a timezone in a Postgres-stack where all timestamps are stored as UTC regardless of input timezone.

As you said, it is possible to do so outside of dateutil, but doing so would require a major sweep over our code and so I decided to push it into a fork of dateutil.rrule, resulting in a minimal change.

Sorry about the confusion!

PR #50, opened 2015-03-02
Implement RFC2445, and make str(rrule) use it.

schinckel:
Resolves #47.

This probably needs some tests. I'm getting test failures on master anyway though.

Issue #47, opened 2015-02-28
Make str(rrule) to print the formatted RFC string.


pganssle:
Migrated launchpad issue [#1406305](https://bugs.launchpad.net/dateutil/+bug/1406305).

Quoting [Manuel Vazquez Acosta](https://launchpad.net/~mva-led):
>Sometimes (while interfacing with a DB, for instance) it is important to get the RFC string representation of a rrule. It would be great the the `str(rrule)` would work out-of-the-box.

He also suggested:
>Though I haven't fully test it, this may be a stating point: http://schinckel.net/2014/05/28/rrule-to-rfc-string/

pganssle:
So this is a bit more complicated than I had hoped, unfortunately, as `rrule` provides several extensions to RFC2445 that prevent many rules from being properly captured. At the moment, my goal is to make `str(rr)` and `rrulestr()` reciprocal, but it seems to me that `rrulestr()` also doesn't support all of the extensions - for example, it does not support `byweekday=TH(-1)`. This may take a bit.

Issue #61, opened 2015-03-20
Break up unit tests by module


pganssle:
I'm thinking that, given the ever-growing number of unit tests, it seems like it might be better to split "tests.py" into "test_rrule.py", "test_parser.py", etc, to more cleanly delineate between the testing modules, and make it easier for people to add new tests with their PRs.

I can do this, but I just thought I'd open an issue first to see if there's any obvious reason why not.

AndrewPashkin:
I noticed that there is a lot of similar tests, like here:
https://github.com/dateutil/dateutil/blob/5415505dba36b71836c05b28279c6e67f15b7977/dateutil/test/test.py#L3268

They can be refactored using parametrized tests, in such fashion:
https://pytest.org/latest/parametrize.html

PR #62, opened 2015-03-21
Generate (mostly) RFC2445-compliant string for str(rrule)

pganssle:
I have also added support for the `TH(+1)` format in the `BYDAY`/`BYWEEKDAY` parameter to rrule.

This dramatically expands the number of RRULE tests, unfortunately, as all the rrule-creation tests have been duplicated for use with str(rrule). Some of these can be pruned, but probably not that many. I'd like to consider refactoring the tests, though, so that both `testFREQbyXXX` and `testToStrFREQbyXXX` pull from the same common list of `rrules`.

This solves Issue #47 and supercedes PR #50 (upon which it is originally based).


pganssle:
I'm going to go ahead and merge this - the main thing I'm uncertain about is the dramatic expansion of tests. I think maybe we can re-assess that when deciding future refactoring of the tests per #61.

mrmachine:
Can we also have this for `rruleset`? I started work on this but ended up not needing it (yet) so I don't have a nice PR with tests etc., but I do have this implementation if anyone wants to pick it up.

```python
    def __str__(self):
        """
        Output a string that would generate this ``rruleset`` if passed to
        ``rrulestr(forceset=True)``.
        """
        def strftime(dt):
            return dt.strftime('%Y%m%dT%H%M%S')
        output = []
        for rrule in self._rrule:
            bits = str(rrule).splitlines()
            if rrule._dtstart:
                output.append(bits.pop(0))
            output.append('RRULE:%s' % bits[0])
        if self._rdate:
            output.append(
                'RDATE:%s' % ','.join(strftime(dt for dt in self._rdate)))
        for exrule in self._exrule:
            bits = str(exrule).splitlines()
            if exrule._dtstart:
                output.append(bits.pop(0))
            output.append('EXRULE:%s' % bits[0])
        if self._exdate:
            output.append(
                'EXDATE:%s' % ','.join(strftime(dt for dt in self._exdate)))
            output.append('EXDATE:%s' % ','.join(self._exdate))
        return '\n'.join(output)
```

PR #66, opened 2015-03-25
Remove 'setcachsize' from zoneinfo __all__ list

ryanss:
This function was removed from the zoneinfo module in version 2.3 and
causes an error when doing a wildcard import from the zoneinfo module.

```python
>>> from dateutil.zoneinfo import *
      AttributeError: 'module' object has no attribute 'setcachesize'
```

### DIFF
--- a/dateutil/zoneinfo/__init__.py
+++ b/dateutil/zoneinfo/__init__.py
@@ -12,7 +12,7 @@ from contextlib import closing
 
 from dateutil.tz import tzfile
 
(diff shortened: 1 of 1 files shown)


pganssle:
@ryanss Good catch, and thanks for the contribution! There was a test failure but it was a bug that happens occasionally where the Travis CI server can't reach the time zone server - nothing to do with your code, so I'm going to go ahead and merge.

PR #90, opened 2015-06-01
Ignore timezone data

pquentin:


pquentin:
@jarondl, @pganssle, any stance on this? Thanks!

pganssle:
@pquentin Probably a good idea, though I'm wondering if maybe it wouldn't be better to actually start including the most recent pre-built timezone data with the repo itself. I don't see how it hurts, as it's not platform specific, and it would probably be more convenient for people forking directly rather than installing through PyPi.

@jarondl Any idea why the timezone data file is not included in the repo? Is it the size?

jarondl:
@pganssle, I agree with you, let's add `dateutil-zoneinfo.tar.gz` to the repo,
but we should still ignore  `tzdata*.tar.gz`.
@pquentin , can you update your PR?

pquentin:
Updated the PR as asked.

Is there any good reason not to use pytz here?

pganssle:
Looks good to me. As for using pytz, I'd say it's probably not a good idea to add an additional dependency when it's not really necessary, and it seems like quite a bit of work to re-work the whole tz module around pytz.

Issue #78, opened 2015-04-15
Empty string parses to today's date


ksonj:
Maybe this behavior is intended but it certainly violates the principle of least surprise.

    dateutil.parser.parse('')
    >>> datetime.datetime(2015, 4, 15, 0, 0)

Is this documented somewhere? I'd expect this to raise a `ValueError`.

pganssle:
@ksonj I'll double check the documentation, but this is basically the expected behavior. The default parser behavior is that if an element is missing, it is replaced with the corresponding element from the `default` parameter. The `default` parameter is optional, and defaults to a timezone-naive `datetime` object representing midnight today's date.

Since an empty string contains no date elements, all elements are replaced with the values from the default element. The only time `ValueError` is raised is if the date string is malformed in some way.

In 2.5.0 (and in the current experimental version in the master branch here), I've added a `smart_defaults` option which tweaks this behavior somewhat, but would still parse an empty string as a valid date.

I think that it might be a good idea to change this in 2.5.0, though. A date string containing no date likely qualifies as exceptional. I would think it's also particularly important when doing fuzzy date parsing, since there's no general and obvious way to tell whether the date returned came from the string or from the default, e.g.:

```python
    >>> from dateutil.parser import parse
    >>> from datetime import datetime

    >>> dflt_date = datetime(2015, 03, 03)

    >>> parse('What a lovely day it was.', fuzzy=True, default=dflt_date)
    datetime.datetime(2015, 3, 3, 0, 0)

    >>> parse('What a lovely day it was on March 3rd.', fuzzy=True, default=dflt_date)
    datetime.datetime(2015, 3, 3, 0, 0)
```

PR #91, opened 2015-06-01
Fail correctly when using with_fuzzy_tokens and incorrect strings

MichaelAquilina:
For certain test cases, using `with_fuzzy_tokens=True` will cause parse to fail because a tuple is expected to be returned by _parse. This is then incorrectly unpacked which results in a confusing TypeError message. The standard way would be to allow a ValueError to be thrown. 

Example of issue:

    >> import dateutil.parser
    >> dateutil.parser.parse('04/04/0d4', fuzzy_with_tokens=True)

      File "dateutil/parser.py", line 1139, in parse
      File "dateutil/parser.py", line 452, in parse
        res, skipped_tokens = self._parse(timestr, **kwargs)
    TypeError: 'NoneType' object is not iterable
    

This pull request solves this issue for cases similiar to those specified in the test included (and any other errors that result in the internal _parse method failing).

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -793,7 +793,7 @@ class parser(object):
                                     assert mstridx == -1
                                     mstridx = len(ymd)-1
                                 else:
-                                    return None
+                                    return None, None
 
                             i += 1
 
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -448,12 +448,7 @@ class parser(object):
         else:
             effective_dt = default
 
-        if kwargs.get('fuzzy_with_tokens', False):
-            res = self._parse(timestr, **kwargs)
-            if res is not None:
-                res, skipped_tokens = res
-        else:
-            res = self._parse(timestr, **kwargs)
+        res, skipped_tokens = self._parse(timestr, **kwargs)
 
         if res is None:
             raise ValueError("Unknown string format")
@@ -842,7 +837,7 @@ class parser(object):
                         i += 1
 
                     elif not fuzzy:
-                        return None
+                        return None, None
                     else:
                         i += 1
                     continue
@@ -971,7 +966,7 @@ class parser(object):
                         # -[0]3
                         res.tzoffset = int(l[i][:2])*3600
                     else:
-                        return None
+                        return None, None
                     i += 1
 
                     res.tzoffset *= signal
@@ -989,7 +984,7 @@ class parser(object):
 
                 # Check jumps
                 if not (info.jump(l[i]) or fuzzy):
-                    return None
+                    return None, None
 
                 if last_skipped_token_i == i - 1:
                     # recombine the tokens
@@ -1004,7 +999,7 @@ class parser(object):
             len_ymd = len(ymd)
             if len_ymd > 3:
                 # More than three members!?
-                return None
+                return None, None
             elif len_ymd == 1 or (mstridx != -1 and len_ymd == 2):
                 # One member, or two members with a month string
                 if mstridx != -1:
@@ -1068,15 +1063,15 @@ class parser(object):
                         res.month, res.day, res.year = ymd
 
         except (IndexError, ValueError, AssertionError):
-            return None
+            return None, None
 
         if not info.validate(res):
-            return None
+            return None, None
 
         if fuzzy_with_tokens:
             return res, tuple(skipped_tokens)
         else:
-            return res
+            return res, None
 
 DEFAULTPARSER = parser()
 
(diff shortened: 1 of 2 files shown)


pganssle:
@MichaelAquilina This is clearly a problem, but is this actually the correct solution for this? In one of your test cases:

```python
parse('04/04/04 +32423', fuzzy_with_tokens=True)
```

This fails with the `TypeError` you mention, but I'm not sure why it should fail at all. This doesn't seem like a malformed fuzzy date-string to me, it seems like it should parse to `((datetime(2004, 4, 4, 0, 0), ('+32423',))`. It seems to me that the problem is that `fuzzy_with_tokens` has a bug when parsing strings containing numbers. Note that `fuzzy=True` handles that string just fine, and `fuzzy_with_tokens` is (as far as I know) just supposed to be a fuzzy parser that retains the unused tokens.

MichaelAquilina:
@pganssle I tracked down possible failures by reading the code in _parse. The first time I encountered the bug was then I had something similiar to the form '04/04/0d4'. I read through the source code and figured out why it was failing. The biggest issue seems to be that _parse returns None when it fails to parse a string. 

As you rightly pointed out though, I could change it so that it returns the string as part of the fuzzy result. I believe the code which handles the error gracefully by checking if result is None first should still be there however in the case where an unforseen code path fails.

MichaelAquilina:
@pganssle maybe you could provide me with some expected outputs for the inputs that are failing? I could aim to get those passing as part of the requirements.

pganssle:
@MichaelAquilina Yes, I'll take a look at this soon. The main concern I have here is that we may not want to "catch" the `None` error and pass it up as a `ValueError`, because the fact that it's raising a `TypeError` at all may actually be an indication that `_parse` isn't doing the right thing, and we should fix it at that point in the chain. Is there anywhere in the codebase that relies on `_parse` returning just `None`? If not, would returning `None, None` be more appropriate?

As for the outputs, for now, I think you can assume that at the very least `fuzzy_with_tokens` should return a tuple where the first item is what `fuzzy=True` would return, and the second item is the stuff that was ignored by the fuzzy parser.

MichaelAquilina:
@pganssle what should happen in the scenario where all tokens are ignored?

pganssle:
@MichaelAquilina If we're being consistent with the current behavior, it would return `(default, the_tokens)` where `default` is the `default` argument to `parse` and `the_tokens` is the token tuple. Per my reasoning in Issue #78, I think that should change in 2.5.0 and if all tokens are ignored, it should raise a `ValueError`, not being a valid date.

MichaelAquilina:
@pganssle see what you think about the latest update. 

I changed _parse so that it consistently returns a tuple. If `fuzzy_with_tokens` is False the second tuple will always be None. If it is set to True then it will return a tuple of skipped tokens (unless an error has occurred). 

However `parse` will consistently raise a `ValueError` if no datetime was found in the string passed regardless if `fuzzy_with_tokens` is True or False (just like you specified in your last comment).

dateutil/parser.py
```diff
@@ -1066,15 +1063,15 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
...
 
         if not info.validate(res):
-            return None
+            return None, None
 
         if fuzzy_with_tokens:
             return res, tuple(skipped_tokens)
         else:
-            return res
+            return res, None
```
MichaelAquilina: In my opinion, because return will always be returning a tuple - it would probably be best if the internal _parse method always return the skipped_tokens. These can then just be ignored if required.

MichaelAquilina:
@pganssle any update on this?

pganssle:
@MichaelAquilina I think this is fine. I'll make a separate bug report about the issue related fuzzy parsing with tokens, and I can split up the tests if necessary.

The only thing that needs to be fixed is that [Line 796](https://github.com/MichaelAquilina/dateutil/blob/fix_tuple_bug/dateutil/parser.py#L796) is still returning `None`.

Personally, I don't love the fact that there are so many return values in the parser. I think in the next pass through this code, we should raise `ValueError` or `IndexError` or `AttributeError` within the loop rather than returning `None`, so that it will be caught on [Line 1065](https://github.com/MichaelAquilina/dateutil/blob/fix_tuple_bug/dateutil/parser.py#L1065); this will at least centralize the exceptional return conditions. I'll have to dig into that more deeply, though, to see why it's done this way.

MichaelAquilina:
@pganssle good catch. Will update.

Also, I with your agree with your opinion. Some interface cleanup could probably be beneficial after this.

dateutil/parser.py
MichaelAquilina: Any opinion on this @pganssle ?

pganssle:
Ok, this looks fine to me, merging. Thanks for the thorough contribution, @MichaelAquilina!

MichaelAquilina:
Thanks, glad to be of help :)

PR #96, opened 2015-06-21
When possible dont convert years when the year has been specified past two digits

Bachmann1234:
Howdy!

While playing around with dateutil parsing I found what I think is a bug

```
datetime(99, 1, 1, 0, 0).isoformat()
'0099-01-01T00:00:00'

dateutil.parse('0099-01-01T00:00:00')
datetime.datetime(1999, 1, 1, 0, 0)
```

This PR tries to address this. When the library can identify a single possible year token, if that year is greater than 3 digits it will assume the year is correct rather than coercing it.

I also noticed problems with years like 0039. So I tried to address that as well

A better solution would be to detect padding zeros and handle those tokens differently... but I was struggling with that so I took the post processing approach instead.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -342,20 +342,35 @@ class parserinfo(object):
 
         return self.TZOFFSET.get(name)
 
-    def convertyear(self, year):
-        if year < 100:
+    @staticmethod
+    def _find_potential_year_tokens(year, tokens):
+        result = []
+        for t in tokens:
+            try:
+                if int(t) == year:
+                    result.append(t)
+            except ValueError:
+                pass
+        return result
+
+    def convertyear(self, year, tokens=None):
+        potential_year_tokens = self._find_potential_year_tokens(year, tokens) if tokens else []
+        year_token_unambiguous = len(potential_year_tokens) == 1
+        if (year_token_unambiguous and len(potential_year_tokens[0]) > 2) or year >= 100:
+            return year
+        else:
             year += self._century
             if abs(year-self._year) >= 50:
                 if year < self._year:
                     year += 100
                 else:
                     year -= 100
-        return year
+            return year
 
-    def validate(self, res):
+    def validate(self, res, tokens):
         # move to info
         if res.year is not None:
-            res.year = self.convertyear(res.year)
+            res.year = self.convertyear(res.year, tokens)
 
         if res.tzoffset == 0 and not res.tzname or res.tzname == 'Z':
             res.tzname = "UTC"
@@ -1065,7 +1080,7 @@ class parser(object):
         except (IndexError, ValueError, AssertionError):
             return None, None
 
-        if not info.validate(res):
+        if not info.validate(res, l):
             return None, None
 
         if fuzzy_with_tokens:
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -342,30 +342,19 @@ class parserinfo(object):
 
         return self.TZOFFSET.get(name)
 
-    @staticmethod
-    def _find_potential_year_tokens(year, tokens):
-        result = []
-        for t in tokens:
-            try:
-                if int(t) == year:
-                    result.append(t)
-            except ValueError:
-                pass
-        return result
-
     def convertyear(self, year, tokens=None):
-        potential_year_tokens = self._find_potential_year_tokens(year, tokens) if tokens else []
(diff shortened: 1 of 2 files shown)


Bachmann1234:
Looks like the build is failing because the timezone file is not downloading. 

```urllib.error.URLError: <urlopen error ftp error: ConnectionResetError(104, 'Connection reset by peer')>```

File: updatezinfo.py [hunk 1]
```diff
@@ -17,9 +18,19 @@ def main():
...
-        request.urlretrieve(os.path.join(metadata['releases_url'],
-                                         metadata['tzdata_file']),
-                            metadata['tzdata_file'])
+        for x in range(1, 6):
+            try:
+                request.urlretrieve(os.path.join(metadata['releases_url'],
+                                                 metadata['tzdata_file']),
+                                    metadata['tzdata_file'])
+                break
+            except Exception:
```
ckcollab: What exception should this be catching, shouldn't you just catch that one? Some quick googling suggests `URLError`?

*Thanks for that Hypothesis article btw! :)*

Bachmann1234: I took the path of "whatever happens retry" but I guess its better to be more specific :-P

I double checked the docs for urllib.request and yeah. Seems like `URLError` is the one

Bachmann1234:
anybody out there?

pganssle:
@Bachmann1234 Sorry for the delay, I'll try to take a look at this this weekend.

updatezinfo.py
```diff
@@ -17,9 +19,19 @@ def main():
 
     if not os.path.isfile(metadata['tzdata_file']):
         print("Downloading tz file from iana")
-        request.urlretrieve(os.path.join(metadata['releases_url'],
-                                         metadata['tzdata_file']),
-                            metadata['tzdata_file'])
+        for x in range(1, 6):
```
pganssle: This is not necessary. We can restart the CI builds if they fail, and anyone actually using the file can just re-run the script of the download fails.

Bachmann1234:
Cool, I'm away from a dev machine for a couple weeks but I'll address any comments asap

updatezinfo.py
Bachmann1234: Ok, I can remove this commit when I get back. I just thought it would be nice not to have to restart the CI. I also did not like seeing the red x in this pr :-p

Bachmann1234:
@pganssle ok, as requested I removed the bit of code that did the retry logic in the timezone data download

Bachmann1234:
Any other issues?

pganssle:
@Bachmann1234 This is a more complicated patch than some of the other PRs that have been accepted lately, so it's not as easy to give it a quick yes or no. Sorry for the delay. I'll try to make time to look in more detail soon.

Bachmann1234:
@pganssle no problem. Ill hang back for a while

pganssle:
Still processing, but at the moment the thing I don't love about this is that it modifies the interface of the ``parserinfo`` class. I think this might be handled better upstream, but it's not entirely clear where. I've been playing around with a solution using a class like this:

```python
class _ymd(list):
    def __init__(self, *args, **kwargs):
        super(self.__class__, self).__init__(*args, **kwargs)
        self.century_specified = False

    def append(self, val):
        if hasattr(val, '__len__'):
            if len(val) > 2:
                self.century_specified = True
        elif val > 100:
            self.century_specified = True

        super(self.__class__, self).append(int(val))
```

I wonder if maybe it's a good idea to move some of the YMD-parsing logic into a class like this. I have a working prototype of a solution using that, but it hasn't been cleaned up and fully vetted. It does still have the problem that it modifies ``parserinfo``, though, but the modification is there as a convenience (so that `convertyear` is still called by `validate`)

One thing that would be nice with this approach (which I have not yet implemented) is that it would be easy to modify it so that we can specify that we already know that something is the year, month or day as we are going.

@Bachmann1234 I have a working example at PR #129 if you want to take a look at it. However you want to play it, I'd rather keep the changes to the public-facing interface minimal. At the moment, I'd be OK with either you incorporating my suggestions into your version or just refactoring your version such that the public interface of `validate` doesn't have to change. If you need to temporarily dump the tokens into `res`, that's fine. Let me know what you think.

Bachmann1234:
@pganssle Sounds good. I should be able to figure that out. Ill try to look in detail some time this week.

Bachmann1234:
@pganssle I added to your work in https://github.com/dateutil/dateutil/pull/130

PR #98, opened 2015-07-12
Fix tzwin encoding errors

### LINKED ISSUE #92: Encoding error on windows
Just got this in a [Nikola Issue](https://github.com/getnikola/nikola/issues/1806#issuecomment-110844203)

```
python -c "import dateutil.tz; print dateutil.tz.gettz('UTC')"
File "", line 1, in 
File "D:\Python27\lib\site-packages\dateutil\tz.py", line 964, in gettz
tz = tzwin(name)
File "D:\Python27\lib\site-packages\dateutil\tzwin.py", line 92, in init
self._stdname = keydict["Std"].encode("iso-8859-1")
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-4: ord
inal not in range(256)
```

It happens in a chinese locale, on Windows 7, python 2.7.9

The reporter is @xh84

jarondl:
Stop encoding `_dstname` and `_stdname` to bytes with `iso-8859-1`, because we cannot assume that's the
correct codepage. Uses the current solution we use in `tz.py`, which outputs unicode in Python3.
Fixes #92  and http://stackoverflow.com/questions/21296475/python-dateutil-unicode-warning
and https://bugs.launchpad.net/dateutil/+bug/1227221

### DIFF
--- a/dateutil/tzwin.py
+++ b/dateutil/tzwin.py
@@ -4,6 +4,8 @@ import struct
 
 from six.moves import winreg
 
+from .tz import tzname_in_python2
+
 __all__ = ["tzwin", "tzwinlocal"]
 
 ONEWEEK = datetime.timedelta(7)
@@ -42,6 +44,7 @@ class tzwinbase(datetime.tzinfo):
         else:
             return datetime.timedelta(0)
 
+    @tzname_in_python2
     def tzname(self, dt):
         if self._isdst(dt):
             return self._dstname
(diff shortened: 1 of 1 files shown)


pganssle:
Been meaning to spin up a container of some sort to test this with one of the specified locales. Is there a way to include a test for this sort of thing? I suspect it would involve modifying the CI configuration, not just a simple addition to the test suite, but I haven't been able to dig into it yet.

jarondl:
I have no idea how to do that. I don't know if @appveyor supports changing the locale.

pganssle:
@jarondl I can look into it. Does one of the existing tests cover this bug already if you are running it in the right locale? If not, maybe we should add a test now, even if it always passes for standard ASCII locales, and we can work on figuring out how to include locale testing in the CI later.

pganssle:
I tried playing around with this locally and have been unable to reproduce the reported bugs - it seems more complicated than calling `locale.setlocale()`, because it's pulling the list of time zones from the registry, and apparently the registry is not updated even if you change the OS language pack (I have only tested this with Japanese locale).

Either way, I think ```TZTest.testGettz()``` covers the bug as reported in #92 if the windows registry is in a non-UTC locale, so I'm going to go ahead and merge. I'll open a separate issue for adding a non-American locale in the CI.

jarondl:
Thank you for the code review and the merge, let's just hope we didn't break the library :)

Issue #100, opened 2015-07-21
tzlocal uses startup timezone; parse uses runtime timezone


mjschultz:
There appears to be a bug in the library depending on the import ordering and other libraries calling [`tzset`](https://github.com/django/django/blob/7da3923ba0c569aa23d0ab0a47a124af60a18f5b/django/conf/__init__.py#L133-134). Below is a snippet that demonstrates the bug without extra libraries.

When the `tzlocal` class is created it uses the current [`time.timezone`/`time.altzone`](https://github.com/dateutil/dateutil/blob/master/dateutil/tz.py#L108) to find a `_dst_offset` which is used at runtime. This class variable should remain constant through the life of the process.

Over in the parser however, the runtime value of [`time.tzname`](https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py#L562-563) is checked and if it matches the parsed offset, a timezone of `tzlocal` is given.

Of course, if something has modified the process timezone via a `tzset` we now have a situation where we've classified something as `tzlocal()` even though the offset of that `tzlocal()` does not match the expected offset.

(And of course, since UTC never has daylight savings time I end up with the UTC-6 offset!)

I am currently in US/Central (UTC-5, Central Daylight Time).

```python
>>> from dateutil.tz import tzlocal
>>> from dateutil.parser import parse
>>> from datetime import datetime
>>> import time
>>> time.tzname
('CST', 'CDT')
>>> from os import environ
>>> environ['TZ'] = 'UTC'
>>> time.tzset()
>>> time.tzname
('UTC', 'UTC')
>>> t = parse('2015-07-20T20:52:19+00:00')
>>> t
datetime.datetime(2015, 7, 20, 20, 52, 19, tzinfo=tzlocal())
>>> print t
2015-07-20 20:52:19-06:00
>>> tzlocal().utcoffset(datetime(2015, 7, 20, 20, 52, 19))
datetime.timedelta(-1, 64800)
```

mjschultz:
I've made a PR in https://github.com/dateutil/dateutil/pull/101, comments are welcomed. This was the simplest method I could think of that wouldn't have performance ramifications.

pganssle:
Fixed by #107 and #109

PR #101, opened 2015-07-21
Cache tznames with _dst_offset

mjschultz:
This should add caching (and checking) of the tzlocal._tznames attribute to make sure that, even if the timezone is changed after loading, it does not return an incorrect value.

I've added an `env` section to the build script to test the code in two timezones since this bug only exhibits itself when the system starts in non-UTC then is changed to UTC at runtime.

Ref: https://github.com/dateutil/dateutil/issues/100

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ python:
   - "3.4"
   - "pypy"
   - "pypy3"
+env:
+  - TZ="US/Central"
+  - TZ="UTC"
 install:
   - pip install six
   - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install unittest2; fi
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -559,7 +559,7 @@ class parser(object):
                     raise ValueError("Offset must be tzinfo subclass, "
                                      "tz string, or int offset.")
                 ret = ret.replace(tzinfo=tzinfo)
-            elif res.tzname and res.tzname in time.tzname:
+            elif res.tzname and res.tzname in tz.tzlocal._tzname:
                 ret = ret.replace(tzinfo=tz.tzlocal())
             elif res.tzoffset == 0:
                 ret = ret.replace(tzinfo=tz.tzutc())
--- a/dateutil/test/test.py
+++ b/dateutil/test/test.py
@@ -5946,6 +5946,7 @@ END:VTIMEZONE
         dt = parse("2013-03-06 19:08:15")
         self.assertFalse(tz._isdst(dt))
 
+    @unittest.skipIf(sys.platform.startswith("win"), "requires Unix")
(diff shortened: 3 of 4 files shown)


mjschultz:
I've pushed https://github.com/dateutil/dateutil/commit/24854f4d22e1df3c0eaa816f4d7bbcde52a8b279 to demonstrate that the new test fails in non-UTC timezones (except for a few runs that failed to download the timezone database).

I followed that with https://github.com/dateutil/dateutil/commit/9b57f676d8e579586269ba4fd55b635b42badba7 which shows that the fix works with the new tests (again, ignoring the cases where the update of the timezone database failed).

pganssle:
Thanks for the issue report and PR! Some comments:

I'll have to look at this more closely - I only briefly looked at the issue, but my first impression is that the parser shouldn't be accessing "non-public" functions or properties from tz (`tzlocal._tzname`). Obviously it's valid python, but it's not good to rely on implementation details like that.

I am also not sure that caching is the right behavior - this was something I wanted to think more about before responding to #100. I'll respond over there in a little bit with a more considered strategy, though you've obviously thought about this more than me, so my intuitions may be way off.

mjschultz:
Yeah, there are a number of ways to solve this ranging from small to large. This one I know wouldn't break any backward expectations. The trivial-est of changes would be to [swap the tzlocal() and the tzutc() cases in the if statement of parse](https://github.com/dateutil/dateutil/compare/master...mjschultz:issue100?expand=1).

I hesitated on that because someone out there might be expecting tzlocal() if they are in the UTC timezone, so in some sense it would change the user expectations; though I'm not sure how reasonable that is.

As you say, caching is probably not the best approach but the current tzlocal does caching on the tz offset anyway so it's a "more of the same" thing right now. I think the best fix would be to remove the `_dst_offset` from the `tzlocal` and construct it every time, but that would have performance ramifications.

As maintainer, I'll defer to whatever you think is best. At the very least I hope the test makes its way in if you decide this is something dateutil should handle!

carlosxl:
@mjschultz I think I went into a similar issue as you do.

I think the main inconsistency with the `tzlocal` class is that some of its methods reference `time` module in real time (e.g., `.tzname()`) while some use the cached values upon module import (e.g., `.utcoffset()`).  Therefore, caching `tzname` just like we do for `_std_offset` or `_dst_offset` feels like the right way of solving the problem.

On the other hand, in my opinion, fix the timezone information for each `tzlocal` instances might be a good idea too. So I prepared a patch #107 to achieve this. Then I realized that it won't fix your problem where `time.tzname` is still not fixed during instance initialization.

Would please take a look and see if caching `_tzname` in the `__init__` method makes sense to you. Also I'm not very familiar with the testing framework, would please check if your testing case can cover the changes in #107 as well?

Thanks!

c.c. @pganssle

PR #103, opened 2015-07-26
Tzinfo docfix

### LINKED ISSUE #80: parser.parse does not apply time zone to date if ignoretz is True and tzinfos is given
Quoting the [docs](https://dateutil.readthedocs.org/en/2.4.2/parser.html#dateutil.parser.parse):

> **tzinfos** – A time zone, to be applied to the date, if ignoretz is True. This can be either a subclass of tzinfo, a time zone string or an integer offset.

I hope that i'm not misinterpreting docs, but i would expect `parse` to return aware datetime.

```python
Python 2.7.8 (default, Oct 20 2014, 15:05:19) 
Type "copyright", "credits" or "license" for more information.

IPython 2.3.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import dateutil

In [2]: dateutil.__version__
Out[2]: '2.4.2'

In [3]: from dateutil.parser import parse

In [4]: parse("Sat Oct 11 17:13:46 UTC 2003", ignoretz=True, tzinfos='Europe/Tallinn')
Out[4]: datetime.datetime(2003, 10, 11, 17, 13, 46)
```

Instead it returns naive datetime.

pganssle:
Fix inaccurate documentation - fixes #80.

Also I've adjusted the formatting of the docs. There's still some significant repetition of text here, it would be best to figure out the right way to transclude that information so it can be more easily updated, but I'll leave that for a later fix.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -9,7 +9,7 @@ a date/time stamp is omitted, the following rules are applied:
 - If AM or PM is left unspecified, a 24-hour clock is assumed, however, an hour
   on a 12-hour clock (`0 <= hour <= 12`) *must* be specified if AM or PM is
   specified.
-- If a time zone is omitted, it is assumed to be UTC.
+- If a time zone is omitted, a timezone-naive datetime is returned.
 
 If any other elements are missing, they are taken from the `datetime.datetime`
 object passed to the parameter `default`. If this results in a day number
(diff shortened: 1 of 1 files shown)


jarondl:
LGTM.
In the future, now that readthedocs.com supports `napoleon`,
do you think we should migrate to numpydoc style?

pganssle:
@jarondl I'm not familiar enough with the advantages of numpydoc style to understand the advantages. The biggest issue for me is when you have large numbers of "inherited" kwargs that you want to document (e.g. `parse()` passing on its `*kwargs*` to the `parser()` constructor) - it'd be nice to be able to just include an inline reference to the "inherited" kwargs from the base class/function.

Issue #28, opened 2015-02-05
Allow comma as a separator between seconds and microseconds


b4stien:
According to http://en.wikipedia.org/wiki/ISO_8601#Times (relevant quote: http://en.wikipedia.org/wiki/ISO_8601#cite_note-15) the preferred decimal sign should be a comma (although a dot is also valid).

Currently a comma is not recognised (but a dot is).

````
>>> parse("2015-02-05T13:34:04,000+0000")
[...]
ValueError: unknown string format

>>> parse("2015-02-05T13:34:04.000+0000")
datetime.datetime(2015, 2, 5, 13, 34, 4, tzinfo=tzutc())
```

pganssle:
This should probably be addressed while fixing (or otherwise making consistent) Issue #19.

b4stien:
Is there something I can do to help? Do you want me to write some tests?

pganssle:
@b4stien I don't think the tests will be a big problem, they're usually pretty easy to write once the code is there. If you'd like to try tackling this problem, I've tried recently to improve the documentation of the algorithm, so hopefully it's getting easier to understand from reading the code.

The way the parser currently interprets things is quite complicated, and there can be various collisions in interpretation, so be careful what you change. There's probably a way to simply add commas as decimal points, but at the moment dots are treated differently than other separators because they may be *part of a number*, not just a separator. I worry that by allowing commas as decimal points, there could be similar interpretation collisions. These issues have to be dealt with to resolve Issue #19 anyway, and it's possible that the best way to fix things would be to overhaul the way the parser tokenizes and processes the strings rather than to add in comma support on top of the existing framework.

b4stien:
So I tried for a while, and thanks for your insights, but I'm afraid I didn't end up with something satisfying...

pganssle:
@b4stien Yes, one of the biggest problems is that what formats are and are not supported is not actually documented. That was my goal with Issue #26, in fact. I think that someone (maybe me, though you are welcome to take a crack at it) needs to take the time and figure out what date formats are currently supported and in what combinations, then write up that documentation.

From there, we can see which ones make sense to keep and how to re-write the parser to maintain the same functionality but be a bit easier to maintain and extent.

pganssle:
Note: This bug was also reported on the launchpad as [issue #974463](https://bugs.launchpad.net/dateutil/+bug/974463).

PR #106, opened 2015-08-13
Parse python logger timestamps

ryanss:
By default, the Python logging module uses a format that includes `%(asctime)s`. The Python docs define this variable as:

Human-readable time when the LogRecord was created. By default this is of the form ‘2003-07-08 16:49:45,896’ (the numbers after the comma are millisecond portion of the time).

Example: `2003-09-25 10:49:41,502`

Currently, dateutil fails to parse a string like this:
```python
>>> parse("2003-09-25 10:49:41,502")
  File "/usr/local/lib/python2.7/site-packages/dateutil/parser.py", line 1008, in parse
  File "/usr/local/lib/python2.7/site-packages/dateutil/parser.py", line 395, in parse
    raise ValueError("Unknown string format")
ValueError: Unknown string format
```

This patch enables dateutil to parse these timestamps which are extremely common in Python log files.

```python
>>> parse("2003-09-25 10:49:41,502")
datetime.datetime(2003, 9, 25, 10, 49, 41, 502000)
```

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -46,6 +46,7 @@ import datetime
 import string
 import time
 import collections
+import re
 from io import StringIO
 from calendar import monthrange, isleap
 
@@ -58,6 +59,9 @@ __all__ = ["parse", "parserinfo"]
 
 
 class _timelex(object):
+    # Fractional seconds are sometimes split by a comma
+    _split_decimal = re.compile("([\.,])")
+
     def __init__(self, instream):
         if isinstance(instream, binary_type):
             instream = instream.decode()
@@ -145,7 +149,7 @@ class _timelex(object):
                 # numbers until we find something that doesn't fit.
                 if nextchar in numchars:
                     token += nextchar
-                elif nextchar == '.':
+                elif nextchar == '.' or (nextchar == ',' and len(token) >= 2):
                     token += nextchar
                     state = '0.'
                 else:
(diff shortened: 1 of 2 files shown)


pganssle:
Thanks @ryanss. I haven't given this a thorough look-over yet, but I believe this solves Issue #28. I remember when I looked at this some months ago I found it a bit difficult to handle all the edge cases, so I'll have to see what I was thinking about at the time.

pganssle:
@ryanss I think my previous concern was that there could be some weird edge cases where this approach could be problematic and since there isn't actually a spec saying what formats and what variations on formats we support, it's hard to say whether this could break any of it.

That said, I think the best thing to do is to take our unit tests as a *de facto* spec for what formats are supported, and since this doesn't break the unit tests, it's a :+1: from me.

I did make a PR against your branch to pre-compile the regular expression. Since this tokenizer is called in the innermost loop of the parser, I think it's best to move the overhead of creating the regular expression into a static member. Let me know if you disagree.

ryanss:
Completely agree, this is much more efficient. I've updated my pull request to reflect the compiled regular expression changes.

pganssle:
Looks good, merging.

wimglenn:
@pganssle  

It seems to ignore the year in the string, and use the current year (or the year of `default` if that was specified)

```
>>> import dateutil.parser
>>> dateutil.parser.parse("May 20,2019")
datetime.datetime(2020, 5, 20, 0, 0)
```

git bisected it to this logger change, specifically https://github.com/dateutil/dateutil/commit/bc69c3f3151a3bc9a743d80d29d48701468984ee

In python-dateutil<2.5.0 the result was parsed correctly.

Any way to make dateutil throw an exception here instead?

PR #107, opened 2015-08-14
Initialize `_std_offset` and `_dst_offset` per tzlocal instance

carlosxl:
Currently the `_std_offset` and `_dst_offset` are one-off set during module import using the timezone info from `time.timezone`.

For many users especially Django users (Django set timezone info according to TIME_ZONE setting), the time.timezone may change when you set environment variable `os.environ['TZ']` and call `time.tzset()` (ref: https://docs.python.org/2/library/time.html#time.tzset).

Therefore, new instances created by `dateutil.tz.tzlocal()` should be able to reflect this change, i.e., the timezone offset should not be fixed at the import time.

Use the following code to reproduce the issue:
```
from dateutil import tz
import time
import os

print time.timezone
print tz.tzlocal()._std_offset


os.environ['TZ'] = 'Etc/UTC'
time.tzset()

print time.timezone
print tz.tzlocal()._std_offset
```

Before patch:
```
>>> from dateutil import tz
>>> import time
>>> import os
>>> 
>>> print time.timezone
-28800
>>> print tz.tzlocal()._std_offset
8:00:00
>>> 
>>> 
>>> os.environ['TZ'] = 'Etc/UTC'
>>> time.tzset()
>>> 
>>> print time.timezone
0
>>> print tz.tzlocal()._std_offset
8:00:00
```

After patch:
```
>>> from dateutil import tz
>>> import time
>>> import os
>>> 
>>> print time.timezone
-28800
>>> print tz.tzlocal()._std_offset
8:00:00
>>> 
>>> 
>>> os.environ['TZ'] = 'Etc/UTC'
>>> time.tzset()
>>> 
>>> print time.timezone
0
>>> print tz.tzlocal()._std_offset
0:00:00
```

### DIFF
--- a/dateutil/tz.py
+++ b/dateutil/tz.py
@@ -104,12 +104,12 @@ class tzoffset(datetime.tzinfo):
 
 
 class tzlocal(datetime.tzinfo):
-
-    _std_offset = datetime.timedelta(seconds=-time.timezone)
-    if time.daylight:
-        _dst_offset = datetime.timedelta(seconds=-time.altzone)
-    else:
-        _dst_offset = _std_offset
+    def __init__(self):
+        self._std_offset = datetime.timedelta(seconds=-time.timezone)
+        if time.daylight:
+            self._dst_offset = datetime.timedelta(seconds=-time.altzone)
+        else:
+            self._dst_offset = self._std_offset
 
     def utcoffset(self, dt):
         if self._isdst(dt):


jarondl:
Thank you @carlosxl ,
Your version does make more sense.
Merged.

pganssle:
@jarondl @carlosxl I haven't had a chance to compare, does this solve Issue #100? If so, does it obviate PR #101?

carlosxl:
@pganssle @jarondl good point! I did not realize that there was similar issue reported already, but I do think that I tried to solve a similar issue as #100. I went through the PR #101 and actually got some thoughts. I think the two issues are not exactly the same, some further code change is required in order to fix #100. I left comments in PR #101 directly.

Regards,

mjschultz:
No, this should solve #100 and invalidate my PR #101. I do like this approach better from a code cleanliness perspective, the only reason I didn't go this route originally was because of performance concerns with computing these values at runtime.

I don't think it will matter much unless someone is doing a bunch of dateutil parsing or tzlocal instantiation in a tight loop. On my box the difference is seen below:

```
[dateutil@80a08fa...]$ python -m timeit -s "from dateutil.tz import tzlocal" "tzlocal()"
10000000 loops, best of 3: 0.092 usec per loop
```

vs. the new:

```
[dateutil@master]$ python -m timeit -s "from dateutil.tz import tzlocal" "tzlocal()"
100000 loops, best of 3: 2.06 usec per loop
```

PR #109, opened 2015-08-18
Add test around doing tzset post-initialization

mjschultz:
This adds a regression test for issue #100 which has been resolved via #107.


pganssle:
Great, thanks for the test @mjschultz!

PR #113, opened 2015-08-25
relativedelta: Do not forget to compare microseconds

elprans:
relativedelta forgets about microseconds when comparing.


pganssle:
Weird. Is it possible this was originally intended as a feature? I'm going to merge, since it seems like this is definitely the right thing to do. Hopefully no one was counting on this.

PR #115, opened 2015-09-01
Tox: install unittest2 under Python 2.6

moreati:
This fixes testing under Tox and matches the behaviour of Travis


pganssle:
I don't totally understand this well enough to just accept right away (especially since it concerns tests in a non-Travis environment, so the unit tests won't be much guide), but I'll test tonight and accept when it doesn't break everything.

Thanks!

jarondl:
@pganssle, it works for me, and makes sense, so I'm merging.
Thank you @moreati .

PR #122, opened 2015-09-10
Fix timelex

ms-boom:
fix wrong tokens on unicode string

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -71,12 +71,6 @@ class _timelex(object):
             instream = StringIO(instream)
 
         self.instream = instream
-        self.wordchars = ('abcdfeghijklmnopqrstuvwxyz'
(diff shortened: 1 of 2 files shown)


pganssle:
I like where this is going, but I think we're actually losing some flexibility by enforcing `isalpha()` and `isdigit()`, etc. Right now, you can subclass `_timelex()` and override the relevant character sets, but in your version, that's no longer possible.

That said, I think that this *is* the right *default* approach. Today I'll make a PR against your branch with a suggested improvement.

pganssle:
@jarondl This seems like a major change - any thoughts on the matter?

See also my PR at ms-boom/dateutil#1

pganssle:
I'm going to go ahead and merge this. We can always undo it later if it causes some significant problems.

jarondl:
Hi @pganssle, and @ms-boom,
a post-factum +1 from me. It is much better than what we had.

Issue #57, opened 2015-03-08
Incorrect timezone during PDT -> PST Switch


pganssle:
Migrated from launchpad issue [#1390262](https://bugs.launchpad.net/dateutil/+bug/1390262):

Quoting [Yupeng](https://launchpad.net/~liaoyupeng) on 2014-11-07:

    root@yupeng-desktop:~# date
    Sun Nov 2 01:43:42 PDT 2014
    root@yupeng-desktop:~# python
    Python 2.7.3 (default, Apr 10 2013, 06:20:15)
    [GCC 4.6.3] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    >>> from datetime import *
    >>> from dateutil.tz import *
    >>> datetime.now(tzlocal()).tzname()
    'PST'

pganssle:
I've looked into this and tracked it down to the [tz.tzlocal.__isdst](https://github.com/dateutil/dateutil/blob/master/dateutil/tz.py#L128) function, and as far as I can tell, there is no good solution. The time "November 2, 2014, 1:00 AM Pacific" is an ambiguous representation and apparently defaults to the PST side.

I think that the best thing to do, moving forward, is to get a good sense of where the ambiguity can be alleviated and either document the right way to deal with it or provide additional methods/functions that can help with that. For example, if you just want to know the *current* local time zone, `utcnow()` should be able to pretty unambiguously tell you what the current time it is, and you can localize based on that. The problem is that `__isdst` operates on any datetime, and datetime objects without time zones are inherently ambiguous around DST transitions.

zed:
@pganssle : there is no ambiguity here. `.now(tz)` uses `tz.fromutc()` and `utcnow()` (or equivalent) already e.g., [`datetime.now(pytz_timezone)` always works](http://stackoverflow.com/q/31886808/4279).

pganssle:
@zed It is certainly true that "November 2, 2014, 1:00 AM Pacific" is an ambiguous time:

```python
>>> pytz.timezone("America/Los_Angeles").utcoffset(datetime.datetime(2014, 11, 2, 1, 0))

pytz\tzinfo.pyc in utcoffset(self, dt, is_dst)
    406             return None
    407         elif dt.tzinfo is not self:
--> 408             dt = self.localize(dt, is_dst)
    409             return dt.tzinfo._utcoffset
    410         else:

pytz\tzinfo.pyc in localize(self, dt, is_dst)
    347         # ambiguous case
[...]

PR #127, opened 2015-09-18
Pep 0495

abalkin:
This PR is not intended to be merged.  I am opening it as a way to invite dateutil developers to join the discussion of [PEP 495](https://www.python.org/dev/peps/pep-0495/) "Local Time Disambiguation".

This PR demonstrates the changes that will be needed for dateutil to support the new "fold" attribute in datetime objects.


pganssle:
Thanks @abalkin. I'll take a look at the discussion. Seems like it would be helpful in solving #57 without the complicated effort of trying to detect the actual local time zone in a platform-independent way.

pganssle:
PEP has been accepted, closing.

PR #129, opened 2015-09-27
Demonstrate direction for new YMD parsing (don't merge)

pganssle:
Demonstration of an alternate approach for PR #96


pganssle:
This has served its purpose.

Issue #123, opened 2015-09-10
Make `_timelex` public


pganssle:
Currently the tokenizer `_timelex` is a private interface. I think it is worthwhile to make that public and pass the class to `parserinfo` so that applications built on dateutil can subclass the tokenizer either to optimize its performance for specific use cases or to add compatibility that might not be appropriate in the general case.

We may want to make a similar consideration about `_tzparser` as well.

I have also mentioned this in ms-boom/dateutil#1 and #122.

PR #130, opened 2015-09-30
Handle pre 100 dates redux

Bachmann1234:
I took your PR and modified it to work with the last test

I also noticed a few places where you were using `@classmethod` where `@staticmethod` may be more appropriate

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -372,9 +372,31 @@ class parserinfo(object):
 
 
 class _ymd(list):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, tzstr, *args, **kwargs):
         super(self.__class__, self).__init__(*args, **kwargs)
         self.century_specified = False
+        self.tzstr = tzstr
+
+    @staticmethod
+    def token_could_be_year(token, year):
+        try:
+            return int(token) == year
+        except ValueError:
+            return False
+
+    @staticmethod
+    def find_potential_year_tokens(year, tokens):
+        return [token for token in tokens if _ymd.token_could_be_year(token, year)]
+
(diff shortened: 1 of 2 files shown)


pganssle:
@Bachmann1234 Thanks for all the work on this - I think it's more or less ready to merge.

Unfortunately, I guess I was accidentally working from a fork of #122 when I made my demo, which is why those classmethods are even there in the first place. Don't worry about it, though. I'll do the merge manually and clean that part up.

That said, for future reference, I tend to prefer `@classmethod` to `@staticmethod` unless there's a good reason for it. As far as I can tell, there's no performance impact to using `@classmethod`, and it makes things simpler when subclassing, and so I set them as classmethods in preparation for #123.

If you know of any good reason why `@staticmethod` should be preferred when `@classmethod` is not strictly necessary, let me know, I'd be interested to learn.

Bachmann1234:
@pganssle I actually discussed `@staticmethod` at great lengths with people and honestly its hard to argue for. I use it when I don't need `cls` but many have pointed out that once you dont need `cls` you might as well consider just having it as a function because you are basically just using the classname as a namespace. 

ill drop the commit

Bachmann1234:
I dropped it btw :-)

pganssle:
Merged manually in [commit 4704a151](https://github.com/dateutil/dateutil/commit/4704a151ade5e5c0ad262e2e34bfd6506e7e4419). Thanks again for the work on this @Bachmann1234, sorry it took me so very long to get it merged in. Hoping to get a release out this weekend, so it should be in that.

Bachmann1234:
Hey man, no problem! I get how it can be hard to find the time. Thanks!

PR #134, opened 2015-10-25
Hotfix-133: Fix tzwin.py not importing when needed on Windows

labrys:
Exported `tzname_in_python2` into another file `tzcommon.py` and imported it in `tz.py` and `tzwin.py`

### DIFF
--- a/dateutil/tz.py
+++ b/dateutil/tz.py
@@ -19,7 +19,6 @@ from dateutil.tzcommon import tzname_in_python2
 try:
     from dateutil.tzwin import tzwin, tzwinlocal
 except ImportError:
-    print 'import error'
     tzwin = tzwinlocal = None
 
 relativedelta = None
--- a/dateutil/tz.py
+++ b/dateutil/tz.py
@@ -14,10 +14,12 @@ import sys
 import os
 
 from six import string_types, PY3
+from dateutil.tzcommon import tzname_in_python2
 
 try:
     from dateutil.tzwin import tzwin, tzwinlocal
 except ImportError:
+    print 'import error'
     tzwin = tzwinlocal = None
 
 relativedelta = None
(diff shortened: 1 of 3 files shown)


pganssle:
Why are you moving it to a third file if the problem is just that it's a relative import? You should still be able to import the function from tz.

dateutil/tz.py
```diff
@@ -14,10 +14,12 @@
 import os
 
 from six import string_types, PY3
+from dateutil.tzcommon import tzname_in_python2
 
 try:
     from dateutil.tzwin import tzwin, tzwinlocal
 except ImportError:
+    print 'import error'
```
pganssle: This is not something that should be in production code.

labrys:
Importing `tzname_in_python2` from `tz` directly caused `tz.py` to still fail to `import tzwin` with an import error.  Exporting the decorator to third file was an easy fix.  If you have a better suggestion, I'm all for it.

pganssle:
I'm just thinking that the reason your version works is that you're importing `dateutil.tzcommon` rather than `.tzcommon`, so switching to `from dateutil.tz import tzname_in_python2` will work just the same.

Either way, I'll need a few hours before I can properly analyze this PR / bug anyway, because I need to create a Windows VM. A Windows-specific bug *would* have to come up within *days* of me completely annihilating my Windows partition ;).

labrys:
Hehe, just figures with annihilating your Windows partition. I tried it as `from dateutil.tz...` first and it still failed.  Figured it might be because of being a recursive import or something stupid like that.

miigotu:
Cant just move `tzname_in_python2` into `dateutil.tzwin.py` ?

labrys:
Yeah, but I was tired at the time didn't think of it :).  It's kind of like the corollary to occam's razor: The simplest answer is usually the one you don't end up thinking of.

labrys:
@miigotu Actually I remember why I didn't want to move it to `dateutil.tzwin.py`.  In case `tzwin` failed to import for any reason `tzname_in_python2` would still be available to 'tz.py'

miigotu:
I think dateutil should handle this dynamically for the tzlocal vs tzwinlocal. Aka if caller is tzlocal and OS is win it should return tzwinlocal and vice versa

labrys:
Agreed, it should but it currently does not.

miigotu:
@labrys If you can, send me an email, miigotu@gmail

PR #136, opened 2015-10-27
use heapq to improve performance

dekoza:
Patch based on [an article by Brian Beck](http://blog.brianbeck.com/post/22129050/speeding-up-dateutil-pythons-heapq-module-turns).

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -1293,13 +1293,11 @@ class rruleset(rrulebase):
         self._genitem(rlist, iter(self._rdate))
         for gen in [iter(x) for x in self._rrule]:
             self._genitem(rlist, gen)
-        rlist.sort()
         exlist = []
         self._exdate.sort()
         self._genitem(exlist, iter(self._exdate))
         for gen in [iter(x) for x in self._exrule]:
             self._genitem(exlist, gen)
-        exlist.sort()
         lastdt = None
         total = 0
         heapq.heapify(rlist)
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -17,6 +17,7 @@ except ImportError:
 
 from six import advance_iterator, integer_types
 from six.moves import _thread
+import heapq
 
 __all__ = ["rrule", "rruleset", "rrulestr",
            "YEARLY", "MONTHLY", "WEEKLY", "DAILY",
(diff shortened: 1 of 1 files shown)


pganssle:
This looks good. I'm going to just double check the profiling before I merge.

jarondl:
Pretty cool.

pganssle:
I tried to replicate this, but I don't have an rruleset that is slow at the sort step. I tried randomly generating one, but I'm not quite sure I know how to set the parameters of the randomization correctly. I shot @exogen (Brian Beck) an e-mail to see if he might be able to provide any insight.

That said, it doesn't seem to be significantly *worse* to use `heapq`, and it does conceptually make sense, so we can probably just accept this as is. Still, I'd like to track down whether the bottleneck has changed since 2007 or if I just don't have the right data set. I don't see anything obvious that would have changed.

Here are my results: [profiling](https://gist.github.com/pganssle/2572b41e57f9de3a8c9b), [code](https://gist.github.com/pganssle/c68277475bbe769f2c93).

exogen:
Hi all! Been a while since I've seen this code.

You'll need a pretty big rruleset before you see any benefits from the patch. Here's an example one (which is exactly quite a bit smaller than the real ones I had before): https://gist.github.com/exogen/5e951c2e3d71cb2be8b4

(At the time, I was trying to create an rruleset for a bus schedule, and bus schedules aren't as simple as "every 15 minutes", but are published as fixed timetables – that's why I created each time in the timetable as a WEEKLY rrule. If I were to reimplement this today I'd probably avoid using an rruleset altogether and just implement the timetable lookup myself.)

Anyway, here's the time I get from that file with current dateutil:

```sh
$ time python rruleset.py 
2016-02-18 05:00:00

real    0m4.808s
user    0m4.727s
sys     0m0.063s
```

...and with this patch applied:

```sh
$ time python rruleset.py 
2016-02-18 05:00:00

real    0m0.576s
user    0m0.486s
sys     0m0.071s
```

The more rrules you have, the more valuable this patch becomes. Hope that helps! :)

exogen:
I wonder if the initial calls to `rlist.sort` and `exlist.sort` are also unnecessary now? I don't think `heapq.heapify` needs the list to be initially sorted.

pganssle:
@exogen Thanks so much for your help with this. With your code, I'm able to cleanly replicate this, so I am going to accept the PR.

@dekoza I think that Brian is right and `heapq.heapify` does not need the list to be initially sorted. I removed `rlist.sort` and `exlist.sort` and all the tests passed just fine. Do you want to make those changes and push another commit, or should I accept and make the changes myself?

dekoza:
I'll do this, thanks.

pganssle:
Brilliant. Merging! Thanks again @dekoza and @exogen

Issue #135, opened 2015-10-26
ValueError: tzinfo.utcoffset() returned 1440; must be in -1439 .. 1439


labrys:
Trying:
`datetime.datetime.now(tzwin.tzwinlocal()).utcoffset()`
on Windows raises:
`ValueError: tzinfo.utcoffset() returned 1440; must be in -1439 .. 1439`

This is after fixing the issue in #133 with `tzwin.py` not importing.
Found while troubleshooting SiCKRAGETV/SickRage-issues#3450

labrys:
Confirmed this in the current version as well with 
`Python27\python -c "import datetime, dateutil.tzwin; print datetime.datetime.now(tzwin.tzwinlocal()).utcoffset()"`
and 
`Python34\python -c "import datetime, dateutil.tzwin; print(datetime.datetime.now(dateutil.tzwin.tzwinlocal()).utcoffset())"`

pganssle:
This one I can confirm:

```python
>>> from dateutil.tzwin import tzwinlocal
>>> twl = tzwinlocal()
>>> print(twl._stdoffset)
-300
>>> print(twl._dstoffset)
-4294967536L
```
Shouldn't be too hard to figure out why that's happening.

pganssle:
I think that this is some funky thing having to do with the way the data is represented. Note:

```python
>>> iv = 4294967236
>>> print(~int(bin(((2 << 32) - 1) & ~iv), 2))
-60
```
I think it's a signed 32-bit integer being represented as an unsigned 32-bit integer for whatever reason.

I'll make the relevant changes tonight.

labrys:
Sounds good.

labrys:
Oh and thanks for all the hard work you've put in.  I don't mean to be a pain in the butt :grinning:

pganssle:
No problem, I appreciate the feedback about what needs fixing and your help in pinpointing the bugs!

miigotu:
ping me if any fixes for this are merged please, so I can update it in SR.

labrys:
Will do.

pganssle:
@labrys @miigotu Proposed fix at #142 if you guys want to take a look and see if it works for you.

labrys:
With this fix `utcoffset` does not throw errors.  

I assume the values are correct since per the python standard library documentation, the value returned must be a `datetime.timedelta` object specifying a whole number of minutes east of utc(west is negative).   
US Eastern Standard Time is UTC-5.  
So:
```python
[...]

PR #141, opened 2015-11-01
Fix tzwin import issue

pganssle:
Per discussion on #134, I've done a bit of refactoring to get this to work.


pganssle:
Merging this. Hopefully this leaves the public interface intact.

miigotu:
So how exactly should I call/differentiate between windows and unix now? Do I need to check sys.platform and selectively import like this?

```
try:
    if sys.platform..startswith('win'):
        from dateutil.tzwin import tzwinlocal as tzlocal
    else: 
        from dateutil.tz import tzlocal
except (ValueError, ImportError):
    tzlocal = None
```

labrys:
@miigotu As of #141 it appears `tz.tzlocal` doesn't switch to `tzwinlocal` however `tz.tzwinlocal` currently detects windows correctly and otherwise is None.

Something like the following should work and allow you to use `tzlocal`
```python
tz.tzlocal = tz.tzlocal if tz.tzwinlocal is None else tz.tzwinlocal
```
`tz.tzwin(name)` does not have a direct correlation in `dateutil.tz` but `tz.gettz(name)` does automatically return `tzwin` objects on windows.

###### Other Windows issues:
-  #135 `ValueError: tzinfo.utcoffset() returned 1440; must be in -1439 .. 1439` **appears to be fixed in proposed fix #142**
- #143 `tzname()` inconsistent between `tzwin()` and `tzwinlocal()` and neither reflect actual DST.
- #144 `tzwin()` does not accept some values returned by `tzname()`

pganssle:
@miigotu Yes, @labrys is right that something like this is the best way to go. I think that in general it's not considered pythonic to aim for different behavior on different platforms, but rather to use a try-and-see duck-typing style approach.

In this case, I'd recommend:

```python
from dateutil import tz
tzlocal = tz.tzwinlocal or tz.tzlocal
```

I don't recommend redefining symbols within dateutil's namespace.

In the end, I'm wondering if maybe we should change the behavior of `tzlocal` such that it defaults to `tzwinlocal` on Windows. I suspect that this may be helpful in issue #57.

miigotu:
`In the end, I'm wondering if maybe we should change the behavior of tzlocal such that it defaults to tzwinlocal on Windows.`

Yes that was what I was hoping for.

My code above was asking about my code outside of dateutil =P. Does pretty much the same thing except I was importing tzwinlocal from tzwin instead of tz. Just was thinking it would be more reliable that way. I also need to make sure I get my usage/import of gettz right, or does that one always work on win?

pganssle:
@miigotu Well, one reason to keep them separate might be that they sorta do two different things. `tz.tzlocal()` gives you the time zone that *Python* retrieves for you (basically a `tzinfo` implementation that uses the timezone related features of the [`time`](https://docs.python.org/2/library/time.html) module, the other one gives you the time zone that *Windows* has.

If those things are different, it makes sense to have two different representations for them. If they're the same, then we should probably just drop `tzwinlocal` anyway, since it would be redundant.

As for using `sys.platform` to determine the platform - I have no information about how generally that applies and whether you want to use it. I think *in general* the approach should be that you don't pay attention to what platform you are actually *on*, and instead pay attention to the capabilities of that platform - so if you have to use `winreg` on Windows and something else on *nix, just try importing `winreg` - if it's defined, you're on a Windows-like system, if it's not, you're on another type of system.

I don't think the implementation of `gettz()` is platform specific, but it does return `tzlocal()` on Windows rather than `tzwinlocal()`. If you're not using `tzwinlocal()`-specific features, then `gettz()` or `tzlocal()` should probably be fine.

miigotu:
Yeah, all of this started because we were using tzlocal only and it had issues on windows with dst and some other things. If these changes make it work to just use tzlocal \o/

The reason we have to use dateutil is sort of critical to our app and relies on accurate results. We get "airdates" of episodes and "airtimes" of shows that are generally in the timezone that the show aired, and we also get the network and region of that show, which we have a mapping of what timezone each network is in. We then use this to convert the airdate and airtime to our local time through a hassle of a process. If the times aren't accurate after the conversion, we dont have the correct time to start looking for the episode so we might be up to 24 hours too early or late depending on how inaccurate our conversion was.

Will have to see what @labrys concludes, as I'm sure he has been playing with it a bit. I don't have windows here to test atm and have been a bit lazy to install a vm. I was hoping to get enough info to do it without the vm =P

Anyways, I'll spin up a vm and work it out. If I find any other issues I'll let you know.

labrys:
The most elegant solution would be for the library to do the detection for you, but still have the availability to directly call the individual versions if it was required in an edge case.  After all, dateutil is already performing a `try: import winreg` and duplicating that in each module that calls dateutil should be unnecessary.  

For SickRage's implementation this (along with #135, #142, and #143) would likely solve any remaining issues.

Having said that, I think there should be a single base class that handles all `tzinfo` functions (fairly naively like `tzutc`) in an OS independent manner.  For example, `tzname()` would just return the value in `self._name`; `utcosffset()` would return `self._dst_offset` if `self._dst_active` otherwise it would return `self._std_offset`.

Additionally, there should be OS specific classes derived from the base class that populates values with the correct OS specific information.  For the most part, I imagine, this could be accomplished through `__init__` and overriding property getters and setters, leaving the base class's methods largely unchanged.  These classes should all be able to reduce to the base class so pickling and unpickling can work in an OS independent manner.  This should also simplify equality checks.  Name issues like those in #144 would become a little more important for cross-platform compatibility, however after some initial checks with gettz(), it does not appear to have this problem, so maybe it can be leveraged as a solution.

Lastly there should be a public interface that should be called directly and which does the auto-selection of the OS specific class, but does not prevent direct access to the OS specific or base class, if needed.

I understand that this would be a fairly major rewrite, and you're the one to layout the roadmap for the dateutil library, but I think that the benefits would justify the effort.

I'm writing this on fairly little sleep :sleeping:, so I hope it came out coherently enough. :grinning:

PR #147, opened 2015-11-02
Fixed tzfile parsing for leapcnt-containing tzfile, with tests.

### LINKED ISSUE #146: tzfile does not correctly parse files due to its ignoring of leap data
The relevant part in the tzfile implementation does not read from the file if leapcnt is given:

https://github.com/dateutil/dateutil/blob/master/dateutil/tz/tz.py#L309-L312:

            # Not used, for now
            # if leapcnt:
            #    leap = struct.unpack(">%dl" % (leapcnt*2),
            #                         fileobj.read(leapcnt*8))

...which is wrong ;-) as you have to read (or seek forward in the file) this data even if you don't intend to use it. I suspect this hasn't been a problem in practice as the following data parts (isstd and isgmt information) basically aren't used anymore, after they've been set as ttinfo attributes.

hjoukl:
The previous implementation contained a commented code part which must not be
omitted if leapcnt is set or otherwise the following parts read from the file are
actually garbled, so this code has been uncommented. The existing
testLeapCountDecodesProperly test didn't really cover this at all which has been
fixed. Additional test methods for appropriate reading of the isstd and isgmt
information contained in the TZTest.NEW_YORK tzfile data have been added.
Fixes #146.


pganssle:
@hjoukl Any chance you can rebase this against current master? I'm ready to accept it, but the time zone tests have moved into `test_tz.py`.

hjoukl:
@pganssle Ok, will rebase against current master as soon as I get to it - might take some days.

hjoukl:
Having trouble with the rebase. Looks like commit 58dead1534b29b7872eec8c6c4bf86d6cb0f7b95 replaced all line endings (from previous LF to CRLF) on tz/tz.py.

Is this really intentional? 

I'm pretty much a git noob but https://help.github.com/articles/dealing-with-line-endings/ suggests that a per-project setting might help to avoid troubles here. At least if my assumption is true that only a commit from a windows box with a setting like git config --global core.autocrlf false could cause this, i.e. ending up with actual CRLF text files *in* the repo.

pganssle:
@hjoukl Yeah, CRLF is definitely not intentional. Probably some issue with my editor. The only file that should have CRLF endings is `NEWS`.

pganssle:
Give me a second and I'll switch the endings to LF for all files but NEWS.

pganssle:
@hjoukl Is that better?

hjoukl:
@pganssle yeah that looks way better.

pganssle:
@hjoukl Any progress on this? I'd like to get a new release out ASAP.

hjoukl:
Rebased, hopefully. Btw I threw out the pointless
`self.assertEqual(datetime(2007, 3, 31, 20, 12).tzname(), None)`, as testLeapCountDecodesProperly now actually does test something.

pganssle:
Yeah, I think that line was never necessary in the first place.... Merging. Thanks again for the contribution!

hjoukl:
Great, thanks to you for your work on this fine module.

PR #154, opened 2015-12-02
Fix doctest string for fuzzy_with_tokens.

gpoesia:
It seems the docstring was generated as if smart_defaults was True or something like this. At least now the code in the docstring returns year 2047.


pganssle:
@gpoesia Thanks for the contribution, looks good to me, so I'm merging.

PR #160, opened 2015-12-15
Add UNTIL to the RRULE __str__ representation

armicron:


dateutil/rrule.py
```diff
@@ -405,6 +405,9 @@ def __init__(self, freq, dtstart=None,
             until = datetime.datetime.fromordinal(until.toordinal())
         self._until = until
 
+        if count and until:
+            raise ValueError("either until or count may be provided")
```
pganssle: Hm.. This seems like it might cause some breaking changes if people have been using both `count` and `until`. I think for now we should change this to a [`DeprecationWarning`](https://docs.python.org/2/library/exceptions.html#exceptions.DeprecationWarning), and in future versions it'll be some sort of `ValueError`.

pganssle:
@armicron If you'd like to change the `ValueError` to a `DeprecationWarning`, I can merge this for 2.5.0.

armicron:
@pganssle done

pganssle:
Merging.

Issue #169, opened 2016-01-11
Release 2.5.0


pganssle:
I've been really negligent in getting this release out, I think it's time to wrap up whatever work needs to be done and get a release out ASAP. I'm creating this issue to discuss exactly what should be included in 2.5.0 and what should be pushed to later releases.

**2.5.0**:
-  tzwin fixes: ✓
    - PR #142 ✓
    - tzname() inconsistent between tzwin() and tzwinlocal() and neither reflect actual DST (#143) (PR #178) ✓
    - tzwinlocal's repr is incorrect (#148) (PR #179 -> #184) ✓
    - tzwin for the same timezone string does not compare equal (#151) (PR #180 -> #184) ✓
- tz fixes: ✓
    - dateutil does not accept a dt argument of None (#132) (PR #181) ✓
    - tz submodule not created by setup.py install (#165) ✓ 
- RRule
    - heapq performance enhancement - PR #136 ✓
- Parser:
    - Move `smart_defaults` back into a feature branch (#30) - in the time since adding this, I've been thinking we may want to implement this differently, and I don't want to get stuck maintaining API compatibility. (This should automatically fix #126 for the release, but that should also be updated on the feature branch). I think this should be re-implemented with a combination of #125 and breaking up `parse` into smaller, customizable functions. (PR #191) ✓
    - PR #130 (properly handle dates prior to year 100) ✓
    - Issue #78 (raise error on empty string) (PR #187) ✓
- Relativedelta:
    - Issue #40 - This is mostly implemented, I'll clean up and make PR. (PR #190) ✓
- build:
    - Either fix `updatezinfo` so that it doesn't fail so much or add some fallback logic to the build. (#176)  ✓ 

Things to push to 2.5.1:
- Issue #70 (API changes - don't want to rush)
- Issue #79 (I have this half-done, but it's a bit complicated to do so I need to add tests and I don't want to put things off)
- tzwin does not accept some values returned by tzname (#144)
- Issue #145 - This is apparently a big problem for at least one project, so we don't want it to persist through two releases. (Seems to have been mostly resolved on their side, so we can push it to 2.5.1)

@jarondl Anything critical I'm missing?

pganssle:
ping @labrys - Just want to check that I didn't miss anything on the tzwin stuff - does this look good to you?

labrys:
i think that covered it, ill take a closer look though, time permitting.

pganssle:
[...]

PR #170, opened 2016-01-11
#169 fix updatezinfo, so that it doesn't fail so much

bjamesv:
addresses issue #169 build issue.

adds simple retry loop with linear delay before retries

### DIFF
--- a/updatezinfo.py
+++ b/updatezinfo.py
@@ -8,6 +8,11 @@ from six.moves.urllib import request
 
 from dateutil.zoneinfo import rebuild
 
+from time import sleep
+
+MAX_RETRIES = 1 #iana FTP server is sometimes overloaded
+WAIT_SEC = 6
+
 METADATA_FILE = "zonefile_metadata.json"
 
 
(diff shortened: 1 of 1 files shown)


pganssle:
@bjamesv I appreciate the effort, but this is not how I'd like to solve the issue, because this script is not used *only* during the build, and I want it to fail properly if it fails during normal end-user use.

There are two possible strategies here:

1. Checking for failure in a build-time script (e.g. a change to the "install" portion of `[.travis.yml](https://github.com/dateutil/dateutil/blob/master/.travis.yml#L14) and [appveyor.yml](https://github.com/dateutil/dateutil/blob/master/appveyor.yml#L18) re-running the script a fixed number of times (these build scripts are *not* used by end-users and are only part of the CI).

2. Adding a mirror for the tz data that will *not* fail and falling back to the mirror on failure. I think this is a reasonable behavior to add as it is an actual added value for end users.

Likely the final solution will be a combination of the two.

pganssle:
Closing this in favor of #177, though I really do appreciate the effort.

Issue #176, opened 2016-02-11
Fix build failure problem


pganssle:
There is an intermittent build failure issue caused by some interaction with `updatezinfo.py` and the IANA time zone servers. It's not clear (to me) why we get so many failures, but it may have something to do with some rate-limiting that they haven't told us about.

I think the initial fix to this is to add a "wait-retry" loop to the Travis and Appveyor CI configurations. Longer term, I'm thinking we may want to separate out the timezone metadata and rehost the tzfiles at that repo, and the IANA server can be considered a mirror.

pganssle:
Stop-gap fix in #177.

PR #177, opened 2016-02-15
Retry logic for travis and appveyor

pganssle:
A partial mitigation of #176, adding retry logic to the `updatezinfo.py` script.

A next step (not necessarily to be done before the 2.5.0 release (#169) is to host our own mirror of all the tzdata files on HTTPS. I suspect part of the issue might be that we're downloading over ftp, which seems to be a bit finicky when I played with it using `ftplib`.


pganssle:
Also, it seems the `get-pip` script has moved, so that has been updated in the appveyor configuration.

PR #179, opened 2016-02-16
Add REPR functionality to tzwinlocal

pganssle:
Built off #178 for convenience, this adds `__repr__` functionality to `tzwinlocal()`. The 'StandardName' will be used for all repr (this does not affect tzname()).

Fixes #148

PR #178, opened 2016-02-16
Fix tzwinlocal issue. Add tzwin test context manager.

pganssle:
This resolves #143. There were two issues - one that the "nth weekday of the month" was calculated, the other is that, for whatever reason, in the "local" timezone key, the `SYSTEMTIME` struct has a different ordering than it has elsewhere.

Running the full test suite on Windows now requires setting `DATEUTIL_MAY_CHANGE_TIMEZONE` - I added a context manager that temporarily changes the timezone while running the tests, but I don't want that happening willy-nilly on some dev's system, so I added in an environment-variable-based guard.

PR #180, opened 2016-02-16
Add comparison functionality to tzwin objects

### LINKED ISSUE #151: tzwin for the same timezone string does not compare equal
Consider this:

```
>>> dateutil.tz.gettz('Eastern Standard Time') == dateutil.tz.gettz('Eastern Standard Time')
False
```

Other timezone classes such as tzfile or tzlocal compare equal:

```
>>> dateutil.tz.gettz('America/New_York') == dateutil.tz.gettz('America/New_York')
True
>>> dateutil.tz.gettz('America/New_York') == dateutil.tz.gettz('America/Halifax')
False
>>> dateutil.tz.tzlocal() == dateutil.tz.tzlocal()
True
```

Note that, as a consequence of this, the expression ```dateutil.tz.gettz('UTC') == dateutil.tz.gettz('UTC')``` will return different results on different platforms (False on Windows, True on Unices).

pganssle:
Fixes #151.

I will note that even if the tzwin objects are identical on all relevant characteristics (i.e. they represent the exact same offset at all times), they will compare unequal if their names are different. Not sure how I feel about this, but it seems like the best way to go in terms of least surprise.

This is built directly on top of #179 because I need the TzWin context manager for the tests.


pganssle:
@labrys If you want to test the tzwin updates, this is the best PR to use. It's cumulative with #178 and #179 because of how I handled the test environment (frankly, I should have probably just continued committing on the first PR). If you want to run the test suite and include some of the tzwinlocal tests, you need to set the environment variable `DATEUTIL_MAY_CHANGE_TZ` before running the tests.

I think this covers all the changes that I set out to fix for release 2.5.0.

labrys:
K will do. Hopefully I can test this thoroughly by this weekend.  Once again thanks for all your hard work.

pganssle:
Superceded by #184

PR #181, opened 2016-02-17
Add tzinfo compatibility with non-datetime times

### LINKED ISSUE #132: dateutil does not accept a dt argument of None
Per the Python Standard Library Documentation in regards to a `tzinfo` objects `tzname(...)`, `dst(...)`, and `utcoffset(...)` methods:

> These methods are called by a datetime or time object, in response to their methods of the same names.  A datetime object passes itself as the argument, and a time object passes None as the argument. A tzinfo subclass’s methods should therefore be prepared to accept a dt argument of None, or of class datetime.

pganssle:
Fixes #132. I don't think the tests here are compatible with the #180 stack because I changed how `dateutil.tz` is imported, so when those are merged, I'll rebase and clean up.

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 from datetime import datetime, timedelta
+from datetime import time as dt_time
 from six import BytesIO, StringIO
 
 import os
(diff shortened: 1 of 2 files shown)


pganssle:
Merging - will rebase the #180 stack once merged.

PR #183, opened 2016-02-20
Bump metadata version to 2.0, add support for tzdata mirrors.

pganssle:
I've created a mirror of the IANA database on a [github pages repository](https://dateutil.github.io/tzdata/). This is the first step in having a separate update channel between the code and the tzdata files. For now, having an HTTPS mirror of the tzdata tarball should help to further alleviate the CI failure issues.

### DIFF
--- a/updatezinfo.py
+++ b/updatezinfo.py
@@ -5,6 +5,7 @@ import json
 import io
 
 from six.moves.urllib import request
+from six.moves.urllib import error as urllib_error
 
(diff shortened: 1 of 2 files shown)


pganssle:
I should note that I'm also planning on maintaining a "latest version" of the zonefile_metadata.json file that will exist at a stable URL (currently https://dateutil.github.io/zoneinfo_metadata/zonefile_metadata.json), so if you want to upgrade to the latest version you can pull from there.

PR #184, opened 2016-02-20
Add REPR and comparison functionality to tzwinlocal

### LINKED ISSUE #148: tzwinlocal's repr is incorrect
The representation of tzwinlocal is inconsistent with tzwin and tzlocal.
````python
import dateutil.tz as tz

print(tz.tzlocal())  # tzlocal()
print(tz.tzwin('Eastern Standard Time'))  # tzwin('Eastern Standard Time'
print(tz.tzwinlocal())  # <dateutil.tzwin.tzwinlocal object at 0x024C7D50>
```
Also inconsistent after proposed fix #142

### LINKED ISSUE #151: tzwin for the same timezone string does not compare equal

pganssle:
This is a combination of PRs #179 and #180 because of some funkiness with the rebase that was muddying up the git history. It fixes Issues #148 and #151.

#### PR #179 - Add REPR functionality to tzwinlocal
Built off #178 for convenience, this adds `__repr__` functionality to `tzwinlocal()`. The 'StandardName' will be used for all repr (this does not affect tzname()).

Fixes #148.

#### PR #180 - Add comparison functionality to tzwin objects
I will note that even if the tzwin objects are identical on all relevant characteristics (i.e. they represent the exact same offset at all times), they will compare unequal if their names are different. Not sure how I feel about this, but it seems like the best way to go in terms of least surprise.

Fixes #151.

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -353,6 +353,9 @@ class TZTest(unittest.TestCase):
         # bug 892569
         str(tz.gettz('UTC'))
 
+    def testGetTzEquality(self):
+        self.assertEqual(tz.gettz('UTC'), tz.gettz('UTC'))
+
     def testBrokenIsDstHandling(self):
         # tzrange._isdst() was using a date() rather than a datetime().
         # Issue reported by Lennart Regebro.
@@ -468,6 +471,37 @@ class TzWinTest(unittest.TestCase):
         for t_date, expected in transition_dates:
             self.assertEqual(t_date.replace(tzinfo=tw).tzname(), expected)
 
+    def testTzWinEquality(self):
+        # https://github.com/dateutil/dateutil/issues/151
+        tzwin_names = ('Eastern Standard Time',
+                       'West Pacific Standard Time',
+                       'Yakutsk Standard Time',
+                       'Iran Standard Time',
+                       'UTC')
+
+        for tzwin_name in tzwin_names:
+            # Get two different instances to compare
+            tw1 = tz.tzwin(tzwin_name)
+            tw2 = tz.tzwin(tzwin_name)
+
+            self.assertEqual(tw1, tw2)
+
+    def testTzWinInequality(self):
+        # https://github.com/dateutil/dateutil/issues/151
+        # Note these last two currently differ only in their name.
+        tzwin_names = (('Eastern Standard Time', 'Yakutsk Standard Time'),
+                       ('Greenwich Standard Time', 'GMT Standard Time'),
+                       ('GMT Standard Time', 'UTC'),
+                       ('E. South America Standard Time',
+                        'Argentina Standard Time'))
+
+        for tzwn1, tzwn2 in tzwin_names:
+            # Get two different instances to compare            
+            tw1 = tz.tzwin(tzwn1)
+            tw2 = tz.tzwin(tzwn2)
+
+            self.assertNotEqual(tw1, tw2)
+
     @unittest.skipUnless(TZWinContext.tz_change_allowed(),
         'Skipping unless tz changes are allowed.')
     def testTzwinLocalName(self):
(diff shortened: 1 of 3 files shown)


pganssle:
@labrys Thanks again for your help identifying and testing all these Windows time zone issues. I think this really makes some significant improvements in this space.

pganssle:
Codecov diff status failure jumped the gun when the travis results came in and is not updating now that it is updated. I'm disabling the diff status check if it's going to cause these problems.

labrys:
Missing single quotes around the `tzwinlocal` repr timezone name
```python
>>> print(tz.tzwinlocal())
tzwinlocal(Eastern Standard Time)
>>> print(tz.tzstr('Eastern Standard Time'))
tzstr('Eastern Standard Time')
>>> print(tz.tzwin('Eastern Standard Time'))
tzwin('Eastern Standard Time')
```

labrys:
Also inconsistent with `tzlocal` repr:
```python
>>> print(tz.tzlocal())
tzlocal()
>>> print(tz.tzwinlocal())
tzwinlocal(Eastern Standard Time)
```

pganssle:
@labrys Good point, I'll fix the quotes thing. I think I'll leave it inconsistent with tzlocal(). Ideally, we'd go the other way, tzlocal() should say "tzlocal('Eastern Standard Time')" or something like that, but I'm not sure how easy it is to figure that sort of thing out.

I think that once a `tzwinlocal()` object is instantiated, its time zone is fixed (even if the local time zone changes later). I haven't looked into it, but I assume that's also the case with `tzlocal()`.

labrys:
@pganssle My problem with the `tzwinlocal` repr is that per the python standard docs on [repr](https://docs.python.org/2/library/functions.html#func-repr):
> For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval(), otherwise the representation is a string enclosed in angle brackets that contains the name of the type of the object together with additional information often including the name and address of the object.

Thus it should be `tzwinlocal` or possibly return an equivalent object repr such as `tzstr('Eastern Standard Time')` which may make more sense as it would create an equivalent time zone regardless of the local time zone at the time of recreation of the object.  Although this would only work if you could assert that the `tzwinlocal` name and `tzstr` objects would be equal.

pganssle:
Hm, that's a fair point. OK, for now we'll go with `tzwinlocal()`. We can come back to this later - it seems to me that including the time zone name in `str()` (but omitting it from `repr()`) might be a good compromise, since at as it is now `tzwinlocal()` doesn't actually give you any information about *which* time zone you're in, and `tzwinlocal()` won't produce the same object if you call it after a change in the time zone.

labrys:
Also on a related note:
```python
>>> now = datetime.now()
>>> tz.tzlocal().tzname(now) == tz.tzwinlocal().tzname(now)
True
>>> tz.tzlocal() == tz.tzwinlocal()
False
```
Those both represent `Eastern Standard Time` so both should expect to evaluate equal.

pganssle:
@labrys I think that that last one is acceptable for now. `tzlocal()` is a wrapper around the `time` functions related to time zone, while `tzwinlocal()` is pulling its information from the Windows registry keys.

I think the concept of "equality" of time zone objects is a bit fuzzy. While the two things do represent the same time zone, one may be richer than the other and I think it's worth giving careful thought as to how the two might diverge (maybe the answer is 'not at all', but I dunno).

I'd rather err on the side of having the two compare not equal because I feel like that is less likely to create lasting API implications (e.g. people won't rely on two things comparing equal if they represent the same value even if they are of different types).

labrys:
This goes back to why I feel they should all be based on a single base class (other than tzinfo) with common attributes defining that timezone (e.g. offset).  Then typecasting to the base class can compare equal without having to manually compare attributes within.
e.g.:
```python
>>> tzw = tzwinlocal()
>>> tzl = tzlocal()
>>> tzbase(tzw) == tzbase(tzl)  # since offset, dst, and name are equal
True
```

pganssle:
Yeah, I think this is a good idea, but it's a pretty big change and isn't crucial for the 2.5.0 release.

Issue #40, opened 2015-02-23
Relativedelta constructed from fractions of hours... (Feature request)


ghost:
I would like to request feature to allow relativedelta construction from e.g. hour fractions.
In some business applications I met the practice to store total hours count represented as float. In one even as total days count (represented by float). Now I need to integrate with them.
thx

pganssle:
@babbbrak I'm not sure how much bearing it would have on the implementation of this sort of thing, but how would you recommend dealing with a situation where a smaller value is specified along with a fractional value? E.g. 1.48 days = 1 day, 11 hours, 31 minutes, 12 seconds. If I specify: `relativedelta(days=1.48, seconds=4)`, should the `seconds` keyword override the days, vice-versa, or should an error be thrown?

I'm also a little concerned with the possibility that people might not be paying attention to their data types and pass a float that has a rounding error in it, causing unexpected behavior in the implementation.

One possible suggestion I have would be to add a submodule for float-to-ordinal and ordinal-to-float conversions. I think this might have wider uses (there are a few places this comes up in `parser`, for example), and it would let you choose your own default fallback positions if you pass the outputs to a `relativedelta`.

ryanss:
I believe you can close this issue as this is already how `relativedelta` works:

```python
>>> datetime(2015, 1, 1) + relativedelta(days=1.5)
datetime.datetime(2015, 1, 2, 12, 0)
>>> datetime(2015, 1, 1) + relativedelta(hours=1.5)
datetime.datetime(2015, 1, 1, 1, 30)
>>> datetime(2015, 1, 1) + relativedelta(hours=1.5, minutes=4)
datetime.datetime(2015, 1, 1, 1, 34)
```

@pganssle In your example `relativedelta(days=1.48, seconds=4)` nothing is overwritten, the seconds from the partial `days` and the `seconds` parameters are added together:

```python
>>> datetime(2015, 1, 1) + relativedelta(days=1.48, seconds=4)
datetime.datetime(2015, 1, 2, 11, 31, 16)
```

pganssle:
@ryanss Good catch, but I think that the support actually should be made more explicit. Right now, it's basically a coincidence that this works, because `relativedelta` doesn't explicitly check if it's been given a float or an integer, and all relative information at the level of a day or smaller are first cast to `timedelta` during addition operations. When a `timedelta` object is passed a fractional argument, it converts the remaining fraction to seconds (which is added to the existing seconds argument). So:

[...]

PR #190, opened 2016-02-24
Implement explicit fractional support for relativedelta objects.

pganssle:
Fixes (mostly) #40.

* Implement explicit fractional support for relative components.
* Add fractions to the repr.
* Deprecate fractional support for absolute components.
* Added .normalize() method to provide integer-only relativedeltas.


pganssle:
Merging.

Issue #125, opened 2015-09-11
Parser (or new function in parser module) should be able to output the parsed format


pganssle:
There's a good amount of overhead in parsing a date compared to calling `strptime` or something of that nature. It seems that at least pandas calls `dateutil.parser.parse` in a tight loop. In a lot of cases, what happens is that you have a column of  say 1000 dates that you want to turn into datetime objects, you don't know *what* the format is, but you know that it's always the same. Right now, the parser has to infer the type *every time* it parses the date. It would be nice if it could emit something that could either be parsed with `strptime()` (if that's possible), or we could provide a fast function that takes a timestr and a format (as output by the parser).

I think this could also provide additional flexibility to the parser module, since a lot of issues seem to arise from the fact that the parser itself is the only thing that knows what the format is, and if you want to do something unusual with that information, you have to modify the parsing function itself.

I think this can go hand-in-hand with #123 as part of an effort to make the parser a bit more modular and provided better public interfaces for dealing with date parsing.

PR #191, opened 2016-02-25
Rolled back smart_defaults and fallback_on_invalid_dates modifications to parser

### LINKED ISSUE #126: New smart defaults feature breaks python 3.0 - 3.4
See https://github.com/dateutil/dateutil/commit/1652d81f0d982f9f73aea4082c15584b22bacf36#diff-ed798069162e7ab7db9a45da6c4a076fR488

`iteritems` doesn't work in python 3.0 - 3.4, see [PEP 0469](http://legacy.python.org/dev/peps/pep-0469). It might work again in 3.5, but as it stands now it breaks in 3.4.

pganssle:
Rolling back `smart_defaults` and `fallback_on_invalid_date` introduced in PR #30, My explanation for this is found in Issue #169:

> In the time since adding this, I've been thinking we may want to implement this differently, and I don't want to get stuck maintaining API compatibility. (This should automatically fix #126 for the release, but that should also be updated on the feature branch). I think this should be re-implemented with a combination of #125 and breaking up `parse` into smaller, customizable functions.

Since these features have not been included in a public release yet, hopefully it will not break anyone's dependencies here.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -13,22 +13,8 @@ of a date/time stamp is omitted, the following rules are applied:
 
 If any other elements are missing, they are taken from the
 :class:`datetime.datetime` object passed to the parameter ``default``. If this
-results in a day number exceeding the valid number of days per month, one can
-fall back to the last day of the month by setting ``fallback_on_invalid_day``
-parameter to ``True``.
-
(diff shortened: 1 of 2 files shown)


pganssle:
Merging.

PR #192, opened 2016-02-28
Prepare 2.5.0 release

### LINKED ISSUE #169: Release 2.5.0

pganssle:
Resolves #169.

I also added the zoneinfo tarball back into the repo. I think we decided some time ago to include it, but for whatever reason it was removed during one of the pull requests.


pganssle:
Ugh. How did I miss that the `test` directory is missing in the source releases *again*? That will have to be fixed in 2.5.1 on PyPi, but I'll change the source distributions here.

PR #195, opened 2016-02-29
NEWS: convert to utf8, remove dos line endings

keszybz:
Makes it nicer to view in the terminal.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -80,7 +80,7 @@ Version 2.5.0
   were failing due to connection issues when downloading the IANA timezone
   files. In addition to adding our own mirror for the files (gh pr #183), the
   download is now retried a number of times (with a delay) (gh pr #177)
-- Many failing doctests were fixed by @moreati. (gh pr #120)  
+- Many failing doctests were fixed by @moreati. (gh pr #120)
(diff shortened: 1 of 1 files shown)


pganssle:
I wonder why it even had dos line endings in the first place? I've been editing it on Linux.

keszybz:
I think that most editors are smart enough to not display and preserve CRLF line endings.

pganssle:
@jarondl Any thoughts on this? I don't think it matters that much, and the only reservation I have is that I think that on Windows, Notepad doesn't recognize LF as a linebreak at all, in which case it would become completely unreadable, whereas if I `cat` the file, it just doubles the linebreaks. Since most editors will figure it out either way, and the one common editor that's going to fuck it up will fuck it up *really badly* if we don't use DOS line endings, I'd say leave it DOS for just this file.

I'm fine with switching to UTF-8 encoding.

labrys:
@pganssle Notepad will definitely go stupid without CRLF.

keszybz:
I mostly care about the encoding, line-endings are less important.

pganssle:
@keszybz OK, do you want to switch the line endings back to CRLF and I'll accept the PR?

Thanks for your contribution!

PR #202, opened 2016-03-04
Resolve recursive import.

htgoebel:
This solves https://github.com/pyinstaller/pyinstaller/issues/1848.

### DIFF
--- a/dateutil/tz/__init__.py
+++ b/dateutil/tz/__init__.py
@@ -1,20 +1,4 @@
 from .tz import *
-from six import PY3
 
 __all__ = ["tzutc", "tzoffset", "tzlocal", "tzfile", "tzrange",
            "tzstr", "tzical", "tzwin", "tzwinlocal", "gettz"]
-
-def tzname_in_python2(namefunc):
-    """Change unicode output into bytestrings in Python 2
-
-    tzname() API changed in Python 3. It used to return bytes, but was changed
-    to unicode strings
-    """
-    def adjust_encoding(*args, **kwargs):
-        name = namefunc(*args, **kwargs)
-        if name is not None and not PY3:
-            name = name.encode()
-
-        return name
-
-    return adjust_encoding
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -14,7 +14,6 @@ import sys
 import os
 
 from six import string_types, PY3
-from .__init__ import tzname_in_python2
 
 try:
     from .win import tzwin, tzwinlocal
@@ -27,6 +26,22 @@ rrule = None
 
 ZERO = datetime.timedelta(0)
 EPOCHORDINAL = datetime.datetime.utcfromtimestamp(0).toordinal()
+
+def tzname_in_python2(namefunc):
+    """Change unicode output into bytestrings in Python 2
+
+    tzname() API changed in Python 3. It used to return bytes, but was changed
+    to unicode strings
+    """
+    def adjust_encoding(*args, **kwargs):
+        name = namefunc(*args, **kwargs)
+        if name is not None and not PY3:
+            name = name.encode()
+
+        return name
+
+    return adjust_encoding
+
 
 class tzutc(datetime.tzinfo):
 


pganssle:
This is failing on Windows. Failing in a weird way, mind you, but we moved the tzname_in_python2 to .__init__.py for a reason, as it is used in both `tz` and `tzwin`.

If you can explain how I can reproduce the error, I'd be happy to look into it.

htgoebel:
To reproduce the error, simply try to freeze a simple program with PyInstaller, see https://github.com/pyinstaller/pyinstaller/issues/1848.

The test-failure on Windows seem not to be related to my change, as they say:
```
  File "c:\python27\lib\unittest\loader.py", line 100, in loadTestsFromName
    parent, obj = obj, getattr(obj, part)
AttributeError: 'module' object has no attribute 'test_tz'
```

pganssle:
Well, considering that you [broke the tzwin module](https://github.com/htgoebel/dateutil/blob/master/dateutil/tz/win.py#L14), I can't imagine that the test failure on Windows is unrelated to your changes. The mutual inclusion of `tzname_in_python2` across both files was the whole reason that `tz` was rearranged in this way.

I'm not entirely sure what the problem is with the way this is being imported, though. I'll try and see what's happening with pyinstaller.

htgoebel:
So what you are doing here is a multi-recursive import.:
```
 tz
 +-- tz.tz [1]
      +-- tz.__init__ [2]
          +-- tz.tz  [3]
      +-- tz.win  [4]
          +-- tz.__init__ [5]
````
Explanation:
1. Here a new module name `tz.__init__` is created, put into `sys.modules` and it's code is executed.
2.Here a new module name `tz.__init__` is created. This is *not* the same as `tz`, even if both share the same code.
3. When `tz.__init__` tries to import `tz.tz`, `tz.tz` is found in `sys.modules`, so `tz.__init_` happily takes. Sidenode: (Since `tz.tz` is not imported completely, `tz.__init__` will see only a few identifiers - which in this case does not raise any issue since `tzname_in_python2` does not require any of the missing identifiers.)
4. `tz.tz` then continues and import `tz.win`
5. When `tz.win` imports `tz.__init__`, this is found in `sys.modules` and recursion stops here.

Practically, this works, but it is wrong: It creates a *new* module `tz.__init__`. PyInstaller takes modules names ending with `__init__` as the packages "main" module, and thus fails.

Now, after my change, `tz.win` still wants to import `tz.__init__.tzname_in_python2`, which now does not exist. So you are right, my change broke `tz.win`.

But as you can see in my explanation, the original code is wrong. `from .__init__ import ...` *is not a valid import*, since `__init__` is not to be used as module name.

**So, what can be a solution?**

Either
1. Move the whole contents of `tz.tz` into `tz` - which is nearly empty anyway.
2. Move tzname_in_python2 into some extra module, sitting "below" both `tz.tz` and `tz.win`. This module could be named `_compat`, or `_util`, with the leading underscore marking it as private.

htgoebel:
Addendum: I just see, that some of the windows test runs fail in `loadTestsFromName`, but others indeed fail with `ImportError: cannot import name 'tzname_in_python2'`. When I first looked, I picked one of the first type.

pganssle:
Yes, that makes sense and I agree with your assessment. I appreciate you taking the time to explain the details.

I think Option 2 is the right thing to do. I'm hoping to better unify the `tz` and `tzwin` modules in the future, and that may well involve more shared code between the two.

pganssle:
Closing this PR, I'll rearrange the package as discussed above.

htgoebel:
That's fine for me,then :-)

pganssle:
@htgoebel By the way, as a practical matter, do you have a suggestion for how to test this situation? Given that it works fine in practice, it's probably a good idea to add a unit test to prevent this sort of thing happening in the future.

htgoebel:
I'd suggest grepping the source for `from\s+.*__init__\s+import`. If you want a perfect test, you'd need to parse all files (using ast) and walk the parse-tree.

codewarrior0:
> So what you are doing here is a multi-recursive import

:+1: Informative! I never would have expected that was what is happening.

PR #208, opened 2016-03-07
Changed key retrieval to unicode string.

pganssle:
This should address the reported issue from Sickrage/sickrage-issues#1098, reported by @labrys.

I have only been able to test this on a US locale so far, and I have no idea how to set up Appveyor to switch to a locale with unicode time zones. I'd love to get that set up, though, since we've had these testability issues before.

### DIFF
--- a/dateutil/tz/win.py
+++ b/dateutil/tz/win.py
@@ -3,6 +3,7 @@ import datetime
 import struct
 
 from six.moves import winreg
+from six import text_type
 
 try:
     import ctypes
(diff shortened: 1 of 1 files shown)


pganssle:
I managed to test this, and it turns out that this is actually a bug upstream in Python 2.7. `_winreg` does not properly know how to handle unicode strings for their keys, as it tries to convert them to ANSI on a call to `OpenKey`.

Apparently there's a package for this [`winreg_unicode`](https://pypi.python.org/pypi/winreg_unicode), but I don't think it's worth adding in another dependency.

@labrys I'm going to merge this PR anyway, but I'm going to open a new issue to leave open to see if there's a way to fix this.

Issue #149, opened 2015-11-10
rrule bug when start date is month end and frequency is monthly


broncos168:
Please bear me. I am new to Python and this is my first time post in github. Thanks.

Code snippet to demo the bug:

```python
from datetime import date
from dateutil import rrule
from dateutil import relativedelta

start_date=date(2014, 12, 31)  # date(2014, 12, 29/30/31) 
end_date=date(2015, 12, 31)
freq = rrule.MONTHLY
interval = 1
dts = list(rrule.rrule(freq, interval=interval, dtstart=start_date, until=end_date))
for dt in dts:
    print(dt.date().isoformat())
```

In this edge case, outputs are not monthly
```
2014-12-31
2015-01-31
2015-03-31
2015-05-31
2015-07-31
2015-08-31
2015-10-31
2015-12-31
```

```python
d1=date(2015, 2, 28)
d2=date(2015, 5, 31)
rel_delta1 = relativedelta.relativedelta(d2, d1)
rel_delta2 = relativedelta.relativedelta(d1, d2)
print(rel_delta1, rel_delta2)
```

Output:
```python
relativedelta(months=+3, days=+3) relativedelta(months=-3)
rel_delta1 is not the negative of rel_delta2, vice verse
```

phep:
Hi,
While the behaviour is admittedly quite disconcerting, I'm not quite sure this is a valid bug since RFC5545 says:
```
Recurrence rules may generate recurrence instances with an invalid
date (e.g., February 30) or nonexistent local time (e.g., 1:30 AM
on a day where the local time is moved forward by an hour at 1:00
AM).  Such recurrence instances MUST be ignored and MUST NOT be
counted as part of the recurrence set.
```
The current behaviour respects the RFC.

Maybe there might be some parameter added to let the user decides which behaviour s/he's expecting (be strict and exclude, or floor/ceil the occurence).
I'm just discovering dateutil so this might have already been implemented somehow.

pganssle:
@phep Good call. I don't think it's necessary to add any sort of user option for this. If I recall correctly, if you want to get the last day of the month, you can use `bymonth=-1`. I'll double check that, though.

phep:
Yes, it is true that one can use rrules to express offset from start or end of month, but this is not necessarily semantically equivalent to the given example.

[...]

PR #213, opened 2016-03-10
Adjustments to rrule documentation

### LINKED ISSUE #105: rrule until keyword documentation confusing
Migrated from [lp:1483977](https://bugs.launchpad.net/dateutil/+bug/1483977).

[Francisco Saldana](https://launchpad.net/~i-francisco) wrote at 2015-08-12 05:31:45:
> Documentation for python-dateutil rrule https://labix.org/python-dateutil#head-470fa22b2db72000d7abe698a5783a46b0731b57

> States:
>> If given, this must be a datetime instance, that will specify the limit of the recurrence. If a recurrence instance happens to be the same as the datetime instance given in the until keyword, this will be the last occurrence.

> How can the recurrence instance (a rrule object) be the same as the datetime instance? Is it referring to the datetime instance given in dtstart?

@FrankSalad - Possibly this was you?

pganssle:
Resolves #105 and #149.

I've added an explicit note to the `rrule` documentation indicating the possibly surprising behavior of rrules where `dtstart` is at the end of the month. I've also (hopefully) clarified the `until` parameter's documentation and added a note about the deprecation of using `count` along with `until`.

When I was testing the new rrule doctest, I also noticed a few parser doctests that were failing. They fail at least as far back as version 2.2, so I'm assuming the behavior of `parser` changed between the time the docs were written and 2.2. Rather than try and change the behavior back (not sure why we'd want to - the behavior exhibited in the docs was kinda surprising), I just updated the doctests.

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -328,6 +328,29 @@ class rrule(rrulebase):
     Where freq must be one of YEARLY, MONTHLY, WEEKLY, DAILY, HOURLY, MINUTELY,
     or SECONDLY.
 
+    .. note::
+        Per RFC section 3.3.10, recurrence instances falling on invalid dates
+        and times are ignored rather than coerced:
+
+            Recurrence rules may generate recurrence instances with an invalid
+            date (e.g., February 30) or nonexistent local time (e.g., 1:30 AM
+            on a day where the local time is moved forward by an hour at 1:00
+            AM).  Such recurrence instances MUST be ignored and MUST NOT be
+            counted as part of the recurrence set.
+
+        This can lead to possibly surprising behavior when, for example, the
+        start date occurs at the end of the month:
+
+        >>> from dateutil.rrule import rrule, MONTHLY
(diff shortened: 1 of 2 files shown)


pganssle:
I should mention that @phep provided the answer for how to resolve #149.

Itsmediego:
Why it miss a date??


import datetime
from dateutil.rrule import rrule, MONTHLY

start = datetime.datetime(2009, 01,29)
end = datetime.datetime(2009, 04, 01)

for diff in [dt for dt in rrule(MONTHLY, dtstart=start, until=end)]:
	print diff


2009-01-29 00:00:00
2009-03-29 00:00:00

**SHOULD I HAVE 2009-02-28 00:00:00 ??**

pganssle:
@Itsmediego See the discussion in #149, this is the expected behavior of the RFC. I believe that implementing the RFC 7529 extensions (see #285) will give a more satisfactory result here.

PR #215, opened 2016-03-11
Refactor relativedelta a bit, improve test coverage.

pganssle:
I've improved the test coverage on the `relativedelta` module, and while I was in there I did some refactoring to make the code a bit more readable.

I should note that I did not add any tests for the `weekday` objects, because that already has good test coverage in the `rrule` module, and will eventually be moved into a common file to prevent code duplication.

I also deliberately didn't add a test for [this `TypeError`](https://github.com/dateutil/dateutil/blob/master/dateutil/relativedelta.py#L131) - I think it is considered an antipattern to explicitly check inheritance like that (rather than duck-typing), so I didn't want to "endorse" this behavior that may go away.

### DIFF
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -176,16 +176,10 @@ class relativedelta(object):
                 compare = operator.lt
                 increment = -1
 
-            ii = 0
             while compare(dt1, dtm):
                 months += increment
                 self._set_months(months)
                 dtm = self.__radd__(dt2)
-                ii += 1
-                if ii > 1:
-                    print('ii == {}'.format(ii))
-                    print(dt1)
-                    print(dt2)
 
             # Get the timedelta between the "months-adjusted" date and dt1
             delta = dt1 - dtm
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -2,6 +2,9 @@
 import datetime
 import calendar
 
+import operator
+from math import copysign
+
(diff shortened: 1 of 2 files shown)


dateutil/relativedelta.py
```diff
@@ -157,19 +160,34 @@ def __init__(self, dt1=None, dt2=None,
...
             else:
-                while dt1 < dtm:
-                    months -= 1
-                    self._set_months(months)
-                    dtm = self.__radd__(dt2)
+                compare = operator.lt
+                increment = -1
+
+            ii = 0
+            while compare(dt1, dtm):
```
pganssle: I'm not sure that this "while" loop is necessary. I accidentally checked in the "print if it ever reaches the second iteration" code, but I ran the following (rather time-consuming) script:

```python
from dateutil.relativedelta import *

dt1 = datetime(2000, 1, 1)
dtend = datetime(2028, 1, 1)
td = timedelta(days=1)

while dt1 < dtend:
    dt2 = dt1
    while dt2 < dtend:
        dt2 += td
        relativedelta(dt1, dt2)
    dt1 += td
```

And I have yet to find a single pair of datetimes which reaches a second iteration, so I think that this while loop is unnecessary. I'm not entirely sure why it's there in the first place.

pganssle: I'll note that this `while` loop is present in the first commit in the repo, so I'm not sure why it was added.

PR #218, opened 2016-03-14
Change all line endings to LF.

pganssle:
There's been some inconsistency in the line endings for whatever reason. This should fix that.


pganssle:
Going ahead with the merge even though the CI isn't done, since there were no actual substantive changes.

PR #224, opened 2016-03-18
Use .localize() whenever possible

lbolla:
When using pytz's timezone's, it's best to use `localize()` rathen than `replace(tzinfo)`. See http://pytz.sourceforge.net/#localized-times-and-date-arithmetic

This solves issues like the one described here:
http://stackoverflow.com/questions/24856643/unexpected-results-converting-timezones-in-python

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -594,13 +594,13 @@ class parser(object):
                 else:
                     raise ValueError("Offset must be tzinfo subclass, "
                                      "tz string, or int offset.")
-                ret = ret.replace(tzinfo=tzinfo)
+                ret = _localize(ret, tzinfo)
             elif res.tzname and res.tzname in time.tzname:
-                ret = ret.replace(tzinfo=tz.tzlocal())
+                ret = _localize(ret, tz.tzlocal())
             elif res.tzoffset == 0:
-                ret = ret.replace(tzinfo=tz.tzutc())
+                ret = _localize(ret, tz.tzutc())
             elif res.tzoffset:
-                ret = ret.replace(tzinfo=tz.tzoffset(res.tzname, res.tzoffset))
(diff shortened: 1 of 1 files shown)


pganssle:
Thanks for the work on this.

I'll look into it more closely, but I'm strongly inclined _not_ to do this. `pytz` is a separate project, and there's no reason for us to program against their API. If you want to use `pytz` time zones, you should probably just localize them yourself after they've been parsed.

I think I'd be much more amenable to a solution taking advantage of the fact that `tzinfos` can be callable, perhaps one where tzinfos first tries calling with 3 arguments (tzname, tzoffset, parsed_date), and if that fails, tries calling it with 2 arguments (tzname, tzoffset). That way, you are free to incorporate `localize` into a `tzinfos` function.

That said, I'm hoping the improvements I suggested in #125 would alleviate this at least to some degree, since the end user would have access to the timezone object and all the date components before it returns a date, and they would be free to inject timezone localization as they see fit.

lbolla:
Using `.__call__()` rather than `.localize()` seems arbitrary, too. Anyway, thanks a lot for your suggestion about calling `localize()` on the parsed date: it seems to do the trick for my use case. Cheers!

pganssle:
@lbolla Well, the point isn't that you'd call the `tzinfo` object, but the datetime `tzinfos` protocol already allows for `tzinfos` to be an arbitrary callable function that returns a `tzinfo` instance.

My proposal above was to extend the interface such that `datetuil` would pass not only the parsed timezone information, but also the parsed _date_ to `tzinfos` as an optional 3rd argument, thus allowing end-users to pull whatever time zones they want.

That said, I don't see any particular need for this, and I am not yet convinced that `pytz`'s approach of eagerly resolving the DST status of a given time zone is warranted, and I'm not sure it's something that we should be building in explicit support for in dateutil.

lbolla:
Thanks for the clarification!

Issue #128, opened 2015-09-19
Wrong dst() value for time in Portugal tz


abalkin:
Pick a time in the Summer of 1996:
```
>>> t = datetime(1996, 6, 1, tzinfo=tz.gettz('Portugal'))
```
the time zone abbreviation is computed correctly:
```
>>> t.tzname()
'WEST'
```
([WEST](http://www.timeanddate.com/time/zones/west) = Western European Summer Time).

However, 

```
>>> t.dst()
datetime.timedelta(0)
```

and as a consequence:

```
>>> t.timetuple().tm_isdst
0
>>> from time import strftime
>>> strftime('%Z', t.timetuple())
'WET'
```

The problem is that `t.dst()` is computed as a difference between `t.utcoffset()` and the offset of the last standard time before `t`, but Portugal switched to `WEST` in the Spring of 1996 from `CET`, not `WET` and therefore UTC offset did not change, while isdst did.

pganssle:
@abalkin Is there actually enough information in the `tzfile` to solve this problem other than by using the dubious property of changes in the abbreviation? I know the `tzdata` inputs have a "SAVE" column, but that seems to be removed in the `zic`.

I suppose that the "next best solution" is to make two assumptions:

1. There is no real zone where the DST offset is == 0
2. There are no zones where both the DST offset and the base zone changed in  *both* the STD->DST transition *and* the DST->STD transition.

In which case, if we know it's DST from the `isdst` flag, and the calculated `dst()` is 0, then we should calculate the DST offset from the difference in offset with the _next_ standard time.

This still leaves us open to potential zones where the base zone change goes the other way (making the `dst()` value 2 hours instead of 1).

abalkin:
>  Is there actually enough information in the tzfile to solve this problem other than by using the dubious property of changes in the abbreviation?

Yes, moreover, lossless UTC-local conversion can be implemented using POSIX localtime_r API and without mktime or equivalent.  See python/cpython#20.  (Note that only pure python implementation is ready at this point.)

abalkin:
@pganssle - sorry, I thought I was responding on #112.  In this case, you are right: the necessary information is only available in raw files and is not preserved by zic.

abalkin:
@pganssle - please see my [Computing .dst() as a timedelta](https://mail.python.org/pipermail/datetime-sig/2015-September/000882.html) post on the Datetime-SIG mailing list.

pganssle:
[...]

Issue #112, opened 2015-08-24
utc -> local timezone conversion fails during a DST transition


zed:
[The bug was reported on launchpad in 2012.](https://bugs.launchpad.net/dateutil/+bug/944123) Let me know if it has been migrated to github already. I haven't found it. Here's a [standalone runnable test][1].

Input: `2011-11-06 05:30:00 UTC+0000`, `America/Toronto`
Expected: `2011-11-06 01:30:00 EDT-0400` 
Got: `2011-11-06 01:30:00 EST-0500`

The issue is reproducible with the latest `dateutil` version from PyPI.

Is it documented that the conversion should fail?
Does `dateutil` usage look ok in [the test][1]?

[1]: https://gist.github.com/zed/3838828

pganssle:
Hm. Worth tracking down. I think, by the way, that `tz.gettz()` is the proper way to get timezone files. I'm not sure why this functionality is duplicated in `zoneinfo`.

pganssle:
@zed FYI, the difference between the `zoneinfo` and `tz` `gettz()` functions is that the `zoneinfo` one is specifically used for getting the timezone from `dateutil`'s zoneinfo tarball.

pganssle:
I've also looked into this actual problem a bit, and I think the problem is that we do not override the `fromutc()` function to take transition times into account. One problem with just adding `fromutc()` that the transition times are [stored as wall times](https://github.com/dateutil/dateutil/blob/master/dateutil/tz/tz.py#L397) in the `tzfile` object, so we may need to figure out how best to handle the unit conversions.

pganssle:
On second thought, according to [this comment](https://github.com/python/cpython/blob/8d1d7e6816753248768e4cc1c0370221814e9cf1/Lib/datetime.py#L1954) (which, frankly, I'm having a bit of trouble following, since I'm not really sure why the notation makes a distinction between `y.n` and `x.n`), overriding `fromutc()` is generally unnecessary, and looking at this more closely, I see what the problem is.

This relates to the fact that `pytz` eagerly evaluates the DST status of the timezone, which allows them to handle ambiguous date representations for the user. In a sense they are extending `datetime` using `tzinfo` objects. This allows them to correctly handle DST status of "nominally" ambiguous times (e.g. times where, in the specified zone, the date represented by `dt.replace(tzinfo=None)` is ambiguous), but causes problems with datetime arithmetic, since they don't re-evaluate the DST status in response to changes in the datetime.

[...]

PR #225, opened 2016-03-21
Time zone handling of ambiguous dates

pganssle:
This PR covers a few enhancements to the `tz` module.

I did some light refactoring of `tzfile` implementation as I was going over it, and I also realized that we were using some monstrously inefficient search in a frequently called method - basically, when trying to pinpoint which two transitions we're currently between, we were doing a linear search of all transition times, when we should have been using the binary search provided by the `bisect` method.

Additionally, I've implemented a method for handling ambiguous datetimes, as discussed in #112. The use of stateful `tzinfo` files is not ideal, but as far as I can tell this is the best way to actually handle ambiguous dates. This does nothing to address non-existent dates, but I think that is somewhat less of a problem for the moment.

As I was refactoring `tzfile`, I also addressed the odd edge case (#128) where a DST transition and a change in the base offset occur at the same time, cancelling one another out. We don't actually have enough information in the compiled binaries to address this the right way, but I've at least added the heuristic that if a zone is marked DST but it has zero time difference with its most recent standard time, infer the DST status from the _next_ standard time instead. I also changed it so that these DST offsets are cached in the `tzfile`, not calculated on the fly.


abalkin:
@pganssle - note that python/cpython#20 includes a fairly complete tzfile reader in the test.datetimetester module.  Feel free to borrow from and/or comment on it.

abalkin:
@tim-one - if this solution works then @pganssle has figured out something that we did not.  Could you take a look?  My first reaction was that this is a variant of a pytz kludge, but @pganssle says that it is not.

pganssle:
@abalkin @tim-one I did _not_ say it's not a variant of the pytz kludge. I suspect the approaches are very similar, since in both cases we're trying to store information about the `datetime` object in the `tzinfo`.

What I was saying was that I think this approach is a bit of a lighter touch, because:

1. it uses the parent `fromutc()` for the actual calculations, and only determines the "if it's DST" part based on how much the shift in the nominal timestamps was.
2. `dateutil` timezones are lazily-calculated, and as far as I can tell, `pytz` timezones are eagerly calculated, so you can't perform date arithmetic on them:

    ```python
    import pytz
    from dateutil import tz

    TORd = tz.gettz('America/Toronto')
    TORp = pytz.timezone('America/Toronto')
    HOUR = timedelta(hours=1)

    t0_u = datetime(2011, 11, 6, 5, 30, tzinfo=tz.tzutc())  # Folded date

    t0_d = t0_u.astimezone(TORd)
    t0_p = t0_u.astimezone(TORp)

    print("PYTZ during fold: {} (Should be EDT)".format(t0_p.tzname()))
    print("dateutil during fold: {} (Should be EDT)".format(t0_d.tzname()))

    print("PYTZ after fold: {} (Should be EST)".format((t0_p + HOUR).tzname()))
    print("dateutil after fold: {} (Should be EST)".format((t0_d + HOUR).tzname()))
    ```

    Result:

    ```
    PYTZ during fold: EDT (Should be EDT)
    dateutil during fold: EDT (Should be EDT)
    PYTZ after fold: EDT (Should be EST)
    dateutil after fold: EST (Should be EST)
    ```

I think the correct way to do things is to have the `fold` quality be an aspect of the datetime object itself. That said, I'm not sure how folding is handled with date arithmetic in gaps. For example, I would expect that if you add `timedelta(hours=1)` to `datetime(2011, 6, 4, 30, tzinfo=tz.gettz('America/Toronto'))`, you should be able to automatically get back `datetime(2011, 6, 1, 30, tzinfo=tz.gettz('America/Toronto', fold=1)`. Haven't taken a close enough look at the PEP495 PR to see if that's the case, but it's definitely not the case here (since there are no hooks in to `tzinfo` during date arithmetic operations as far as I know).

tim-one:
As I understand it, this approach (which I will creatively call "this approach" ;-) only stores a `fold` flag in a tzinfo, and only sets it on datetimes returned by `fromutc()` (where it may create a new tzinfo distinct from the input's, if the result is in a fold but the input was not).

pytz effectively stores all of the zone name, daylight status, and total UTC offset in a tzinfo, and wants the user to keep on giving it chances to replace a datetime's current tzinfo (via `normalize()` calls).

If so, they're similar in some ways and different in others.  pytz is keen to force timeline arithmetic, but this approach is not.  Either way should work fine to ensure correct UTC -> local -> UTC round trips.  Neither is the _intended_ way to implement hybrid zones (a single tzinfo for a named zone with various transitions), but the intended way really can't be done (with always-correct conversions) before the PEP.

I see one edge case difference with PEP behavior:  as I recall, the PEP forces `fold` to 0 for datetimes produced by arithmetic, but because such cases just reuse the input's tzinfo, in this approach the result will have the input's tzinfo (and so also its `fold` value).

Anything else?

pganssle:
@tim-one I tend to agree with that assessment. The one thing I'll mention is that it might be a good idea for me to rename the `_fold` flag here, because it's not actually intended to be an indicator of whether or not the datetime _is_ in a fold, rather it's supposed to be the _default value_ for `fold` for a given ambiguous date. In this setup, `_fold` is intended to be meaningless except when handling ambiguous dates.

This allows you to do things like this:

```python
from dateutil import tz
from dateutil.rrule import rrule, SU, YEARLY

from datetime import datetime

NYC = tz.gettz('America/New_York')

# 1:30 AM during a transition on the EDT side
dst_edt = datetime(2015, 11, 1, 5, 30, tzinfo=tz.tzutc()).astimezone(NYC)

# 1:30 AM during a transition on the EST side
dst_est = datetime(2015, 11, 1, 6, 30, tzinfo=tz.tzutc()).astimezone(NYC)

# 1st Sunday in November every year (DST rule)
rrule_kwargs = dict(freq=YEARLY, bymonth=11, byweekday=SU(+1), count=3)

dst_rrule_edt = rrule(dtstart=dst_edt, **rrule_kwargs)
dst_rrule_est = rrule(dtstart=dst_est, **rrule_kwargs)

strfmt = '%Y-%m-%d %H:%M:%S %Z%z'
for d_edt, d_est in zip(dst_rrule_edt, dst_rrule_est):
    print("EDT: {}".format(d_edt.strftime(strfmt)))
    print("EST: {}".format(d_est.strftime(strfmt)))
```

Result:

```
EDT: 2015-11-01 01:30:00 EDT-0400
EST: 2015-11-01 01:30:00 EST-0500
EDT: 2016-11-06 01:30:00 EDT-0400
EST: 2016-11-06 01:30:00 EST-0500
EDT: 2017-11-05 01:30:00 EDT-0400
EST: 2017-11-05 01:30:00 EST-0500
```

Without having to worry about whether the `rrule` calculation passed through some intermediate stage. Again, without a hook in to the `tzinfo` object during arithmetic operations, I don't think there's any good way to cover all cases.

Question, after PEP 495 is implemented, if I do something like this:

```
from dateutil import tz

NYC = tz.gettz('America/New_York')

before_dst = datetime(2015, 11, 1, 0, 30, tzinfo=NYC)
in_transition = before_dst + timedelta(hours=1)
```

Will the arithmetic operation correctly set `in_transition.fold` to 1?

pganssle:
I did, by the way, find one example of a somewhat annoying edge case:

```python
from dateutil import tz

UTC = tz.tzutc()
NYC = tz.gettz('America/New_York')
hr = timedelta(hours=1)

before_dst = datetime(2015, 11, 1, 0, 30, tzinfo=NYC)

in_transition = before_dst + hr
print("Arithmetic as NYC: {}".format(in_transition.tzname()))           # EST
in_transition_via_utc = (before_dst.astimezone(UTC) + hr).astimezone(NYC)
print("Arithmetic as NYC: {}".format(in_transition.tzname()))           # EDT!
print("Arithmetic as UTC: {}".format(in_transition_via_utc.tzname()))   # EDT
```

This is a consequence of the fact that `before_dst` is specified directly with the fold-naive `NYC` tzinfo, then when `fromutc()` is first called on `NYC`, the object itself is modified. I did it this way to prevent proliferation of `tzinfo` instances, but I think it's better to just generate a shallow copy of `tzinfo` with the `_fold` status changed every time a `_fold` transition takes place.

pganssle:
Hm.. Turns out the test failures on Windows are because for some reason the `zoneinfo.gettz()` `tzfiles()` are somehow completely immune to copying, I'm starting a new issue for that.

tim-one:
There's little you can do about arithmetic.  If I understand it, this approach won't be wholly compatible with pytz or with a PEP 495 implementation.

It's _intended_ that arithmetic within a single zone follow the "naive time" model:  that every day has exactly 24 hours.  With a few exceptions, datetime arithmetic in Python ignores the tzinfo members, except that datetime += timedelta mindlessly copies the source tzinfo to the result.  The arithmetic itself acts as if tzinfo were `None`.

The intent is that if people _want_ their datetimes to jump around to match local daylight (etc) transitions, then they should either use timestamps, or convert to UTC, do arithmetic in UTC, and convert back to local time at the end.

The unfortunate (we wouldn't do this if we could do it all over) exceptions are subtracting, or comparing, two datetimes.  Then *if* they have the same tzinfo member, the tzinfo member is again wholly ignored.  But if the tzinfo members differ (not the same object), both datetimes are effectively converted to UTC before the subtraction/comparison.

So, e.g., suppose we have two local datetimes `d1` and `d2` with `d1` corresponding to the later of an ambiguous local hour (so has `fold==1`) and `d2` the earlier (so has `fold==0`).  Under the PEP, they'll probably have the same tzinfo member, and then -- same as today! -- `d1 - d2` will be `0`:  tzinfo is ignored.  But under your scheme (if I understand it), they'll have distinct (not the same object) tzinfo members, and the subtraction will yield 1 hour.

The best way for users to get around this, when they care, is the intended way:  if they want "timeline arithmetic" (arithmetic that accounts for transitions moving the local clock around), they should work in UTC and convert back to the local zone when they're done.  The only thing that stops that working "perfectly" today is that it's currently not possible to do interzone conversions correctly in all cases (due to the current lack of an `isdst/fold` indicator to disambiguate ambiguous local times).  Alas, I still see no way to repair that, and preserve all the intents, short of storing `fold` in the datetime instance.

pganssle:
@tim-one Isn't `fold` being stored in the datetime instance in PEP 495?

Also, right now I've got the `tzinfo` as an "eager pass-through" of the `fold` attribute, so I agree that it's not currently 100% compatible with PEP 495's intent. I was thinking that I might switch it to a "soft pass-through" in the future, where multiple instances of a `tzinfo` file are not created when the `fold` was pulled from the `datetime`. I figured I'd wait until the real implementation is live, though, to implement this.

That said, I have to say I'm not in love with the approach of "all datetime arithmetic is naive arithmetic". The concept of "what is 1 hour from this datetime" seems unambiguous if you are *either* doing calculations in UTC *or* with `tzinfo` objects attached. However, it's probably a good idea for the addition and subtraction operators to be consistent, so barring an implementation of `datetime` that automatically converts to UTC during arithmetic operations on timezone-aware objects, I'll make sure to switch the implementation to "pass-through" folds.

Probably this sort of thing is where something like [arrow](http://crsmithdev.com/arrow/) can shine by providing `datetime`-like objects that have the intuitive behavior.

abalkin:
Stdlib datetime objects behave intuitively if you understand what they are.  They do not attempt to model any absolute notion of a "point in time."  They just model a clock with seven hands: year through microsecond.  A timedelta is a knob that allows you to get a new configuration of  hands and tzinfo is a label that helps interpret what the given configuration of hands means.  Morever, a datetime instance is a clock that is stopped.  An instance showing 1970-01-01T00:00:00.000000 now will show the same time in an hour and in 30 years unless it is destroyed.

The timedelta arithmetic is just a convenient set of elementary operations that can be used to model passing of time and timezone changes, but it does not by itself attempt to model what people do to their clocks in different parts of the world.

@pganssle - if you have not already - please take a look at datetime-sig archives.  We went through these issues at length last year.

pganssle:
Almost anything is intuitive if you understand it ;).

That said, I guess my intuitions were colored by the fact that I was thinking of the operation of "add 1 hour to X". If I think of the operation as "add 1 day to X", my answer for "what is the intuitive result of the operation over a DST transition" changes, so I guess I agree that there's no "obvious default" for arithmetic on timezone-labeled dates.

Maybe we should introduce (in dateutil) an `absolutedelta` class corresponding to the `relativedelta` class we have now. That way, if you want one behavior or the other, you can signal it by selecting an appropriate `delta` class.

tim-one:
The datetime SIG contains discussions of several possible ways to get "timeline arithmetic".  Most popular was to trigger "I want timeline arithmetic" by deriving from a new tzinfo marker subclass.

In any case, it can't be done without changes to the core, short of the kind of approach pytz took, including pytz requiring the user to keep calling `normalize()` after arithmetic.  Note that just adding a new kind of timedelta doesn't solve the problem:  `datetime - datetime` then has no idea whether you want timeline or classic subtraction.

The default cannot change:  classic arithmetic has been the rule since the start, and that was over a dozen years ago.  And Guido still likes that decision.

In the meantime, you "should" either do nothing about arithmetic (that's my advice), or copy what pytz does.  Anything between those will leave you with unique behaviors nobody expects.  People who need timeline arithmetic in std Python now can get it by working in UTC, converting to local time only when necessary.  However, my guess is that `dateutil` will need to implement its own `fromutc()` to handle all the pathologies in the zoneinfo database correctly.  As documented, the default `fromutc()` is _not_ ambitious.

Note that no changes are required in the core to do correct UTC -> local conversions in all cases (provided you implement `fromutc()`).  It's only the other direction that's an intractable problem in some cases (no way to disambiguate local times in a fold).  The PEP is a solution to that.  Before then, blame Python!  It is, after all, Python's fault ;-)

pganssle:
@tim-one I've created an issue (#228) to further explore the `absolutedelta` concept, but I think it's very reasonable considering we already have a `relativedelta` object. I think this is the right solution for this library, since we don't have any backwards-compatibility issues, and it allows for the behavior to be done explicitly if so desired. I suspect that for the most part, people will be perfectly happy to use either `timedelta` or `relativedelta`. I don't think it's the job of `tzinfo` objects to specify how arithmetic operations occur, they should just be providing information about the time zone so that you *can* get the behavior you want.

As for the subtraction situation, there's already precedent in the `relativedelta` object - `relativedelta`s can be constructed either by explicitly specifying the calendar date offset *or* from two `datetime` objects.

pganssle:
OK, I think this is ready to go (though it will need modifications when a Python 3.6 release candidate comes out).

PR #153, opened 2015-11-30
#93 support specifiying time zone by full name

aschatten:
@jarondl this is an updated version of https://github.com/dateutil/dateutil/pull/150. Had to re-create because I somehow messed up my branch and was not merge upstream changes properly.
Fixes #93.

PR #227, opened 2016-03-24
Aschatten issue 93 parse tz full name

pganssle:
Merged #153 into base, with some changes.


pganssle:
I am thinking this can still be reorganized a bit, I'm not crazy about all the public interface changes in `zoneinfo` for example, and I'm thinking maybe we can replace `_tzparserinfo()` with a radix tree or something of all time zone names.

pganssle:
Actually, this seems like a pretty big change, I'm going to push it to the next feature release.

PR #229, opened 2016-03-26
Fix yearfirst/dayfirst bug.

### LINKED ISSUE #81: Parser doesn't use dayfirst and yearfirst when no separator
It always uses the format: YYMMDD or YYYYMMDD despite if yearfirst is set to False or dayfirst is set to True.
I think it should try:
* yearfirst=False and dayfirst=False: MMDDYY
* yearfirst=True and dayfirst=False: YYMMDD
* yearfirst=True and dayfirst=True: YYDDMM
* yearfirst=False and dayfirst=True: DDMMYY

### LINKED ISSUE #217: yearfirst bug fixed?
xref https://github.com/pydata/pandas/issues/12611

I guess this was a bug in < 2.5?

```
In [1]: import dateutil.parser

In [2]: p = dateutil.parser.parser()

In [3]: p.parse(StringIO('090131'), dayfirst=False, yearfirst=False)
Out[3]: datetime.datetime(2009, 1, 31, 0, 0)

In [4]: dateutil.__version__
Out[4]: '2.4.1'
```

```
In [1]: import dateutil.parser
In [3]: p = dateutil.parser.parser()

In [4]: p.parse(StringIO('090131'), dayfirst=False, yearfirst=False)
Out[4]: datetime.datetime(2031, 9, 1, 0, 0)

In [5]: p.parse(StringIO('090131'), dayfirst=False, yearfirst=True)
Out[5]: datetime.datetime(2009, 1, 31, 0, 0)

In [5]: import dateutil

In [6]: dateutil.__version__
Out[6]: '2.5.0'
```

pganssle:
Fixes #81 and resolves #217.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -464,7 +464,10 @@ class _ymd(list):
                     self.find_probable_year_index(_timelex.split(self.tzstr)) == 0 or \
                    (yearfirst and self[1] <= 12 and self[2] <= 31):
                     # 99-01-01
-                    year, month, day = self
+                    if dayfirst:
+                        year, day, month = self
+                    else:
+                        year, month, day = self
                 elif self[0] > 12 or (dayfirst and self[1] <= 12):
                     # 13-01-01
                     day, month, year = self
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -777,3 +777,40 @@ class ParserTest(unittest.TestCase):
         self.assertEqual(parse(pstring),
                          datetime(1924, 8, 29))
 
+    def testNoYearFirstNoDayFirst(self):
+        dtstr = '090107'
+
+        # Should be MMDDYY
+        self.assertEqual(parse(dtstr),
+                         datetime(2007, 9, 1))
+
+        self.assertEqual(parse(dtstr, yearfirst=False, dayfirst=False),
+                         datetime(2007, 9, 1))
+
+    def testYearFirst(self):
+        dtstr = '090107'
+
+        # Should be MMDDYY
+        self.assertEqual(parse(dtstr, yearfirst=True),
+                         datetime(2009, 1, 7))
+
+        self.assertEqual(parse(dtstr, yearfirst=True, dayfirst=False),
+                         datetime(2009, 1, 7))
+
+    def testDayFirst(self):
+        dtstr = '090107'
+
+        # Should be DDMMYY
+        self.assertEqual(parse(dtstr, dayfirst=True),
+                         datetime(2007, 1, 9))
+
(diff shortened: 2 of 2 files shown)


muffinmad:
Biefly looking at [this](https://en.wikipedia.org/wiki/Date_format_by_country) I don't see that someone use YDM format. I think if year is in the first place then `dayfirst ` must be always `False`.

Chronial:
@pganssle Was the feature you where talking about added somewhere? We would like to upgrade to the current version, but are still pinned to 2.5.1 due to this.

pganssle:
@Chronial No, sorry, but it's the next big project I'm planning. It may be some time. There are ways to work around it if need be, if you want to upgrade. Depending on your inputs, some judicious use of `datetime.strptime` before `dateutil.parser.parse` can get you where you need to be. See [this comment](https://github.com/dateutil/dateutil/issues/402#issuecomment-311277352) for example.

Chronial:
Thx for the response, but we are parsing external data and it might be in one of many formats – exactly the kind of situation dateutil is perfect for.

Would you maybe be willing to just revert this change, or add another parameter / rename one of the other two? The thing is in Europe I can be quite sure to never receive an American date, but it might be an ISO-Date (so basically the same situation as #402) – while nobody actually uses YDM anywhere.

In your previous comment you called this "random parser switches", but I would argue that exactly this is the most important parser switch: Am I in a culture where you do DMY or MDY? ISO dates are always ISO dates.

pganssle:
@Chronial I would rather not create a bunch of uncertainty about the behavior. I'd rather users be able to give much more granular specification of their context, so I think it's best if we just power through and try and make that happen rather than flip flop on the behavior here.

pganssle:
@Chronial There is another intermediate solution, by the way, I'm sure it's in one of the issues related to this, don't have time at the moment to find it, but I think there's a solution that involves parsing multiple times with different values for the `*first` flags and inferring which one to use based on how the returned dates differ. Obviously not the best solution, but it may alleviate some pain temporarily.

Chronial:
Your project :), and we are alright being pinned to 2.5.1. I would just ask that in whatever solution you end up with, that you think of us non-Americans who just want a simple option to do the sensible thing outside of the US :).

We don't want to tweak the parser – it's just that the default behavior is just plain wrong in our culture.

qpkorr:
Well said Chronial.  I've just been bitten by this one again - since in Australia we normally use dd-mm-yyyy - but we also wish to support ISO format (yyyy-mm-dd). dateutil used to support them both perfectly. 
 Guess I'll try to pin to 2.5.1 too - though I don't think it's available by default on debian stretch.

PR #232, opened 2016-03-28
option to prefer year over day

merutak:
e.g. interpret 10/11 as October 2011, not October 11th or November 10th

Example:
dateutil.parser.parse('10/11', preferyearoverday=True)
Passing 'preferyearoverday=False' should be backwards-compatible with preexisting code.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -276,7 +276,7 @@ class parserinfo(object):
     PERTAIN = ["of"]
     TZOFFSET = {}
 
-    def __init__(self, dayfirst=False, yearfirst=False):
+    def __init__(self, dayfirst=False, yearfirst=False, preferyearoverday=False):
         self._jump = self._convert(self.JUMP)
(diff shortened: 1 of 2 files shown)


pganssle:
Hm. This is worth considering, but I definitely don't like the name of the argument, and I continue to be uncomfortable with the proliferation of arguments to `parser`.

I sorta think that `yearfirst`, `dayfirst` and this should be collapsed in some way so that we can put the entire format spec in one argument, rather than having a host of boolean options. Then we can phase out the use of `yearfirst`/`dayfirst` before a 3.0 release, or just leave them as legacy options.

I think that this should be postponed until #125 is addressed, because when implementing #125, my hope is to break up the `parse` function into smaller, extensible functions, to make it easier for end users to tweak the behavior of any one part of the parser as desired. Another part of it will be encoding all or most of the assumptions that go into the parser into the `parserinfo` class, so it's much easier to configure things like this.

pganssle:
Pretty sure this isn't going into the final API, closing.

PR #244, opened 2016-04-03
Run updatezinfo.py script with coverage

pganssle:
For now we can consider the fact that we're running updatezinfo to be part of testing the `rebuild` script for purposes of the code coverage. At some point we should probably write actual unit tests for it, though.


pganssle:
Odd. The `codecov` coverage for this PR doesn't seem to include the `rebuild` script. Locally I get 75% coverage on `rebuild`.

pganssle:
Given that this doesn't seem to have worked, I'm closing it.

PR #263, opened 2016-04-28
Format wkst properly in rrule.__str__

### LINKED ISSUE #262: rrule.__str__ doesn't render wkst properly
`rrule.__str__` is rendering `WKST` as an integer instead of as a weekday. For example:

```python
>>> rule = dateutil.rrule.rrulestr("DTSTART:20000101\nFREQ=WEEKLY;WKST=SU")
>>> str(rule)
'DTSTART:20000101T000000\nFREQ=WEEKLY;WKST=6'
```
This breaks compatibility with RFC5545 and also with `rrulestr`:
```python
>>> dateutil.rrule.rrulestr(str(rule))
    ...
ValueError: invalid 'WKST': 6
```

dplepage:
This fixes #262 by converting `wkst` to a weekday string when stringifying an rrule.

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -707,7 +707,7 @@ class rrule(rrulebase):
             parts.append('INTERVAL=' + str(self._interval))
 
         if self._wkst:
-            parts.append('WKST=' + str(self._wkst))
+            parts.append('WKST=' + repr(weekday(self._wkst))[0:2])
 
         if self._count:
             parts.append('COUNT=' + str(self._count))
--- a/dateutil/test/test_rrule.py
+++ b/dateutil/test/test_rrule.py
@@ -4382,6 +4382,12 @@ class RRuleTest(WarningTestMixin, unittest.TestCase):
                               byminute=(1,),
                               dtstart=datetime(2010, 3, 22, 12, 1)))
 
+    def testToStrWithWkSt(self):
+        self._rrulestr_reverse_test(rrule(WEEKLY,
+                              count=3,
+                              wkst=SU,
+                              dtstart=datetime(1997, 9, 2, 9, 0)))
+
     def testToStrLongIntegers(self):
         if not PY3:  # There is no longs in python3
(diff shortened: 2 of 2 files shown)


dateutil/rrule.py
```diff
@@ -707,7 +707,7 @@ def __str__(self):
             parts.append('INTERVAL=' + str(self._interval))
 
         if self._wkst:
-            parts.append('WKST=' + str(self._wkst))
+            parts.append('WKST=' + repr(weekday(self._wkst))[0:2])
```
pganssle: Just idle speculation, I'm wondering if maybe `str(weekday)` should return `MO` for `MO(0)`. Then we could just replace this with `str(weekday(self._wkst(0))))`.

I think that's a matter for a different PR, though.

pganssle:
This looks good to me, so I'm going to go ahead and merge it.

Thanks for the bug report and PR!

dateutil/rrule.py
dplepage: I don't think it's possible to call `MO(0)`:
```python
>>> dateutil.rrule.MO(0)
    ...
ValueError: Can't create weekday with n == 0
```

That said, `str(weekday(self._wkst))` appears to produce the right value already. I only did `repr(day)[0:2]` here because that was the pattern used further down in the file, and I figured I'd follow suit in case there was something more complex going on that I didn't understand.

dplepage:
Thanks for merging this so promptly!

dateutil/rrule.py
pganssle: Ah, I see what's going on. I actually meant that you'd use `MO(None)`, but I didn't really look closely enough, I didn't realize that `_wkst` was storing an integer and not a `weekday`  object.

I think the reason it's split up like that just below is because for `byweekday` you can use weekday offsets like `SU(1)`, and to conform to the RFC, that has to print as `+1SU`. Since `_wkst` can't have that offset parameter, it's not a concern.

No worries, though, seems fine as is.

Issue #260, opened 2016-04-25
Refactor tzwin as a tzrange subclass


pganssle:
When working on #249, I realized that all the "time zone" parts of `tzwinbase` are essentially just a different implementation of `tzrange`. I think it's probably a good idea to refactor the library such that `tzrange` is the base class for `tzwin`.

This, along with #253, can be part of a general unification of all the interface of the time zone classes.

pganssle:
Implemented in #302.

Issue #253, opened 2016-04-16
Detection of non-existent / ambiguous times


pganssle:
It is not clear to me that there is, currently, any good way to detect a non-existent time using dateutil time zones. I think maybe the right thing to do is to convert all dateutil time zones to use a common base class (`_common._tzinfo`) or mixin that implements some function for detecting non-existent and/or ambiguous times.

This will make it much easier to deal with, for example, an `rrule` that uses a `dateutil` time zone but which might generate a non-existent time. With `pytz`, you can already do something like this:

``` python
from functools import partial
import pytz
from dateutil import rrule

RR = list(rrule.rrule(freq=rrule.DAILY, dtstart=datetime(2002, 4, 6, 8, 30), count=2, byhour=2))

NYCp = pytz.timezone('America/New_York')

def datetime_exists(x, tzi):
    try:
        tzi.localize(x, is_dst=None)
    except pytz.AmbiguousTimeError:
        pass
    except pytz.NonExistentTimeError:
        return False

    return True

RR1 = list(filter(partial(datetime_exists, tzi=NYCp), RR))
```

Which filters out the non-existent dates. We should probably have some `is_ambiguous()` and `exists()` functions for `dateutil` timezones.

pganssle:
Implemented in #302.

PR #264, opened 2016-05-01
Full ambiguous timezone support.

### LINKED ISSUE #249: Ambiguous time support for all time zone classes.
Before the 2.6.0 release, all timezones with transitions should support ambiguous times.

See #225 and #248.

The classes with non-fixed offset support are:

- [x] `tzlocal` (PR #248)
- [x] `tzfile` (PR #225)
- [x] `tzwin` (PR #225)
- [x] `tzwinlocal` (PR #225)
- [ ] `tzstr`
- [ ] `tzrange`
- [ ] `tzical`

Right now the implementations are all ad hoc because transitions are determined differently for each class. Ideally, we'd refactor all the time zone classes around a single base class and try and be a bit more systematic about how this is done, but for now I think what we've got is fine.

pganssle:
This PR is not ready for merging, I just want to fire off the CI as I work on it. This will resolve #249. 

So far it includes:
* Improved documentation for many classes in the `tz` module.
* Fold support for `tzstr`, `tzrange` and `tzlocal`.
* Improved test coverage for `tzical`

Before merging, the following still needs to be done:
 - [x] Fold support for `tzical`

Deferring these two:
 -  <strike>Documentation for `tzfile`</strike>
 - <strike>Implementation of #253 </strike>

I will also possibly implement #260, depending on how it's best to implement #253.

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -306,7 +306,6 @@ class TzFoldMixin(object):
             self.assertEqual(t0.utcoffset(), timedelta(hours=-5.0))
             self.assertEqual(t1.utcoffset(), timedelta(hours=-4.0))
 
-
     def testFoldIndependence(self):
         tzname = self._get_tzname('America/New_York')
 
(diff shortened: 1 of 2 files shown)


pganssle:
Tests have passed, so I'm going to merge this. I'm deferring the tzfile documentation and implementation of #253 for a separate PR.

PR #277, opened 2016-07-16
Added support for timedelta objects when constructing tzoffsets

pganssle:
Thought I had already added this feature. This allows `tzoffset` to be constructed from a `timedelta`.


pganssle:
I think what I actually added before was support for `timedelta` objects in `tzrange`.

I'm going to hold off on merging this because I want to update the docs to make sure it's clear that `timedelta` as an input is new in 2.6.0.

pganssle:
I'm going to merge this and make a second documentation PR so I don't have to rebase.

Issue #278, opened 2016-07-19
Incorrect fuzzy date parsing. 


shmuga:
Hello! 
I have detected some strange datetime parsing.
Version **2.5.3**

```python
>>> from dateutil.parser import parse
>>> parse("27.08", fuzzy=True)
datetime.datetime(2016, 7, 27, 0, 0)
```

shmuga:
I tried to fix this bug. But there are some cases which can't be fixed.
For example - if someone uses a date like '27.08' or '12.25' we can probably define where is the month and where is a day. But when the date is like '12.05' we can't define it. So the first part will be  the month and the last one is a day.

pganssle:
This does not appear to be fuzzy in any way.

I'm not really sure how to best interpret this as a date - should it bet August 2027, August 27th of the current year? Hard to know, but I guess the assumption that it's month-day or day-month is valid.

However, it's certainly a bug that it's ignoring the `08` portion entirely without throwing an error for *non* fuzzy parsing:

```
>>> from dateutil.parser import parse
>>> parse('27.08')
datetime.datetime(2016, 7, 27, 0, 0)
>>> parse('08.27')
datetime.datetime(2016, 7, 8, 0, 0)
```

Very strange.

shmuga:
I've looked through the code and found that the problem is with tokenizing this date.
The output of tokenizer is the same string.
I made a try to create rule for tokenizer - but it breaks lot of tests connected with time.
I'll look more for it and make one more try.

Could you help how do you think it would be better to create array of tokens? ['27', '.' , '08'] ?

jbrockmendel:
I'd classify "27.08" as "not sufficiently obvious that this is a date".  In the wild isn't this much more likely to just be a float?  Or part of "$27.08"?

PR #279, opened 2016-07-20
Bug #278: Incorrect fuzzy date parsing.

shmuga:
Bug related to #278.
- Added test case to bug.
- Fix bug when first part of ymd object is larger than 12 and length of ymd array == 2 - it's exactly date not month.
- Added rule for case when date in format "DD.MM" or "MM.DD"

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -436,6 +436,9 @@ class _ymd(list):
             elif dayfirst and self[1] <= 12:
                 # 13-01
                 day, month = self
+
+            elif self[0] > 12 and self[1] <= 12:
+                day, month = self
             else:
                 # 01-13
                 month, day = self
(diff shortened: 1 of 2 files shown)


dateutil/test/test_parser.py
```diff
@@ -579,6 +579,11 @@ def testFuzzyIgnoreAMPM(self):
 
         self.assertEqual(parse(s1, fuzzy=True), datetime(1945, 1, 29, 14, 45))
 
+    def testFuzzyWithDotNotation(self):
+        s1 = "27.08"
+        today_year = date.today().year
+        self.assertEqual(parse(s1, fuzzy=True), datetime(today_year, 8, 27, 0, 0))
```
pganssle: This should be changed in the following ways:

1. Just pass a specific date to `default` so you don't have to worry about retrieving the current year
2. Remove the `fuzzy` keyword. It is unnecessary.

dateutil/parser.py
```diff
@@ -712,6 +715,11 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
                         if len_li == 4:
                             res.minute = int(s[2:])
 
+                    elif len_li == 5 and len_l == 1 and value_repr.count('.') == 1:
+                        after_split = value_repr.split('.')
```
pganssle: I don't think this is the right way to handle this. The tokenizer should be doing the splitting, we should try to avoid hard-coding tokens into the token handling logic.

pganssle:
Doesn't seem like this is still being worked on, closing this.

PR #280, opened 2016-07-21
added default argument to parser.parse() method to specify whether to fill in missing date values with default values or return a ValueError.

spw1016:
Added the default argument defaultdate=True to the dateutil.parser.parse(). This allows a user to specify whether they want the parser to provide a default day, month, or year element if one is missing from the parsed datetime object. If defaultdate is set to False, the parser will return a ValueError if there is a missing date element.

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,15 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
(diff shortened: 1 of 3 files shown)


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,15 @@
+<component name="InspectionProjectProfileManager">
```
pganssle: Don't check in things like this.

pganssle:
As I've mentioned in other threads on this, I would like to change the way the parser options are specified, and that I think a proliferation of arguments to `parse` is not the way to do it. In the future, I think this sort of thing will be handled best by breaking out the things the parser is doing into smaller modules and allowing them to be overridden.

Issue #282, opened 2016-07-23
datetime.timedelta cannot be added to relativedelta


justanr:
```python
from dateutil import relativedelta
import datetime

relativedelta.relativedelta(years=1) + datetime.timedelta(days=1)
```

Results in a `TypeError: unsupported type for add operation`, despite both being time offsets.

However, it is perfectly possible to add a `relativedelta` to another `relativedelta`:

```python

relativedelta.relativedelta(years=1) + relativedelta.relativedelta(years=1)
# relativedelta(years=+2)
```

I'm happy to submit a PR if there is interest (beyond myself) in this.

pganssle:
Hm.. I think this is a reasonable enhancement, considering that for unambiguous relative units (`day` and below), I believe `relativedelta` acts exactly like a `timedelta`, and `timedelta` doesn't *accept* unambiguous units. It's worth looking into for incompatibilities, though.

justanr:
Alright, I'll start putting together a PR.

PR #283, opened 2016-07-25
Allow datetime.timedelta to be added to relativedelta

justanr:
In response to issue #282. I played with it in the context I first noticed it and everything seems fine.


pganssle:
This looks good to me. I'm going to take a bit of a closer look tonight and then merge.

PR #296, opened 2016-10-07
Fix some Python 3 issues flagged by the -3 flag

rowillia:
I've been working through looking at adopting Python 3, and I've been utilizing the -3 flag to help idenitfy issues with Python 3.

Here's a quick write up I did last week:

https://gist.github.com/rowillia/c0feed97c1863b2d8e5a3ed73712df65

The -3 flagged a few issues in dateutil worth fixing, namely around __eq__ shadowing __hash__ in PY3.

Fixing these issues will help make the -3 flag less noisey for folks running dateutil in production.

### DIFF
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -32,6 +32,8 @@ class weekday(object):
             return False
         return True
 
+    __hash__ = None
+
     def __repr__(self):
         s = ("MO", "TU", "WE", "TH", "FR", "SA", "SU")[self.weekday]
         if not self.n:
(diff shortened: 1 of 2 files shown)


rowillia:
@pganssle Let me know if you've got any questions on this!

pganssle:
@rowillia Yes, sorry for the delay, it's been some time since I've had time to work on `dateutil` stuff.

I guess the issue is that classes default to hashable in Python 2, but to non-hashable in Python 3? I suppose if it hasn't come up already in Python 3 compatibility already, it's not a behavior too many people are relying on.

I would normally say we should add some tests, but I don't know that we want to signal that this behavior is definitely permanent. I could see the some possibility for adding a hash function in to the time zone objects, at least, at some point in the future, so I'm just going to merge as is.

PR #297, opened 2016-10-12
Add test for handling dates near the year zero

### LINKED ISSUE #293: Inconsistent behaviour when dealing with dates close to year 0
I wrote the following simple test using the python-hypothesis library. It generates dates, renders them as strings, then unpacks them using `parser.parse()` and checks the date survived the trip:

``` python

@given(datetimes())
def test_round_trip(d):
    assert parse(str(d)) == d
```

If you try enough examples: this test fails:
- Pass in `d=datetime.datetime(4, 4, 1, 0, 0)` – i.e., April 1st in the year&nbsp;4
- Casting to `str()`, this becomes `'0004-04-01 00:00:00'`
- When unpacked with `parse()`, it becomes `datetime.datetime(1, 4, 4, 0, 0)` – i.e., April 4th in the year&nbsp;1. Mismatch!

The same problem occurs for all `datetime.datetime(M, M, 1, 0, 0)` for `1 < M <= 12`.

Given the four-digit year, I think this string is fairly unambiguous, but it seems to be getting confused somewhere, and treating the `-01` as the year. But [two-digit years are supposed](https://labix.org/python-dateutil#head-19e5d3331806e55ec2ba1b9e39458784cc0a0004) to be within 50&nbsp;years of the current year:

> **Converting two digit years**
> 
> When a two digit year is found, it is processed considering the current year, so that the computed year is never more than 49 years after then current year, nor 50 years before the current year. In other words, if we are in year 2003, and the year 30 is found, it will be considered as 2030, but if the year 60 is found, it will be considered 1960.

So if it really is treating `-01` as the year, then based on this rule it should be returning `datetime.datetime(2001, 4, 4, 0, 0)`.

alexwlchan:
This includes an explicit test case for a bug that was spotted,
and a test using Hypothesis to seek out other bugs in a similar vein.

---

I had a stab at trying to fix #293 myself, but I got lost somewhere in the parsing code.

So this commit adds two tests: the specific bug that was reported, and the round-trip that originally caught it. Over dozens of CI runs, it will try more and more random examples, which increases the likelihood of other bugs like this being caught.

At least one will fail immediately – once it starts passing, we’ll know this bug has been fixed.

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -11,7 +11,7 @@ import six
 from six import assertRaisesRegex, PY3
 from six.moves import StringIO
 
-from hypothesis import given
+from hypothesis import example, given
 from hypothesis.extra.datetime import datetimes
 
 class ParserTest(unittest.TestCase):
@@ -862,11 +862,7 @@ class ParserTest(unittest.TestCase):
         self.assertEqual(parse(dtstr, dayfirst=True, yearfirst=True),
                          datetime(2015, 9, 25))
 
-    def testWithDateCloseToYearZero(self):
-        # https://github.com/dateutil/dateutil/issues/293
-        dtstr = '0004-04-01'
-        self.assertEqual(parse(dtstr), datetime(4, 4, 1))
-
+    # https://github.com/dateutil/dateutil/issues/293
     @example(datetime(4, 4, 1))
     @given(datetimes())
     def testRoundTripParsing(self, dt):


pganssle:
@alexwlchan I'm not sure I totally understand, is this a non-deterministic test? I'm not really comfortable with adding non-deterministic tests into the unit testing suite, because it means that the CI may randomly fail for users who make pull requests even though nothing they did actually caused it to fail.

Is there a way to make this run _only_ with the given examples by default? That way people can run some fuzzing  if they want to, but it doesn't affect the CI.

alexwlchan:
> is this a non-deterministic test?

Yes, Hypothesis will try random examples until it finds a failure. At that point, it will reliably fail until you fix the bug or tell it that it tried a bad example. So there is a possibility that an unrelated patch happens to get “lucky” and find a rare example for the first time.

> Is there a way to make this run *only* with the given examples by default?

I don’t think so. Perhaps instead just move the Hypothesis test into a separate test file, which doesn’t run in CI? Then the test is there for somebody who wants to run it offline, but it won’t cause unexpected CI failures.

alexwlchan:
Closing due to inactivity.

PR #300, opened 2016-10-16
add rejected timestr in exception message providing more information to caller

pbiering:
helpful if exeption is catched above for displaying instead of being blind, e.g. diagnosting broken vcalendar items:

```
vobject.base.ParseError: "In transformToNative, unhandled exception on line None: <class 'ValueError'>: ('Unknown string format:', '1970DTSTART:19700329T020000')"
vobject.base.ParseError: "In transformToNative, unhandled exception on line None: <class 'ValueError'>: ('Unknown string format:', '19701TZOFFSETFROM:-0000')"
```

(don't ask how such entries can occur...)

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -556,10 +556,10 @@ class parser(object):
         res, skipped_tokens = self._parse(timestr, **kwargs)
 
         if res is None:
-            raise ValueError("Unknown string format")
+            raise ValueError("Unknown string format:", timestr)
 
         if len(res) == 0:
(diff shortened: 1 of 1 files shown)


pganssle:
This seems perfectly reasonable to me, but I have a few bugfixes I want to push in a bugfix release, and I'm afraid that this sort of thing should not be changed in a bugfix release, in the event that someone is (possibly improbably) relying on the current error message being the same.

Ideally, we'd also change this to a custom error type (a subclass of `ValueError`) so people can have an "official" way to catch these errors.

Issue #286, opened 2016-07-26
Adjust `_tzinfo` implementation to return _DatetimeWithFold or equivalent.


pganssle:
Moving this discussion from #57.

Per the discussion there, I am convinced that @abalkin's PEP-495-consistent implementation of ambiguous time handling [in this mini-project](https://github.com/abalkin/tz) is the right way to do things if we want to be consistent across versions when PEP-495 goes into effect.

The short form is that we'd add a class like this:

``` python
    class _DatetimeWithFold(datetime):
        __slots__ = ()

        @property
        def fold(self):
            return 1
```

And rather than spawning new instances of `_tzinfo` to handle fold behavior, we'd handle fold behavior with `fold = getattr(dt, fold, 0)`.

PEP-495 has some (to me) counter-intuitive choices, so we should probably make an effort to thoroughly document timezone behavior.

abalkin:
Let me add a link to the [relevant datetime-sig thread](https://mail.python.org/pipermail/datetime-sig/2015-September/000860.html).  Note that the issue of PEP 495 gap/fold rules being the opposite of the "legacy" rules was discussed in that thread.  Here is Tim's recommendation for how to deal with that in a backward compatible manner:

>  I'd suggest leaving the current implementation alone,
and _adding_ a new wrapping of tzfiles based on Alex's code.
dateutil's get-a-zone factory functions would need to grow some way to
spell "I want a pre-495 tzinfo" or "I want a post-495 tzinfo".  New
functions, optional function flags, global setting ... whatever you
think works best.

See [Tim's post](https://mail.python.org/pipermail/datetime-sig/2015-September/000863.html).

pganssle:
I'm not all *that* worried about changing the legacy behavior. It seems like a lot of needless complexity to maintain two versions of all `tzinfo` implementations just so we can have backwards compatibility around ambiguous dates, considering that anyone who cares about handling ambiguous dates was probably using `pytz` anyway, since up until recently we've had no way to represent them.

I dunno, I'll see if there's an easy way to do it, but in some sense I'd rather just say, "This is the new behavior, going forward."

Another, probably less desirable, option is to use `getattr(dt, fold, 1)` (or possibly `getattr(dt, fold, 0)`) instead of `getattr(dt, fold 0)`. This would make the use of a python with a PEP-495 implementation the clear demarcation line where the behavior changes. There's a case to be made there, but I think it's a bit implicit for my taste.

pganssle:
[...]

PR #302, opened 2016-11-03
Implement PEP495 for all tzinfo objects.

### LINKED ISSUE #284: _fold attribute has inverted sense of PEP-495
See discussion on #57. The `_fold` attribute should have the inverted sense.

This may need to be a breaking change, as the default value for `fold` is 0, meaning that, in Python 3.6:

``` python
from datetime import datetime, timedelta, timezone
from dateutil.tz import gettz

LA = gettz('US/Pacific')
t = datetime(2014, 11, 2, 1, tzinfo=LA)
```

`t` should be interpreted as the _first_ 1AM, not the _second_ 1 AM.  This is the opposite of the assumption that has been made in every `tzinfo` subclass provided by `dateutil` up until this point.

Given that we've been pretty bad about ambiguous time zone support, hopefully no one is relying on this specific behavior, and the breaking change won't be so bad.

pganssle:
This fixes #284 and #286 and adds Python 3.6 to the `tox.ini` to test the ability to fall back to real "fold" functionality.

This also adds the `is_ambiguous(dt)` function to all `_tzinfo` subclasses, and exposes the `enfold` function for backwards compatibility purposes, which is a first crack at #253.

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -350,10 +350,13 @@ class TzWinFoldMixin(object):
             t0_u = t_n.replace(tzinfo=tzi).astimezone(tz.tzutc())
             t1_u = t0_u + timedelta(hours=1)
         else:
-            t_n = dstoff + timedelta(minutes=30)
+            # Get 1 hour before the first ambiguous date
+            t_n = dstoff - timedelta(minutes=30)
 
-            t1_u = t_n.replace(tzinfo=tzi).astimezone(tz.tzutc())
-            t0_u = t1_u - timedelta(hours=1)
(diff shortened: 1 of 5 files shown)


pganssle:
@abalkin As requested, all these are implemented as wrappers around (effectively) `fromutc()`, though as I mentioned, `tzfile()` and `tzical()` may have edge cases where it fails (not in the test suite).

pganssle:
This is super weird, running the tests locally with `tox -e pypy`, they all pass, and for some reason Appveyor is failing to import `six` on anything but Python2.7? Very odd.

pganssle:
The latest commits now cover #260.

PR #303, opened 2016-11-03
Tweak CI due to version inconsistencies and other issues.

pganssle:
It seems that something related to #302 is causing only old versions of pypy to fail, and travis only supplies old versions. This change will make pypy3 passing optional (no newer version is available) and forces an upgrade in travis.

There is also some sort of issue in Appveyor where the `pip` command isn't being found in the path for the right version of Python for some reason. This PR explicitly uses the right version of pip.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,38 @@ python:
   - "3.5"
   - "pypy"
   - "pypy3"
+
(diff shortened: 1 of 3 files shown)


pganssle:
Changing base of PR.

PR #319, opened 2016-11-09
respect parsed tz instead of preferring tzlocal

### LINKED ISSUE #318: RFC822 timestamp using GMT is parsed into BST if local time is UK time

```python
from datetime import datetime
from dateutil.tz import tzutc


print(d1 == d2)
```

jstewmon:
if the computer is in a timezone where daylight savings is observed,
such as Europe/London, and a date string specifies a timezone, the
parsed datetime may be incorrect if the parsed datetime falls within
daylight savings but the date string specified the standard time

this is especially problematic for users in GMT/BST when parsing
dates in http or email headers, which are frequently specified as
GMT

fixes #318

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -602,8 +602,6 @@ class parser(object):
                     raise ValueError("Offset must be tzinfo subclass, "
                                      "tz string, or int offset.")
                 ret = ret.replace(tzinfo=tzinfo)
-            elif res.tzname and res.tzname in time.tzname:
-                ret = ret.replace(tzinfo=tz.tzlocal())
             elif res.tzoffset == 0:
                 ret = ret.replace(tzinfo=tz.tzutc())
             elif res.tzoffset:
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-from ._common import unittest
+from ._common import unittest, IS_WIN
 
 import os
 import time
@@ -13,6 +13,11 @@ from dateutil.parser import *
 import six
 from six import assertRaisesRegex, PY3
 from six.moves import StringIO
+
+if IS_WIN:
+    from ._common import TZWinContext as TZContext
+else:
+    from ._common import TZEnvContext as TZContext
 
 
 class ParserTest(unittest.TestCase):
@@ -134,15 +139,10 @@ class ParserTest(unittest.TestCase):
                          datetime(2003, 9, 25, 10, 36, 28))
 
     def testDateCommandFormatRespectTz(self):
-        orig = os.environ.pop("TZ", None)
-        os.environ["TZ"] = "Europe/London"
-        time.tzset()
-        self.assertEqual(parse("Wed, 02 Oct 2002 13:00:00 GMT"),
-                         datetime(2002, 10, 2, 13, 0, tzinfo=tzoffset('GMT', 0)))
-        del os.environ["TZ"]
-        if orig is not None:
-            os.environ["TZ"] = orig
-        time.tzset()
+        tz = "British Summer Time" if IS_WIN else "Europe/London"
+        with TZContext(tz):
+            self.assertEqual(parse("Wed, 02 Oct 2002 13:00:00 GMT"),
+                            datetime(2002, 10, 2, 13, 0, tzinfo=tzoffset('GMT', 0)))
 
     def testDateCommandFormatStrip1(self):
         self.assertEqual(parse("Thu Sep 25 10:36:28 2003"),
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -2,6 +2,9 @@
 from __future__ import unicode_literals
 from ._common import unittest
 
+import os
+import time
+
 from datetime import datetime, timedelta, date
 
 from dateutil.tz import tzoffset
@@ -10,6 +13,7 @@ from dateutil.parser import *
 import six
 from six import assertRaisesRegex, PY3
 from six.moves import StringIO
+
 
 class ParserTest(unittest.TestCase):
 
@@ -50,7 +54,6 @@ class ParserTest(unittest.TestCase):
 
             def read(self, *args, **kwargs):
                 return self.stream.read(*args, **kwargs)
-
 
         dstr = StringPassThrough(StringIO('2014 January 19'))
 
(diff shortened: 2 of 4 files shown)


dateutil/parser.py
```diff
@@ -602,8 +602,6 @@ def parse(self, timestr, default=None, ignoretz=False, tzinfos=None, **kwargs):
                     raise ValueError("Offset must be tzinfo subclass, "
                                      "tz string, or int offset.")
                 ret = ret.replace(tzinfo=tzinfo)
-            elif res.tzname and res.tzname in time.tzname:
```
pganssle: This would break compatibility, I'm pretty sure we shouldn't do this as implemented. The most I'd be comfortable with is moving this down the list as the last resort rather than one of the first.

dateutil/test/test_parser.py
```diff
@@ -125,11 +127,23 @@ def testDateCommandFormatWithLong(self):
                                    tzinfos={"BRST": long(-10800)}),
                              datetime(2003, 9, 25, 10, 36, 28,
                                       tzinfo=self.brsttz))
+
     def testDateCommandFormatIgnoreTz(self):
         self.assertEqual(parse("Thu Sep 25 10:36:28 BRST 2003",
                                ignoretz=True),
                          datetime(2003, 9, 25, 10, 36, 28))
 
+    def testDateCommandFormatRespectTz(self):
```
pganssle: Pretty sure this will always fail on Windows, so this would need to be changed in the following ways:

1. Add a `@skipIf` to skip this test on Windows.
2. Use [`dateutil.tests._common.TZEnvContext`](https://github.com/dateutil/dateutil/blob/master/dateutil/test/_common.py#L159) for this, rather than just setting the `os.environ` directly like this.

dateutil/parser.py
jstewmon: I'm not sure how this case would be hit if it's moved to the bottom.

Is the compatibility case that `res.tzname` is not `None`, but res.tzoffset is `None`?

Can you provide a date string that I can use to add a test for the compatibility concern?

dateutil/test/test_parser.py
jstewmon: Ah, thanks for the pointer here. I made the test choose the right context manager and timezone name by platform.

dateutil/parser.py
pganssle: I think I meant more in the abstract, I didn't really think much about the implementation. Really, though, I think a better fix is going to be somewhat more complicated, as I mentioned in #318. Basically, I think what should be done is something like this (eliding over the details because I haven't though through the logic fully):

```python
# Code to determine if it's supposed to be `tzlocal()`
if isinstance(ret.tzinfo, tz.tzlocal):
    if ret.tzname() != res.tzname or not tz.datetime_exists(ret):
        if tz.datetime_ambiguous(ret) and tz.enfold(ret, fold=1).tzname() == res.tzname:
            ret = tz.enfold(ret, fold=1)
        elif res.tzoffset == 0:
            ret = ret.replace(tzinfo=tz.tzutc())
        elif res.tzoffset:
            ret = ret.replace(tzinfo=tz.tzoffset(res.tzname, res.tzoffset))
```

It probably needs to be a bit more complicated than that to handle all the edge cases properly, and to disambiguate ambiguous dates, but the logic is at least somewhat clear:

1. If it could be parsed as `tzlocal()` but applying `tzlocal()` to the date doesn't give you the input `tzname` you expected (likely because of a DST mismatch), check if it's just an ambiguous date.
2. If it's an ambiguous date, keep `tzlocal()` and set the `fold` attribute to 1 to disambiguate per PEP-495.
3. If it's not an ambiguous date, try all the other possible ways to parse the date that aren't `tzlocal()`, and if you find a good one, apply it.
4. If you didn't find a good one, that's an unpleasant situation, but to maintain backwards compatibility, just leave `tzlocal()` attached (we can add a `FutureWarning` if need be on this).

I'm thinking the best implementation would be to instead of using `if / elif`, after the `tzinfos` part has been parsed, just start a queue of possible time zones, so like in this case the queue would be something like `[tzlocal(), `tzutc()`], then assign the first one in the list that creates a non-imaginary date, and if none of them create non-imaginary dates, return the first one in the list.

jstewmon: Ok, thanks for the feedback. You have more context than me on the issue, so I'll defer to your expertise, but I thought about the priority a little differently:

1. If an offset was parsed, it should always be the exact offset in the `tzinfo`.
  - If the offset is 0, `tzinfo=tzutc()`, else `tzinfo=tz.tzoffset(res.tzname, res.tzoffset)`
2. Otherwise, if an name was parsed and the name is in `time.tzname`, use the offset for the explicitly named zone.

```python
            elif res.tzoffset == 0:
                ret = ret.replace(tzinfo=tz.tzutc())
            elif res.tzoffset:
                ret = ret.replace(tzinfo=tz.tzoffset(res.tzname, res.tzoffset))
            elif res.tzname:
                try:
                    tzindex = time.tzname.index(res.tzname)
                except ValueError:
                    pass
                else:
                    offset = tzindex and time.altzone or time.timezone
                    ret = ret.replace(tzinfo=tz.tzoffset(res.tzname, -offset))
                    ret = ret.astimezone(tz.tzlocal())
```

This assumes the date is not ambiguous if the date string has a timezone.

If you don't think this is right, I'm happy to let you take over and resolve however you think is best.

jstewmon: After thinking more about the problem, I agree with you.

I was originally thinking about the problem for scenarios where a date string is produced by a machine - in such cases, we can probably expect that a timezone in a date string is significant and should be honored. For example, the date string `'2016-10-10T08:00:00 CST'` means `2016-10-10 08:00:00-06:00` because `CST` is `UTC-06:00`.

However, in many cases the source of a date string may be from user input where the user considers the time to be authoritative and the timezone to be informative but not necessarily correct. I would call this a "colloquial" date string. For example, a user may write `October 10, 8am CST`, expecting the time to be `2016-10-10T08:00:00-05:00` because CST was given to mean US Central time, not exactly Central Standard Time.

My previous proposal would violate the expectations for colloquial date strings because `CST` would be taken explicitly.

I'm now wondering if the behavior of dateutil actually needs to change for this. Similarly to your suggestion on #318, the `parse` caller could generally enforce the offset of local timezones by passing `tzinfos={time.tzname[0]: -time.timezone, time.tzname[1]: -time.altzone}`. Having these examples in the documentation would probably be sufficient to avoid issues like the one that was observed by botocore users in the UK.

pganssle: I think it's reasonable for the behavior to change just so that by default the closest to the "right" thing is inferred from whatever context we have, but I agree with you and would be very happy for any PRs that improve the documentation. If you don't have time, I will make sure that your suggested documentation goes somewhere in there - I think the docs are in need of a major re-organization and cleanup anyway, as they are somewhat hard to follow.

I'm going to leave this PR open for now because I'd like to branch off of it to implement the functionality I described above, and I can use a modified version of the test you added.

pganssle:
New PR at #517

Issue #322, opened 2016-11-11
API break in rrule.weekday 2.6.0


agateau:
This commit bced9222a9caae04924e43b9ffb2d2cdb2488174 changes the signature of rrule.weekday constructor from `(weekday, n=None)` to `(wkday, n=None)`, which breaks compatibility for code explicitly calling the constructor like this: `rrule.weekday(weekday=2)`.

Looking at the code it does not look trivial to fix though, thanks to the class being called like the argument :/

pganssle:
Interesting. I'm fairly certain you're not supposed to actually call `rrule.weekday` directly like that - if you notice, it's not part of `__all__`:

```python
__all__ = ["rrule", "rruleset", "rrulestr",
           "YEARLY", "MONTHLY", "WEEKLY", "DAILY",
           "HOURLY", "MINUTELY", "SECONDLY",
           "MO", "TU", "WE", "TH", "FR", "SA", "SU"]
```

You're really only supposed to use the `MO`, `TU`, `WE`, `TH`, `FR`, `SA`, `SU` constants.

That said, I will change it back to the old way, because it's not like it was changed this way on purpose.

agateau:
The thing is: I have the weekday as a number, so the constants do not work for me. But if you added `weekdays` to `__all__` I could use `weekdays[2](3)` which would be the same as `weekday(2, n=3)`, right?

agateau:
(or I could recreate weekdays in my code, no changes on your side)

pganssle:
Yes, I could expose `weekdays`, but I think for now what I'll do is change it back in case other people are relying on this, and eventually move all the weekday-related code into a `constants` module where I'd expose a method like `get_weekday(x, n=None)` and other useful constants like `EPOCH`.

I don't want to prematurely change the API just to have to deprecate it in the next release.

agateau:
Sounds good. For now I fixed my issue by recreating weekdays, so my code no longer rely on rrule.weekday.

pganssle:
Ah, actually, I don't think there's an easy / reasonable way to change it back, because `weekday` is actually a pretty bad name for the parameter there in the first place, since the class itself is *also* called weekday. That makes it impossible to do things like `super(weekday, self).__init__(*args, **kwargs)`. I could work around this by using `weekdaybase.__init__(self, *args, **kwargs)`, but I think it's bad form to do that just for the purpose of maintaining backwards compatibility with un-exposed parts of the API.

PR #323, opened 2016-11-13
Fix documentation and adjust weekday param

pganssle:
This exposes some new `tz` files for the documentation and addresses the semi-API break in #322.


pganssle:
Oh, I see why this is this way.

PR #330, opened 2016-12-03
Handles cases where count parameter is 0

### LINKED ISSUE #329: Infinite loop in rrule.rrule if count is set to 0
`rrule.rrule` yields values indefinitely if the `count` argument is set to 0. Example:

```
In [2]: for x in rrule.rrule(freq=rrule.DAILY, count=1): print(x)
2016-12-01 17:53:11

In [3]: for x in rrule.rrule(freq=rrule.DAILY, count=0): print(x)
2016-12-01 17:53:15
2016-12-02 17:53:15
2016-12-03 17:53:15
...
```

Even if 0 is not an acceptable value for that parameter, this looks like a bug.

mshenfield:
Replaced any `if count` to `if count is None`.

Where count is checked to stop iteration, moved the check before an item is yielded, and updated the predicate so that it will still yield the correct results.

Added a test for when count is zero.

Closes #329

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -256,12 +256,12 @@ class rrulebase(object):
         n = 0
         for d in gen:
             if comp(d, dt):
-                yield d
-
                 if count is not None:
                     n += 1
-                    if n >= count:
+                    if n > count:
                         break
+
+                yield d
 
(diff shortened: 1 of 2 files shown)


pganssle:
:+1:

PR #332, opened 2016-12-13
Append to skipped_tokens when not val_is_ampm

jbrockmendel:
Fuzzy-parsing drops tokens in cases where a token is initially assessed with val_is_ampm==True (line 941) and later determined not to be (line 950).  A motivating example:

    >>> dateutil.parser.parse('http://biz.yahoo.com/ipo/p/600221.html', fuzzy_with_tokens=True)[1]
    (u'http://biz.yahoo.com/ipo/', u'/', u'.html')

This pull request prevents the "p" from being dropped, so this becomes:

    >>> dateutil.parser.parse('http://biz.yahoo.com/ipo/p/600221.html', fuzzy_with_tokens=True)[1]
    (u'http://biz.yahoo.com/ipo/p/', u'.html')

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -968,6 +968,9 @@ class parser(object):
 
                         res.ampm = value
 
+                    elif fuzzy:
+                        last_skipped_token_i = skip_token(skipped_tokens, 
+                                                    last_skipped_token_i, i, l)
                     i += 1
                     continue
 
@@ -1032,13 +1035,8 @@ class parser(object):
                 if not (info.jump(l[i]) or fuzzy):
                     return None, None
 
-                if last_skipped_token_i == i - 1:
-                    # recombine the tokens
-                    skipped_tokens[-1] += l[i]
-                else:
-                    # just append
-                    skipped_tokens.append(l[i])
-                last_skipped_token_i = i
+                last_skipped_token_i = skip_token(skipped_tokens,
+                                                last_skipped_token_i, i, l)
                 i += 1
 
             # Process year/month/day
@@ -1065,6 +1063,18 @@ class parser(object):
             return res, None
 
 DEFAULTPARSER = parser()
+
+
+def skip_token(skipped_tokens, last_skipped_token_i, i, l):
+    if last_skipped_token_i == i - 1:
+        # recombine the tokens
+        skipped_tokens[-1] += l[i]
+    else:
+        # just append
+        skipped_tokens.append(l[i])
+    last_skipped_token_i = i
+    return last_skipped_token_i
+
 
 
 def parse(timestr, parserinfo=None, **kwargs):
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -969,7 +969,7 @@ class parser(object):
                         res.ampm = value
 
                     elif fuzzy:
-                        last_skipped_token_i = skip_token(skipped_tokens, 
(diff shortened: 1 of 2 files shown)


dateutil/parser.py
```diff
@@ -1067,6 +1065,18 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
 DEFAULTPARSER = parser()
 
 
+def skip_token(skipped_tokens, last_skipped_token_i, i, l):
```
pganssle: This should not be a public module function like this. Can you move it to a private method on `parser`?

I also don't love how it both mutates `skipped_tokens` and returns a value, but that is a secondary concern and if it's private we won't be married to that particular design decision.

pganssle:
Thanks for this contribution, it looks mostly good, but I just want to avoid changing the public API, so if you make the change I mentioned in the line-comment (move `skip_token` to `parser._skip_token`), I will merge.

jbrockmendel:
Just pushed.  Out of curiosity, why keep all the logic in the loop inside parser._parse rather than smaller functions?

pganssle:
@jbrockmendel That's just how it was originally written, I suppose. It's definitely not a deliberate architectural decision on my part, and in fact it's something I'm looking forward to reversing.

Ideally, the functions would be, as appropriate, distributed between `parser.parser` and `parser.parserinfo`, with actual parsing-related stuff (handling of tokens, etc) in the `parser` objects and with things for resolving the actual values in the `parserinfo` objects, then you could subclass and/or instantiate those objects as appropriate to customize the context the parser sees.

Now that time zones are mostly sorted out, my next big goal is to revamp the parser so that it operates like this, and so that there are some simple baseline ways to instantiate `parser` objects to give you greater control over things like the resolution of 6-digit numbers (`'yymmdd'` vs `'yyyymm'` vs. `'yyddmm'`, etc). I'm hoping to get together a working basic proposal and then put it out for comments from users on what they would most like to see.

The reason I didn't want `skip_tokens` as a module function was because if it's a class method, you can customize it by subclassing `parser`. The reason I didn't want it public was because I don't want any changes to the public API until the new parser approach is instituted (since then if we stop supporting that it's a breaking change).

dateutil/test/test_parser.py
```diff
@@ -551,14 +551,20 @@ def testFuzzy(self):
...
+        s1 = "Today is 25 of September of 2003, exactly " \
             "at 10:49:41 with timezone -03:00."
-        self.assertEqual(parse(s, fuzzy_with_tokens=True),
+        self.assertEqual(parse(s1, fuzzy_with_tokens=True),
                          (datetime(2003, 9, 25, 10, 49, 41,
                                    tzinfo=self.brsttz),
                          ('Today is ', 'of ', ', exactly at ',
                           ' with timezone ', '.')))
 
+        s2 = "http://biz.yahoo.com/ipo/p/600221.html"
```
pganssle: Generally I prefer to break these into separate tests unless there's some compelling reason to do them together. Doesn't really matter, though, so I'll just break it out into two tests next time I'm messing around with the parser tests.

This whole file needs to be parameterized anyway.

jbrockmendel:
Makes sense.  I ended up spending a bunch of last week on this, will try to break the results up into bite-sized pull requests and adhere your preferred name-space conventions.

PR #334, opened 2016-12-19
Resolve ymd validation

jbrockmendel:
This branch implements a _ymd method "_resolve_known_month" called from within resolve_ymd.  This does the following:

1) In cases where there is not a month-name but a month can be inferred, infer mstridx.  For example, if we see "31/04/30", then despite not having explicitly seen "Apr", we can infer mstridx as 1.

2) If the month has been pinned down, check which (year, day) assignments imply valid dates.  Using the "31 Apr 30" example, "Apr 30, 2031" is valid, but "Apr 31, 2030" is not.  This branch returns datetime.datetime(2031, 4, 30), whereas the status-quo raises "ValueError: day is out of range for month".
\
Tests for this are implemented, several of which fail under the status-quo.

3) If there are multiple valid options and neither yearfirst nor dayfirst are specified, an optional argument "guess_fail" (default False) raises in order to refuse the temptation to guess.
\
Finally, choices between multiple valid options are based on US conventions, matching the current logic.

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -859,3 +859,56 @@ class ParserTest(unittest.TestCase):
         self.assertEqual(parse(dtstr, dayfirst=True, yearfirst=True),
                          datetime(2015, 9, 25))
 
+
+    def testYMDResolveOrderingConventions(self):
+        # Start with two relatively simple cases.
+        dstr = "Feb 20 14"
+        expected = datetime(2014, 2, 20)
+        self.assertEqual(parse(dstr), expected)
+
+        dstr = "20 Feb 14"
+        expected = datetime(2014, 2, 20)
+        self.assertEqual(parse(dstr), expected)
+
+        # "Feb 28, 2029" and "Feb 29, 2028" are both valid dates.
+        # We prefer the option with the year in the last position
+        dstr = "28 Feb 29"
+        expected = datetime(2029, 2, 28)
+        self.assertEqual(parse(dstr), expected)
+
+
+
+    def testYMDResolveCalendarChecks(self):
+        # "April 30, 2031" is a valid date, but "April 31, 2030" is not.
+        dstr = "April 31 30"
+        expected = datetime(2031, 4, 30)
+        self.assertEqual(parse(dstr), expected)
+
+    def testYMDResolveInferMSTRIDX(self):
+        # Even if we do not find the string "April", we can infer given the
+        # options that 04 is the only viable month.
+        expected = datetime(2031, 4, 30)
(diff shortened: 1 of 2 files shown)


dateutil/test/test_parser.py
```diff
@@ -859,3 +859,56 @@ def testUnambiguousDayFirstYearFirst(self):
...
+
+    def testYMDResolveFails(self):
+        # In the following case, 02 must be the month, so the question
+        # is whether 20 is the year and 14 is the day, or the other way
+        # around.
+        dstr = "20 14 02"
+        expected = datetime(2020, 2, 14)
+        self.assertEqual(parse(dstr), expected)
+
+    def testYMDResolveRaises(self):
```
pganssle: I don't think we need to be making specific promises about what the string format will look like, so there's no need to test against this. I think it's too fragile to have people testing against the specific format of the `ValueError` raised. My plan was to eventually switch these over from `ValueError` to specific error subclasses. I guess I never made an issue for that before, but now I've made one: #335.

jbrockmendel: The specific error messages in testYMDResolveRaises are not especially relevant.  The point of these tests is that en error is in fact raised in resolve_ymd.  In principle I should re-write the test to run resolve_ymd specifically.  I'll change that shortly.

jbrockmendel:
>> are these mostly just architectural and cosemtic changes, or are there actual functional improvements that this is addressing?

The changes affecting the parser class do not affect the logic; this probably obscures the point of the PR, so I'll get rid of the cosmetic changes to focus on the functional improvements.

The meat of this branch is in _ymd.resolve_ymd and the new method _resolve_known_month.  In the status-quo, resolve_ymd tries to determine which of the available tokens is the Y/M/D based first on exclusion restrictions (month<=12, day<=31...), then based on ordering.  This branch applies missing exclusion restrictions before falling back to the ordering.

dateutil/test/test_parser.py
pganssle: @jbrockmendel Since `resolve_ymd` is not part of the public interface, it should not be tested directly. My point is that the tests should, for the most part, be a guarantee of the specified public behavior, so if the tests are relying on the error raising a specific message, downstream users should be able to rely on the error raising a specific message - that's not a guarantee I'm willing to stand behind, so I'd rather not test for it.

Generally, I would just use `self.assertRaises` rather than `self.assertRaisesRegex`, you can even put it in a `with` block:

```python
with self.assertRaises(ValueError):
    something_that_raises_value_error()
```

pganssle:
@jbrockmendel No need to roll back the cosmetic changes, I just wasn't sure if this was actually making any changes to the actual behavior, or if it's intened to just a "behind the scenes" improvement. Seems like you're saying it's just "behind the scenes" improvement, which is fine. I'll take a more in-depth look a bit later.

Issue #214, opened 2016-03-10
Possibility to have a strict dayfirst


jorisvandenbossche:
pandas uses the `dateutil.parser.parse` functions in its [`to_datetime`](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.to_datetime.html) method (and it's what makes this a powerful method to be able to parse all kind of sorts datetime strings!)

But the pandas `to_datetime` function can parse lists (or arrays) of strings at once, and a regular bug report we get is that you can end up with inconsistent parsing within that list (because of the flexibility of dateutil). 
For example:

```
In [20]: pd.to_datetime(["31/12/2014", "10/03/2011"])
Out[20]: DatetimeIndex(['2014-12-31', '2011-10-03'])
```

So in the above, both strings are parsed in another way. Of course, in this case it can be solved by passing `dayfirst=True` (which we pass through to `dateutil.parser.parse`). But say you have a malformed string:

```
In [21]: pd.to_datetime(["31/12/2014", "10/13/2011"], dayfirst=True)
Out[21]: DatetimeIndex(['2014-12-31', '2011-10-13'])
```

Regardless of the `dayfirst=True` that is passed, not all strings are parsed with using day first, but you end up with a mixture of using dayfirst or not depending on the string.

So our **question is if it would be possible to be able to specify that dayfirst/yearfirst have to be strict** (for example with an extra keyword), so that if the provided string cannot be parsed following the given dayfirst/yearfirst argument, that an error is raised instead of parsing the string anyway without following dayfirst/yearfirst.

The original issue is here: https://github.com/pydata/pandas/issues/3341

pganssle:
Hm. I suppose this is possible, but I'm a bit concerned about the proliferation of options passed directly to the parser like this, which is why I rolled back the `smart_defaults` and `fallback_on_invalid_dates` options in #191 before the 2.5.0 release.

In my opinion, a lot of the proliferation of these command line arguments could be mitigated by giving users access to the date data *before* the date is constructed. Ideally, this would even allow you to generate an `strptime` string on the first run in `pandas`, which can then be applied to the remainder of the dates (and you can re-run the parser if you get a `ValueError`.

@jorisvandenbossche Obviously it'll take a bit longer than implementing a "strict" dayfirst, but do you think something like this will be an acceptable solution?

jorisvandenbossche:
@pganssle Thanks for your quick reply!

[...]

PR #340, opened 2017-01-09
Allow to parse reversed ISO format

cedk:
We use the parser for [Tryton](http://www.tryton.org/) date widget and we have often user who are trying to encode date in this format (locale format without any separator for speed).


pganssle:
@cedk Why can't you use `dayfirst` for this? I'm *very* wary of changing this behavior, because the last time I made a change to something I thought didn't matter (which, IIRC, you were the one who suggested :P), it caused no end of heartache for people who were relying on the more or less undocumented fallback behavior of the parser.

cedk:
@pganssle because `dayfirst` is for ambiguous 3-interger. Here it is about input with length >=8.
About regression, I do not think it introduce any regression because without the new switch case the result of the parser is `None`.

ohuisman:
For us this is an important improvement.

dateutil/parser.py
```diff
@@ -732,11 +732,16 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
                             res.second, res.microsecond = _parsems(s[4:])
 
                     elif len_li in (8, 12, 14):
-                        # YYYYMMDD
+                        # YYYYMMDD or DDMMYYYY
                         s = l[i-1]
-                        ymd.append(s[:4])
-                        ymd.append(s[4:6])
-                        ymd.append(s[6:8])
+                        if int(s[4:6]) <= 12 and int(s[6:8]) <= 31:
```
pganssle: Hm... I think the logic on this part of the parser is not great anyway, so we don't have to worry about this so much, but consider:

```python
>>> parse('20050931')
Traceback: [...]
ValueError: day is out of range for month
```

This means the parser would fail for the unambiguous date '20050931', which is unambiguously May 20, 931. That said, since *all* the dates surrounding May 20, 931 would parse incorrectly if rendered in this format, this might be a reasonable step to take. There are various simple ways to mitigate this, but I don't think they're worth doing at the moment, I think most of this logic is going to get redone int he end.

pganssle:
Hmm.. You raise a good point about this returning `None`. That said, part of the problem is that occasionally people count on `dateutil`'s failure behavior, not just the success criteria. See, for example, #214. 

At this point, I think I'm leaning in favor of this, but I'm going to delay merging until a feature release anyway, so there is time to mull it over anyway. Is it possible that `MMDDYYYY` *and* `DDMMYYYY` are in common use, worldwide? If so, we  may want to make `dayfirst` work with this.

jarekwg:
Yes please!
We use dateutil in our csv parser which is customer facing. We are in Australia.
Customers will be customers and may enter dates as DD-MM-YYYY or YYYY-MM-DD.
The `dayfirst` flag fixes the former case for us, but breaks the latter. Would be nice if we could have any of:
- a flag to specify that month is always in the middle.
- an option that makes the parser understand Australian date format automatically.
- an option to list all formats in which the date should be parsed: eg `allowed_formats=['DMY', 'YMD']`.

mcarans:
For those who need this functionality and find this issue, you can either use higher level functions I've implemented here:
https://github.com/OCHA-DAP/hdx-python-utilities#date-parsing-utilities

Or if you want the same interface as dateutils, then you use hdx.utilities.dateparse.parse which is modified for ambiguous dates to be day first D/M/Y where there are values in front of the year and day last Y/M/D where there are values after the year and for fuzzy_with_tokens, returns datetime, skipped tokens, date tokens.

lrq3000:
@dateutil why did you remove mcarans comment with a link to [his implementation](https://github.com/OCHA-DAP/hdx-python-utilities#date-parsing-utilities) of the very idea we are discussing here? It could both be a workaround for those who really need this meanwhile, and it could serve as inspiration for a PR for this issue (licenses are compatible). I think that is very pertinent and helpful.

pganssle:
@lrq3000 He spammed basically identical messages like this in multiple threads, and issues and PRs for dateutil are not really the place to advertise your project.

mcarans:
[One issue](https://github.com/dateutil/dateutil/issues/998) and [one PR](https://github.com/dateutil/dateutil/pull/999) was my own (on returning date tokens not just skipped tokens) and I looked for issues and PRs on YMD DMY as I fixed that in my library for my own usage and thought others could benefit from it. 

As I've said these issues should be fixed upstream in dateutil. I tried an alternative library, dateparser, but it was not reliable, so at this time dateutil has the best parser out there. However, it seems nigh on impossible to get PRs related to improving the parser approved. I was told "As hard as it is for me to say when you've obviously put hard work into this fix, but I think we will probably not accept any contribution like this. I really appreciate your taking the time to work on this and to help improve dateutil, so I do not want to waste your time by misleading you into thinking that it's likely that this would get merged with some tweaks."

pganssle:
@mcarans Yes, it is a tough situation and I appreciate you trying to remedy it. I also applaud your enthusiasm to create your own library.

That said, the dateutil issue and PR tracker exist to facilitate the development of `dateutil`, not as a help forum for people having trouble with `dateutil` (despite the fact that some people use it that way). "Here's my library that fixes this problem" doesn't help with that goal, and it sends a notification e-mail to a bunch of people watching these existing issues *because they want to know if `dateutil` has resolved the issue*, not because they are looking for replacements.

I realize there was no malicious intent, but I consider them off-topic for this forum. I am going to hide this thread (including my own replies) as off-topic now, I would like to advise people to not continue to create noise on this or other existing tickets by attempting to appeal this decision - that would be discourteous to people trying to follow the issue who do not care about this drama.

mcarans:
@mariocj89 I notice the latest release was done by you. Are you now maintaining dateutil? If so, would you consider making it support reversed ISO format? I had to make a new library with copy and paste from some of dateutil to add this support and it is a pain to keep it up to date with new dateutil releases.

mariocj89:
Hi @mcarans, changes in the parser are quite tricky to get right and not impact a fleet of users. We are leaning more towards having something like a chain parser that will allow you to set a list of formats to try out in order. We will probably jump to get that after the next release, if you know the possible format that will be the recommended way to parse datetime str.

PR #347, opened 2017-02-06
fixes #346: wrong addition of relativedelta

### LINKED ISSUE #346: wrong implementation of adding relativedelta's
Excerpt of `dateutil.relativedelta` for Python3.5, `dateutil==2.6.0`:

    def __add__(self, other):
        if isinstance(other, relativedelta):
            return self.__class__(years=other.years + self.years,
                                 ...
                                 hour=other.hour or self.hour,
                                 minute=other.minute or self.minute,
                                 second=other.second or self.second,
                                 ....

The problem is pretty simple: while for the plural version this might be correct, this is wrongly implemented for absolute minute/hour/second etc.

The problem is when `other.minute` is 0. That is, forcing the minute to be 0 cannot work. The default value is None for a reason :).

So instead of:

    minute=other.minute or self.minute,

it should be:
  
    minute=other.minute if other.minute is None else self.minute,

kootenpv:
testing `is not None` instead of truthiness

### DIFF
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -311,14 +311,22 @@ class relativedelta(object):
                                  microseconds=(other.microseconds +
                                                self.microseconds),
                                  leapdays=other.leapdays or self.leapdays,
-                                 year=other.year or self.year,
-                                 month=other.month or self.month,
-                                 day=other.day or self.day,
-                                 weekday=other.weekday or self.weekday,
-                                 hour=other.hour or self.hour,
-                                 minute=other.minute or self.minute,
-                                 second=other.second or self.second,
-                                 microsecond=(other.microsecond or
+                                 year=(other.year if other.year is not None
+                                       else self.year),
+                                 month=(other.month if other.month is not None
+                                        else self.month),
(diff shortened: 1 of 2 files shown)


pganssle:
Ah, yes, this seems like the right thing to do, good catch. Any chance you can add some test cases?

kootenpv:
I rewrote some of my code as relativedelta, and my tests were not passing. Having installed this branch it was solved :)

I added some tests that should verify "AbsoluteAddition".

kootenpv:
Is this going to be included?

pganssle:
@kootenpv Sorry, haven't had much time to do `dateutil` stuff. I'll make sure to look at this tonight or tomorrow.

pganssle:
LGTM.

kootenpv:
Hey, I'm wondering what the release schedule will be? I'd like to depend again on the main branch rather than on my own :) Couldn't find information on the release cycles, that's why I ask :)

pganssle:
@kootenpv I've been bad about that sort of thing, but I'll try to cut a release reasonably soon, want to get a few more bugfixes and such in first.

PR #354, opened 2017-03-27
add support for comparing to native python timezones

thehesiod:


thehesiod:
I think this can be made generic by doing this on the base class and always comparing `utcoffset()` of each

pganssle:
Hmm... I'm a bit dubious about the idea of calling `other.utcoffset(None)` like that. For the moment, all `datetime.timezone` objects are static, but one could easily subclass `timezone` and create a non-monotonic zone, and `other.utcoffset(None)` would fail.

PR #358, opened 2017-04-07
Removed deprecation warnings

thierryba:


pganssle:
@thierryba @jdufresne What deprecation warnings is this removing? I'm not seeing any warnings in the CI for other PRs, nor in my own instance of python 3.6.

I'm not really opposed to these changes either way, but I can't replicate the issue.

jdufresne:
Take a look at the release notes for Python 3.6:

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases. (Contributed by Emanuel Barry in bpo-27364.)

jdufresne:
@pganssle I too was having trouble consistently reproducing the warnings, so I dug into this a bit deeper. It looks like some combination of leading underscores and `__pycache__` means that the warnings will only appears sometimes. I'm going to follow through with a bug report to python.org about it. In the mean time, you can consisently see the warnings by using the following script from a checkout of the `master` branch. **WARNING: Script will run `git clean -dxff` and could remove unsaved work.**

```bash
#!/bin/bash

set -e
set -x

git clean -dxff

ROOT=$(git rev-parse --show-toplevel)
NAME=$(basename ${ROOT})

VENV=~/.venv/${NAME}
rm -rf ${VENV}
python36 -m venv ${VENV}
. ${VENV}/bin/activate
python -b -Wall setup.py test
```

After running this, I see the following at the top of my test output:

```
.../dateutil/dateutil/tz/win.py:197: DeprecationWarning: invalid escape sequence \{
  tzkeyname = text_type("{kn}\{name}").format(kn=TZKEYNAME, name=name)
.../dateutil/dateutil/tz/win.py:247: DeprecationWarning: invalid escape sequence \{
  tzkeyname = text_type('{kn}\{sn}').format(kn=TZKEYNAME,
.../dateutil/dateutil/parser.py:50: DeprecationWarning: invalid escape sequence \.
  _split_decimal = re.compile("([\.,])")
```

jdufresne:
Python issue: https://bugs.python.org/issue30091

pganssle:
@jdufresne Ah, interesting. That kinda seems like they are going to make non-backwards-compatible changes to Python at some point in the future, but I'd heard that there "won't be a Python 4.0". I guess we'll see if they break SemVer or the idea of keeping everything on Python 3.

Thanks fr the info.

PR #362, opened 2017-04-17
Drop support for Python 2.6

jdufresne:
I would humbly like to suggest dateutil drop support for Python 2.6.

The last release of Python 2.6 was 2013-10-29, over 3 years ago. It is
no longer receiving security fixes.

https://www.python.org/dev/peps/pep-0361/

The pip project itself has recently dropped support for 2.6. Their
numbers estimate that Python 2.6 accounts for ~2% of their downloads.

pypa/pip#4343

For projects that still use Python 2.6, they can continue to pip install
an older version.

I've tried my best to remove as much 2.6 specific code as I can,
including the 'Programming Language :: Python :: 2.6' trove classifier
from setup.py. I've also removed Travis CI testing, which should result
in faster testing and fewer wasted resources.

Code changed:

- Removed Python2.6 from testing configuration
- setup.py cleanups due to fewer version complications
- Removed unittest2 dependency and monkey patching
- Use set literals
- Use dict comprehension
- Remove total_seconds workaround
- Remove TarFile.open() context manager workaround

Thanks for considering.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 cache: pip
 python:
-  - "2.6"
   - "2.7"
   - "3.2"
   - "3.3"
@@ -25,7 +24,6 @@ before_install:
 
 install:
   - pip install six
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install unittest2; fi
   - ./ci_tools/retry.sh python updatezinfo.py
 
 script:
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+UNRELEASED
+----------
+- **Dropped support for Python 2.6.**
+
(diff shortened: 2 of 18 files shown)


pganssle:
I'm amenable to this, but I'd like to get out at least one more bugfix release on timezone support before Python 2.6 is dropped.

pganssle:
@jdufresne If you have time, can you rebase and update this PR? I think we can go ahead and merge.

jdufresne:
Thanks @pganssle. I've rebased the changes on to the latest master. All additional feedback welcome.

pganssle:
Thanks @jdufresne. So far it looks good but I'll give it a more thorough review and merge tonight.

pganssle:
OK, LGTM!

PR #365, opened 2017-04-17
Add Python 3.6 to CI testing

jdufresne:


pganssle:
Hmm... Has the fold support always failed on Windows? I could have sworn I tested it.

jdufresne:
> Hmm... Has the fold support always failed on Windows? I could have sworn I tested it.

Unfortunately, I don't have easy access to a Windows box. So it'll be difficult for me to iterate on testing this without going through the slow process of appveryor builds.

pganssle:
> Unfortunately, I don't have easy access to a Windows box. So it'll be difficult for me to iterate on testing this without going through the slow process of appveryor builds.

Don't worry about it. I'll figure out what's going on.

pganssle:
Fixed the CI issues with #392, turns out some of the tests on Windows made assumptions about equality semantics that don't turn out to be true after PEP-495 is implemented. Kinda disturbing that they work in Python < 3.6, though, since `dateutil`'s backwards compatibility objects should handle this correctly.

PR #367, opened 2017-04-17
Remove unnecessary coerce to unicode/bytes code

jdufresne:
As the file contains `from __future__ import unicode_literals`, the bare
string is already a unicode object. No need to coerce it. Further, it
can always be converted to a bytes object with `.encode()`, no need to
catch `NameError`.


pganssle:
Looks good to me.

PR #368, opened 2017-04-17
Clean up unused imports

jdufresne:
Discovered using flake8. If you'd like, I can add routine flake8 testing
to the CI configuration to prevent new unused imports.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -35,7 +35,7 @@ import time
 import collections
 import re
 from io import StringIO
-from calendar import monthrange, isleap
+from calendar import monthrange
 
 from six import text_type, binary_type, integer_types
 
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -2,12 +2,11 @@
 from __future__ import unicode_literals
 from ._common import unittest
 
-from datetime import datetime, timedelta, date
+from datetime import datetime, timedelta
 
 from dateutil.tz import tzoffset
 from dateutil.parser import *
 
-import six
 from six import assertRaisesRegex, PY3
 from six.moves import StringIO
 
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -13,13 +13,8 @@ import time
 import sys
 import os
 import bisect
-import copy
 
-from operator import itemgetter
-
-from contextlib import contextmanager
-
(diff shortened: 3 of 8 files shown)


pganssle:
I think flake8 testing would be useful, though I'm not sure if we're PEP8-compliant to begin with, so that might throw things off - among the reasons I haven't done it yet. If it can start as an optional CI test or something that simply doesn't introduce *new* PEP8 violations, that seems like it could work.

This one I'm going to have to look into the history to see why these imports are left around. I don't think they're part of the documented public interface, though, so I see no reason why we shouldn't remove them.

dateutil/parser.py
```diff
@@ -35,7 +35,7 @@
 import collections
 import re
 from io import StringIO
-from calendar import monthrange, isleap
```
pganssle: The last commit that this was needed in was ad8c744, removed in 1622e2b4c596. This can be removed.

dateutil/test/test_parser.py
```diff
@@ -2,12 +2,11 @@
 from __future__ import unicode_literals
 from ._common import unittest
 
-from datetime import datetime, timedelta, date
```
pganssle: Can't find any time where this was actually used.

dateutil/tz/tz.py
```diff
@@ -13,13 +13,8 @@
 import sys
 import os
 import bisect
-import copy
 
-from operator import itemgetter
-
-from contextlib import contextmanager
```
pganssle: I'm pretty sure these are leftovers from things moved out of this module, good to go.

dateutil/tz/win.py
```diff
@@ -12,7 +12,6 @@
     # ValueError is raised on non-Windows systems for some horrible reason.
     raise ImportError("Running tzwin on non-Windows system")
 
-from ._common import tzname_in_python2, _tzinfo
```
pganssle: This is no longer necessary because tzrangebase already includes these, good to go.

dateutil/zoneinfo/__init__.py
```diff
@@ -1,9 +1,5 @@
 # -*- coding: utf-8 -*-
-import logging
-import os
 import warnings
-import tempfile
-import shutil
```
pganssle: I think it's fine to remove all of these.

PR #369, opened 2017-04-17
Remove deprecated dateutil.tzwin module

jdufresne:
All code is using dateutil.tz.win instead.

### DIFF
--- a/dateutil/tzwin.py
+++ b/dateutil/tzwin.py
@@ -1,2 +1,0 @@
-# tzwin has moved to dateutil.tz.win
-from .tz.win import *


pganssle:
Not sure why you think this module is deprecated. I moved its contents into `tz.win` for various modules, but there are undoubtedly downstream users who are stll using `tzwin`. I don't see that there's any big cost in leaving this as is for now. If anything, I'd want to give a long deprecation period before removing it.

jdufresne:
Ah, I see. Thanks. Sorry about that.

PR #371, opened 2017-04-17
Avoid duplicate test names

jdufresne:
The two `testToStrYearlyByMonth` tests were functionally identical, so I
combined them into one.

The other tests used slightly different test data, so kept both
instances, but could also delete one if preferred.


pganssle:
Hm. Interesting. Good catch.

At some point I'd like to move the test suite over to pytest and parametrize these tests, so this sort of thing won't be such an issue after that.

PR #372, opened 2017-04-17
Use generator expressions to avoid creating an unnecessary list in memory

jdufresne:
https://www.python.org/dev/peps/pep-0289/
https://docs.python.org/3/reference/expressions.html#generator-expressions


pganssle:
Thanks for all these cleanup PRs @jdufresne

PR #373, opened 2017-04-17
Use 2 spaces after top level definitions

jdufresne:
Brings code closer to flake8 compliance by passing rule:

E305 expected 2 blank lines after class or function definition

http://pycodestyle.pycqa.org/en/latest/intro.html#error-codes


pganssle:
I'm going to have to pull this into my local version and see how it looks. I'm not sure that I really like E305.

pganssle:
Yeah, this looks fine.

PR #374, opened 2017-04-17
Convert file to utf-8 encoding as it is the Python 3 default

jdufresne:


pganssle:
Did you actually change the file encoding, or just the shebang thing at the top (I don't know what the encoding version of a shebang is called)?

I guess according to [this link](https://stackoverflow.com/questions/7048745/what-is-the-difference-between-utf-8-and-iso-8859-1/7048774#7048774) the two actually encode ASCII identically, so maybe it's the same thing.

jdufresne:
AFAICT, the file only contains ascii, which, I believe, is a subset of both encodings. If true, there is nothing to convert.

```
$ file -i dateutil/parser.py
dateutil/parser.py: text/x-python; charset=us-ascii
```

I also used [this guide](https://stackoverflow.com/questions/3001177/how-do-i-grep-for-all-non-ascii-characters-in-unix/9395552#9395552) to check for non-ascii characters and did not find any.

Please let me know if I'm mistaken.

pganssle:
Independently confirmed that this is what the diff looks like even if you explicitly re-encode as UTF-8. LGTM.

PR #375, opened 2017-04-17
Fix all instances of flake8 F841

jdufresne:
F841 local variable `name` is assigned to but never used

Now use seek to skip over lzh_leapcnt pairs in tz.py. As the values are
unused, no need to read them and convert them to Python objects. It is
only wasted cycles and memory. Skip over with seek() instead.

Rewrote testTzwinStar based on surrounding similar tests. Previous
version did not achieve an interesting result.

I believe non of the removals have intended side effects, If I'm
mistaken, please let me know.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -550,11 +550,8 @@ class parser(object):
         """
 
         if default is None:
-            effective_dt = datetime.datetime.now()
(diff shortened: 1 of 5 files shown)


pganssle:
I think this sort of thing is less important in tests, where sometimes you're making sure that the side-effects are or are not throwing an error, but I don't see any obvious problems with these changes.

PR #378, opened 2017-05-04
Fix copy/paste typo in rrulestr

mlorant:
Simple fix, that changes the wording of one exception, that can be misleading


pganssle:
Good catch @mlorant mlorant!

PR #381, opened 2017-05-12
Fixes #348

### LINKED ISSUE #348: Converting an rrule to a string and back results in loss of timezone information in the _dtstart attribute
In the following script:

```python
from dateutil.rrule import rrule, rrulestr, DAILY

start = parse("8 Feb 2017 14:00 UTC")
my_rrule = rrule(DAILY, dtstart=start)

my_rrule2 = rrulestr(str(my_rrule))

assert my_rrule._dtstart == my_rrule2._dtstart
```

the assertion leads to a

```python
TypeError: can't compare offset-naive and offset-aware datetimes
```
This is because `my_rrule._dtstart` has the representation

```python
datetime.datetime(2017, 2, 8, 14, 0, tzinfo=tzutc())
```
whereas the 'reconstructed' `my_rrule2._dtstart` is

```python
datetime.datetime(2017, 2, 8, 14, 0)
```
In other words, converting an `rrule` to a string and back (using `__str__` and `rrulestr`, respectively) leads to a loss of time zone information of the `_dtstart` attribute. This would seem undesirable, since the original and reconstructed object are no longer equivalent.

sanyi:


pganssle:
I'm uncertain about this patch for a few reasons:

1. I am not sure where in the RFC it is specified how time zones should be included in RRULE strings, but I the ones I see them seem to be more consistent with the iCal time zone format, not %Z, in which case this would not generate RFC-compliant rrule strings.
2. My understanding is that there is very limited support in `datetime` for *parsing* `%Z`, which always makes me uncertain about dumping strings with %Z time zones.
3. This is a lossy method for dumping time zones. Consider a rule where the zone for DTSTART is EST and the rule includes times both in EST and EDT. If you were to load this rrulestring and attach a UTC-5 zone to it, all the EDT times it generates would be incorrect.

I think basically this needs to come down to what the RFC says, so any patches should cite the RFC's method, but honestly it feels like including time zones in RRULEs is just folly.

sanyi:
@pganssle I am reconsidering this and I think I agree with you. There are better ways to solve this. I am closing the request.

Issue #343, opened 2017-01-20
Parser.weekday does not support short days with less than 3 chars


arcanefoam:
In German, the short day versions only have 2 letters. So even if I supply my own WEEKDAYS dict with the correct locale, parser.weekday does not work (it tests if the name passed has len >= 3).

pganssle:
I think this should be fixed by #382. Thanks @jonemo!

arcanefoam:
Great! Thanks @jonemo for the patch.

PR #382, opened 2017-05-22
Allow for two day names shorter than 3 letters (#343)

jonemo:
Possibly a fix for issue #343 

Issue summary: Short day names had to be at least 3 characters long to be recognized, which works for the default day names (`Mon`, `Tue`, ...) but not necessarily for custom `WEEKDAYS` names such as German day name abbreviations (`Mo`, `Di`, ...).

* This PR changes the check from a comparison against a constant value (`3`) to a value calculated from the entries of the `WEEKDAYS` setting. 
* Since the same applies to `MONTHS`, applying the equivalent change there. 
* Added a test for the case of German day names, which fails before the change and passes after.

A possible next step would be to cache the return value of `min(len(n) for n in self._weekdays.keys())` to avoid re-computation. Please let me know if that would be preferred here.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -309,7 +309,7 @@ class parserinfo(object):
         return name.lower() in self._jump
 
     def weekday(self, name):
-        if len(name) >= 3:
+        if len(name) >= min(len(n) for n in self._weekdays.keys()):
             try:
                 return self._weekdays[name.lower()]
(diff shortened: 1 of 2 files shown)


pganssle:
@jonemo Awesome, looks good to me. I think that caching the `min` call is a good idea, but there's no rush there.

PR #385, opened 2017-05-24
Adding python 3.6 to CI now that it is officially released

TheGhostHuCodes:
Noticed that we were testing `3.6-dev` with allowed failure, but now the released `3.6`.


TheGhostHuCodes:
Whoops, should probably look at the existing PRs, before shooting stuff off. Duplicate of #365. Closing.

pganssle:
@TheGhostHuCodes Yeah, there's some problem with the `fold` support on Windows I think. I haven't had a chance to figure out why that's happening.

PR #388, opened 2017-05-27
Remove unnecessary dict.keys() calls

jdufresne:
iter(dict) is equivalent to iter(dict.keys()). Fixed all code to simply
act on the dictionary. Provides simpler, more concise syntax and avoids
the extra function call.

Inspired by Lennart Regebro's PyCon 2017 presentation "Prehistoric
Patterns in Python". Available at:

https://www.youtube.com/watch?v=V5-JH23Vk0I

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -309,7 +309,7 @@ class parserinfo(object):
         return name.lower() in self._jump
 
     def weekday(self, name):
-        if len(name) >= min(len(n) for n in self._weekdays.keys()):
+        if len(name) >= min(len(n) for n in self._weekdays):
             try:
                 return self._weekdays[name.lower()]
             except KeyError:
@@ -317,7 +317,7 @@ class parserinfo(object):
         return None
 
     def month(self, name):
-        if len(name) >= min(len(n) for n in self._months.keys()):
+        if len(name) >= min(len(n) for n in self._months):
             try:
                 return self._months[name.lower()] + 1
             except KeyError:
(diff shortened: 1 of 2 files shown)


pganssle:
Hm... I'm conflicted about this. The `keys` method makes it clearer exactly what you're trying to accomplish, so this kind of an "explicit is better than implicit" thing.

In the video, he states without proof that it's better not to use `keys`, but I really don't think the semantics of dictionary iteration is particularly obvious. If I were new to Python, my first thought would be that iterating over a dictionary would return `dict.items()`, not `dict.keys()`.

In terms of performance, I think it's actually not that important either, because the plan with the parser code is to eventually cache that `min` call (so `.keys()` would never be called more than once in that situation). I also think it's unlikely that `tzical.keys()` will ever be called particularly frequently, and I find `list(dict)` returning `list(dict.keys())`  even *more* strange and opaque than iteration returning `keys` is.

jdufresne:
I understand your point and agree it may not be immediately obvious to someone new to Python.

For me, I've seen this pattern enough now that I read `iter(dict)` as equivalent to `iter(dict.keys())` without much thought. So it may just be a matter of what one is used to seeing.

pganssle:
Yeah, I think it's probably more or less readable depending on your experience. This is a pretty simple codebase, so I think given that it's unlikely to have a major impact on performance, we should just leave it as-is.

If there's some compelling evidence that this is slowing people down, or if there's some other more compelling reason to do this, maybe a compromise position would be to make these changes, but add a comment to the effect that these are equivalent to `keys()` calls.

PR #389, opened 2017-06-02
Make classes hashable 

mrigor:
This builds on https://github.com/dateutil/dateutil/pull/296 by making those classes hashable by explicitly providing the `__hash__` method

classes touched:
- relativedelta
- weekday

### DIFF
--- a/dateutil/_common.py
+++ b/dateutil/_common.py
@@ -24,7 +24,14 @@ class weekday(object):
             return False
         return True
 
-    __hash__ = None
+    def __hash__(self):
+        return hash((
+          self.weekday,
+          self.n,
+        ))
+
+    def __ne__(self, other):
+        return not (self == other)
 
     def __repr__(self):
         s = ("MO", "TU", "WE", "TH", "FR", "SA", "SU")[self.weekday]
@@ -32,3 +39,5 @@ class weekday(object):
             return s
         else:
             return "%s(%+d)" % (s, self.n)
+
+# vim:ts=4:sw=4:et
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -512,7 +512,25 @@ class relativedelta(object):
                 self.second == other.second and
                 self.microsecond == other.microsecond)
 
-    __hash__ = None
+    def __hash__(self):
+        return hash((
+            self.weekday,
+            self.years,
+            self.months,
+            self.days,
+            self.hours,
+            self.minutes,
+            self.seconds,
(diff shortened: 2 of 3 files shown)


ngould:
👍

pganssle:
Hmm... Would this be considered a break in backwards compatibility, or no? I don't know of any situations where you would count on something *not* being hashable, so I'm inclined to say that this does not break backwards compat, but I'm curious to know if anyone can think of a counter-example where the change in this behavior might be a problem.

ngould:
@pganssle relativedelta's were hashable as of dateutil 2.2. At some point between 2.2 and 2.6 they stopped being hashable, seemingly because of efforts at python 3 compatibility (see #296). 

So the backwards compatibility break already occurred. This PR would fix it going forward.

pganssle:
Actually, looking back on #296, I see that I explicitly decided that this behavior was to be considered "undefined" at the time of merging:

> I would normally say we should add some tests, but I don't know that we want to signal that this behavior is definitely permanent. I could see the some possibility for adding a hash function in to the time zone objects, at least, at some point in the future, so I'm just going to merge as is.

I think it's reasonable to say that in the unlikely event that anyone was counting on this specific behavior, they shouldn't have been. We can merge this as part of `2.7.0`. I'm going to do a bit of work today on the time zones then fork off a `2.6.x` branch to wrap up bugfixes to release before 2.7.0, then I can start merging some of these features into master. Perhaps for a 2.7.0 release we should finalize the hashing behavior of all objects.

rowillia:
@pganssle @ngould Right, before https://github.com/dateutil/dateutil/pull/296 the __hash__ behavior was undefined.  In Python 2 the __hash__ and __eq__ implementations disagreed (hash would default to object id).  In Python 3 these objects weren't hashable.  That PR simply made the behavior consistent between 2 and 3.

dateutil/_common.py
```diff
@@ -24,11 +24,20 @@ def __eq__(self, other):
...
+        return not (self == other)
 
     def __repr__(self):
         s = ("MO", "TU", "WE", "TH", "FR", "SA", "SU")[self.weekday]
         if not self.n:
             return s
         else:
             return "%s(%+d)" % (s, self.n)
+
+# vim:ts=4:sw=4:et
```
pganssle: In the future I think maybe don't add this comment at the bottom. It's already there in a few other files, so it's not a big deal, but I think editor-specific detritus probably doesn't belong in upstream repos.

PR #393, opened 2017-06-04
Fix improper indent in parser.

### LINKED ISSUE #333: Indentation typo in _parse?
parser.py, lines 794-802:

    if idx == 1:
        res.minute = int(value)

        if value % 1:
            res.second = int(60*(value % 1))
        elif idx == 2:
            res.second, res.microsecond = \
                                    _parsems(value_repr)
            i += 1

The "idx == 2" branch will never be reached.  I'm pretty sure this is supposed to be:

    if idx == 1:
        res.minute = int(value)

        if value % 1:
            res.second = int(60*(value % 1))
    elif idx == 2:
        res.second, res.microsecond = \
                                _parsems(value_repr)
        i += 1

Maybe the "i+=1" is supposed to be applied in both branches?

pganssle:
Fixes #333. Still not 100% sure if the `i += 1` that was never being reached is strictly necessary, but I think that that is supposed to advance the tokens properly. Not entirely sure why the values were already being advanced properly, but all the tests seem to pass, so I think we're good. 

cc: @jbrockmendel @mrForce

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -798,14 +798,16 @@ class parser(object):
                         if idx == 0:               # h
                             res.minute = int(value)
 
-                            if value % 1:
-                                res.second = int(60*(value % 1))
+                            sec_remainder = value % 1
+                            if sec_remainder:
+                                res.second = int(60 * sec_remainder)
                         elif idx == 1:             # m
                             res.second, res.microsecond = \
                                 _parsems(value_repr)
 
-                        if idx == 0 or idx == 1:
-                            i += 1
+                        # We don't need to advance the tokens here because the
+                        # i == len_l call indicates that we're looking at all
+                        # the tokens already.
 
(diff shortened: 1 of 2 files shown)


pganssle:
Actually, I figured out why the `i += 1` call was not important - in the outer `if`, we're already checking that `i == len_l`, which means that `l` consists of *all* the tokens in the string, so advancing the token index by 1 just moves us past the end of the string.

Presumably it was originally in the *outer* branch because the author wanted `i` to move past the end of the string in case something later moved the token index backwards? I could see keeping it in there for that purpose, but I'm not really sure it makes sense, so I'm just dropping it entirely.

jbrockmendel:
Looks good, thanks for looking into this.

pganssle:
@jbrockmendel Thanks for the report!

PR #398, opened 2017-06-09
Replace all star imports with explicit imports

jdufresne:
Fixes all flake8 error F405:

F405 'parse' may be undefined, or defined from star imports: dateutil.parser


pganssle:
LGTM

PR #408, opened 2017-07-04
Quickfix: assertion before setting mstridx

jbrockmendel:
If that branch was reached, asserting `mstridx == -1` immediately after setting `mstridx = len(ymd)-1` should always fail (`ymd` is not empty at this point).  I'll circle back to add a test case to cover this.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -854,8 +854,8 @@ class parser(object):
 
                                 if value is not None:
                                     ymd.append(value)
-                                    mstridx = len(ymd)-1
                                     assert mstridx == -1
+                                    mstridx = len(ymd)-1
                                 else:
                                     ymd.append(l[i])
 


pganssle:
This is a bit bizarre. I think maybe we should figure out if it's even possible for that assertion to be incorrect and maybe raise a more appropriate message than `AssertionError`.

jbrockmendel:
Reading the code.  I spent some time refactoring common patterns to try to cut down the big while loop in `parse`.

If I disable the error-catching, I can trigger this assertion with "2016-15-May".  It looks like it only gets reached when the month is at the end, so it makes sense that it would be rare.

jbrockmendel:
This PR was made with GH's editor as a quick-fix.  Replacing it with a fully-fledged PR #409.

Issue #402, opened 2017-06-27
It's impossible to parse ISO-style and European-style dates at the same time


houseofsuns:
I've got the following two representation for the 1st December: "1.12.2017" and "2017-12-01". The first is European-style (that means day first) and the second is in ISO format. 

I failed at finding an invocation of dateutil.parser.parse() that works on both inputs. The crucial parameter seems to be dayfirst. If set to True the European-style input is parsed correctly and if set to False the ISO-style date is parsed correctly.

After thoroughly studying the documentation I've come to the conclusion that this is currently impossible to do and constitutes a gap in parser functionality.

pganssle:
@houseofsuns Correct. There is an ongoing effort to improve the way the parser works so that downstream users can have more effective fine-grained control over the various fallbacks.

For now, if you know that your dates are in one of two formats, it's easy enough to use a `try`/`catch` block with `strptime`. Here's a pretty extensible helper function to do this until a new more flexible parsing framework is established:

```python
from datetime import datetime
from dateutil.parser import parse
def parse_formats(dt_str, fmt_strs, **parser_kwargs):
    for fmt in fmt_strs:
        try:
            return datetime.strptime(dt_str, fmt)
        except ValueError:
            pass
    
    return parse(dt_str, **parser_kwargs)


if __name__ == "__main__":
    fmt_strs = ['%d.%m.%Y', '%Y-%m-%d']
    dt_pairs = [
        ('1.12.2017', datetime(2017, 12, 1)),
        ('2017-12-01', datetime(2017, 12, 1)),
        ('Dec. 01, 2017', datetime(2017, 12, 1))
    ]

    for dt_str, dt_expected in dt_pairs:
        assert parse_formats(dt_str, fmt_strs) == dt_expected
```

houseofsuns:
Thanks for the quick answer and good to know that the problem is worked on. :)

I'll probably go the route you suggested and use strptime.

ewen-naos-nz:
One of my clients just got broken by this "regression" in behaviour between 2.4.2 (if year is detected first, will always be parsed as YMD):

                    if ymd[0] > 31 or \
                       (yearfirst and ymd[1] <= 12 and ymd[2] <= 31):
                        # 99-01-01
                        res.year, res.month, res.day = ymd

and 2.5.x/2.6.1 which apparently now gives two meanings to dayfirst=True:

1.  day is first, and year is last, ie, D/M/Y

[...]

Issue #289, opened 2016-08-19
A type conversion problem


zhang20054460:
In line 605 of dateutil/dateutil/parser.py , the type of 'res.tzname' may 'unicode', but the type of 'time.tzname' is 'str', so it will be error sometimes.

pganssle:
@zhang20054460 Can you give any details on this, like a minimum working example?

alexsilva:
Win 10 x64
Python 2.6 32bits
python-datetutil 2.6.0


https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py#L608

```
>>> dateutil.parser.parse('2017-01-09T13:13:15Z')
datetime.datetime(2017, 1, 9, 13, 13, 15, tzinfo=tzutc())
Python2.7\lib\site-packages\dateutil\parser.py:605: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  elif res.tzname and res.tzname in time.tzname:

```

The compared object has to be unicode (but it is not the case).
```
>>> import time
>>> time.tzname
('Hora oficial do Brasil', 'Hor\xe1rio brasileiro de ver\xe3o')
>>> type(time.tzname[0])
<type 'str'>
>>> type(time.tzname[1])
<type 'str'>
```

Fix
```
elif res.tzname and res.tzname.encode('utf8') in time.tzname:

```
However if `res.tzname` is not unicode the above code may fail.

```
elif res.tzname and (res.tzname.encode('utf8') if isinstance(res.tzname, unicode) else res.tzname) in time.tzname:
```

psa-jforestier:
This bug also occured on my Windows Seven box (in French) with python 2.7.13 : 
```
C:\Users\u402471\AppData\Roaming\Python\Python27\site-packages\dateutil\parser.py:605: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  elif res.tzname and res.tzname in time.tzname:
```
When pprinting the for debugging, I have : 
res.tzname = u'UTC'
time.tzname = ('Paris, Madrid', 'Paris, Madrid (heure d\x92\xe9t\xe9)')

pganssle:
@alexsilva @zhang20054460 @psa-jforestier Am I correct in my assessment that this *only* occurs on Python 2.7 on Windows and it is only a warning (i.e. it has no effect on the code)?

[...]

PR #415, opened 2017-07-12
Fix warning on French Windows systems when parsing date times under Python 2

AndreLouisCaron:
When parsing a date time in some conditions, ``dateutil.parser.parse()`` emits the following warning:

```
UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
```

This warning comes when testing if a unicode timezone name is in``time.tzname``.  I'm not sure exactly how to trigger this code path, but I can reproduce the warning on my system using the following snippet:

```python
>>> import time
>>> print time.tzname
('Est', 'Est (heure d\x92\xe9t\xe9)')
>>> print u'Est' in time.tzname
__main__:1: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
True
```

The problem is that ``time.tzname[1]`` contains a byte string with non-ASCII characters in the current code page.

To remove the warning, we should ensure to decode ``time.tzname`` using ``locale.getpreferredencoding()`` so that we have somthing that can safely compare with unicode strings.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -34,15 +34,24 @@ import string
 import time
 import collections
 import re
+import locale
 from io import StringIO
 from calendar import monthrange
 
-from six import text_type, binary_type, integer_types
+from six import text_type, binary_type, integer_types, PY2
 
 from . import relativedelta
 from . import tz
 
 __all__ = ["parse", "parserinfo"]
+
+
+TZNAME = time.tzname
+if PY2:
+    TZNAME = tuple(
+        name.decode(locale.getpreferredencoding())
+        for name in TZNAME
+    )
 
 
 class _timelex(object):
@@ -602,7 +611,7 @@ class parser(object):
                     raise ValueError("Offset must be tzinfo subclass, "
                                      "tz string, or int offset.")
                 ret = ret.replace(tzinfo=tzinfo)
-            elif res.tzname and res.tzname in time.tzname:
+            elif res.tzname and res.tzname in TZNAME:
                 ret = ret.replace(tzinfo=tz.tzlocal())
             elif res.tzoffset == 0:
                 ret = ret.replace(tzinfo=tz.tzutc())


AndreLouisCaron:
The warning was already showing up when I ran the test suite on my machine without the fix, so I didn't add any tests.

Let me know if you want me to add some.

pganssle:
@AndreLouisCaron This is reasonably elegant and addresses some of my issues in #289 (see [here](https://github.com/dateutil/dateutil/issues/402#issuecomment-311277352)).

The one thing I would change is that in the event that the time zone string is *not* in the preferred local encoding, I don't want the library to break on import just to avoid a warning.

A regression test would be appreciated but not blocking. If you want to write one, feel free to adapt it from my comment in #289. If you add a try/catch/pass block in there, might be good to mock out an example where trying to decode `TZNAME` raises an error to make sure that the error is caught, no worries either way, though.

pganssle:
Actually, one other concern here that *might* be blocking. I think `time.tzname` is dynamically updated in most circumstances if the time zone changes during the run. This will fix the time zone names at import time. This is less of a problem on Windows since, IIRC, the compiler used to build Python in Windows has a bug that prevents time zone information from updating during a run.

That said, I think this is literally only ever a problem on Windows, so maybe this can be refactored such that it pulls the data from `tzwin.tzwinlocal` or something like that on Windows instead of `time.tzname`. As it is now, I think it can't work.

AndreLouisCaron:
I totally understand the fear of inadvertently breaking subtle things, especially since time and locale implementations are so wildly different.  I don't mind taking my time to change this until you're comfortable :-)

AndreLouisCaron:
> in the event that the time zone string is not in the preferred local encoding

I'll gladly add a test for this, but, out of personal curiosity, when would this ever happen?

On Windows, the kernel is all Unicode and the byte string in the Python stdlib is a result of using legacy system calls.  Since the process can change its current code page, it's technically possible to run into a race condition where another thread changes the code page between the calls to `time.tzname` and `locale.getpreferredencoding()`, but changing the timezone other than at the very start of a program is just playing with fire.

AndreLouisCaron:
> I think `time.tzname` is dynamically updated in most circumstances if the time zone changes during the run.

I don't mind decoding on the fly, but that might add some overhead.  Is that OK?  If not, any suggestions for minimizing the overhead?

AndreLouisCaron:
> I think time.tzname is dynamically updated in most circumstances if the time zone changes during the run.

This made me curious.  Indeed, calling [time.tzset()](https://docs.python.org/2/library/time.html#time.tzset) resets `time.tzname` based on the `TZ` environment variable...

Since it's a tuple of immutable `str` objects, I guess we can cache the value of `time.tzname` and check if its identity has changed (e.g `if time.tzname is cached_value`) to determine if we can re-use the cache or if we should re-decode it.

pganssle:
> I'll gladly add a test for this, but, out of personal curiosity, when would this ever happen?

I doubt it would, but it's not like it's a guarantee of `time.tzname` that it is encoded in the preferred locale encoding, so we can't rely on that even though it's going to be the case in almost all instances. People probably do just mess with their locale settings until things work and then stop, though, so I wouldn't be surprised if there were a few systems out there like this due to crazy configurations.

> I don't mind decoding on the fly, but that might add some overhead. Is that OK? If not, any suggestions for minimizing the overhead?

I don't really care that much about the overhead in Python 2.7. If you're comfortable using a 10-year-old version of Python, you should probably also be comfortable using a 1-year-old version of dateutil if you're that performance sensitive. The parser itself is also probably not a particularly performant beast, so one extra function call probably won't make a difference.

Something like:

```python
def _get_tznames():
    return time.tzname

if PY2:
   try:
        from tz.tzwin import _get_tznames
    except ImportError:
        pass
```

For the moment, it doesn't really matter what `_get_tznames` returns, since the parser has no ability to recognize Windows-style time zone abbreviations (it is expecting 3- and 4-letter abbreviations). An empty tuple is fine with a comment explaining why, or if you want to be proper there are functions in `tz.tzwin` that will get you at least one of the zone names in a way that respects `tzset()` I believe. I can look into it more carefully this weekend maybe.

PR #416, opened 2017-07-12
Fix bug on French Windows systems when using local timezone under Python 2

AndreLouisCaron:
When I try to run the following snippet, I get the following output:

```python
>>> from dateutil.tz import tzlocal
>>>
>>> now = datetime.now()
>>> print tzlocal().tzname(now)
  File "...\dateutil\tz\_common.py", line 22, in adjust_encoding
    name = name.encode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0x92 in position 12: ordinal not in range(128)
```

This happens because ``time.tzname[1]`` contains non-ASCII characters in the current code page.  We need to decode this using ``locale.getpreferredencoding()`` to get a Unicode string
that is safe to use.

Additionally, since Python 2 forces ``.tzname()`` to return a byte string, the ``adjust_encoding()`` utility tries to re-encode to ASCII.  This is obviously not possible so the best path seems to be to remove diacritics to lossily convert the stirng to ASCII.

### DIFF
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -13,8 +13,9 @@ import time
 import sys
 import os
 import bisect
+import locale
 
-from six import string_types
+from six import string_types, PY2
 from ._common import tzname_in_python2, _tzinfo, _total_seconds
 from ._common import tzrangebase, enfold
 from ._common import _validate_fromutc_inputs
@@ -27,6 +28,19 @@ except ImportError:
 ZERO = datetime.timedelta(0)
 EPOCH = datetime.datetime.utcfromtimestamp(0)
 EPOCHORDINAL = EPOCH.toordinal()
+
+
+def _gettznames():
+    """Get current time zone names in Unicode."""
+
+    def decode(x):
+        """Decode from byte string in current locale."""
+        return x.decode(locale.getpreferredencoding())
+
+    names = time.tzname
+    if PY2:
+        names = tuple(map(decode, names))
+    return names
 
 
 class tzutc(datetime.tzinfo):
@@ -191,7 +205,7 @@ class tzlocal(_tzinfo):
 
     @tzname_in_python2
     def tzname(self, dt):
-        return time.tzname[self._isdst(dt)]
+        return _gettznames()[self._isdst(dt)]
 
     def is_ambiguous(self, dt):
         """
@@ -1404,7 +1418,7 @@ def gettz(name=None):
                     else:
                         if name in ("GMT", "UTC"):
                             tz = tzutc()
-                        elif name in time.tzname:
+                        elif name in _gettznames():
                             tz = tzlocal()
     return tz
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,9 @@ before_install:
 
 install:
   - pip install six
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install unittest2; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install Unidecode unittest2; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install Unidecode; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 'pypy-5.4' ]]; then pip install Unidecode; fi
   - ./ci_tools/retry.sh python updatezinfo.py
 
 script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,6 +23,7 @@ install:
   - ps: Start-FileDownload 'https://bootstrap.pypa.io/get-pip.py'
   - "%PYTHON%/python.exe get-pip.py"
   - "%pip_cmd% install six"
+  - "%pip_cmd% install Unidecode"
   - "%pip_cmd% install coverage"
   - "%pip_cmd% install codecov"
 
--- a/dateutil/tz/_common.py
+++ b/dateutil/tz/_common.py
@@ -1,8 +1,14 @@
-from six import PY3
+from six import PY2
 
 from functools import wraps
 
 from datetime import datetime, timedelta, tzinfo
+
+if PY2:
+    from unidecode import unidecode
+
+import locale
+from six import text_type
 
 
 ZERO = timedelta(0)
@@ -18,8 +24,8 @@ def tzname_in_python2(namefunc):
     """
     def adjust_encoding(*args, **kwargs):
         name = namefunc(*args, **kwargs)
-        if name is not None and not PY3:
-            name = name.encode()
+        if PY2 and isinstance(name, text_type):
+            name = unidecode(name)
 
         return name
 
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,10 @@ datetime module available in the Python standard library.
       package_data={"dateutil.zoneinfo": ["dateutil-zoneinfo.tar.gz"]},
       zip_safe=True,
       requires=["six"],
-      install_requires=["six >=1.5"],  # XXX fix when packaging is sane again
+      install_requires=[
+          "six >=1.5",
+          "Unidecode>=0.04.21",
+      ],  # XXX fix when packaging is sane again
       classifiers=[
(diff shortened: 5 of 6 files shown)


AndreLouisCaron:
The existing test suite was already breaking on my machine without the fix, so I didn't add any tests.

Let me know if you want me to add some.

dateutil/tz/tz.py
```diff
@@ -191,7 +205,7 @@ def dst(self, dt):
 
     @tzname_in_python2
     def tzname(self, dt):
-        return time.tzname[self._isdst(dt)]
+        return _gettznames()[self._isdst(dt)]
```
AndreLouisCaron: **Note to reviewers**: there is a performance overhead to decoding the timezone names on the fly.  My initial attempt cached the decoded values to perform it only once, but this broke some tests which mock this.  I preferred leaving the test suite unchanged, but if you're worried about performance, suggestions for improvements are welcome :-)

AndreLouisCaron:
Not sure what's up with `pypy3`, but I'm not in a position to test that right now.  If anybody has an suggestions, please share :-)

pganssle:
I'm really not comfortable adding in new dependencies, particularly in Python 3, but it seems that there is likely a way to implement this ourselves reasonably without the dependency.

My major conceptual concern here is that now we're proactively decoding things to `unicode` that may well be expecting bytes, which may break a lot of stuff. The bytestring is only a problem if you try to decode it, which you know you need to do anyway if you're trying to print strings in Python 2 anyway.

I don't think it's the worst idea in the world, but I'm not convinced that it's adding enough value for the amount of stuff that will likely break if we change the return type of `tzname` (not to mention I think that breaks the `tzinfo` API). I'll have to do a deeper dive and mull this over, but I think the right answer might be "decode your bytestrings properly if you're using Python 2". This is huge part of the reason Python 2 is a total pain.

AndreLouisCaron:
> I'm really not comfortable adding in new dependencies, particularly in Python 3

No worries, I was expecting this to come up (but crossing my fingers) :-)

> My major conceptual concern here is that now we're proactively decoding things to unicode that may well be expecting bytes

I agree.  I'm actually not sure why the value is decoded then re-encoded to ASCII.  How does that fix anything?

Can we remove the decoding entirely?

> I think the right answer might be "decode your bytestrings properly if you're using Python 2".

That's what I was aiming for.  Make sure we use the "real" value for `time.tzname`.  Maybe the logic I put only applies to Windows, in which case, I'll gladly put this in an "if windows" block.

I'll let you think about this.  Let me know what you think and I'll update accordingly.

AndreLouisCaron:
BTW, I initially hit this problem here: Yelp/elastalert#1141.  My current workaround is to force ElastAlert to use UTC, which bypasses this problem.  I'm also helping them port to Python 3.5, which will also offer another alternative.

pganssle:
By the way, no need to worry about `pypy3`. That's set to optional because the `fold` support is broken in some way and I wasn't able to get it working on my local machine to debug it.

pganssle:
> I agree. I'm actually not sure why the value is decoded then re-encoded to ASCII. How does that fix anything?

I think the `tzinfo` API intends that you return `str` in both Python 2 and Python 3. `tzlocal` is a thin wrapper around `time`, so this is really a "problem" with `time.tzname`. I think things are more likely to break in unexpected ways if we start returning unicode strings than if we expect people to handle encoded-bytestrings properly.

Note that this is only a problem on Windows and `time` is broken on Windows. Do you see the same issue if you use `tzwinlocal` instead?

AndreLouisCaron:
> Note that this is only a problem on Windows and time is broken on Windows. Do you see the same issue if you use tzwinlocal instead?

I just tried this example:

```python
import logging

from datetime import datetime
from dateutil.tz import tzlocal
from dateutil.tzwin import tzwinlocal

try:
    now = datetime.now().replace(tzinfo=tzlocal())
    print now.strftime('%Y-%m-%d %H:%M %Z')
except Exception:
    logging.exception('tzlocal')

try:
    now = datetime.now().replace(tzinfo=tzwinlocal())
    print now.strftime('%Y-%m-%d %H:%M %Z')
except Exception:
    logging.exception('tzwinlocal')
```

Both of these fail with the same output:

```
$ python py2-tzname-bug.py
ERROR:root:tzlocal
  File "py2-tzname-bug.py", line 9, in <module>
    print now.strftime('%Y-%m-%d %H:%M %Z')
  File "...\dateutil\tz\_common.py", line 22, in adjust_encoding
    name = name.encode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0x92 in position 12: ordinal not in range(128)
ERROR:root:tzwinlocal
  File "py2-tzname-bug.py", line 15, in <module>
    print now.strftime('%Y-%m-%d %H:%M %Z')
  File "...\dateutil\tz\_common.py", line 22, in adjust_encoding
    name = name.encode()
UnicodeEncodeError: 'ascii' codec can't encode characters in position 12-13: ordinal not in range(128)
```

AndreLouisCaron:
Also, it seems like there are other issues with `tzwin`...  Unless I'm doing something wrong?

```python
>>> from dateutil.tzwin import tzwinlocal, tzwin, tzres
>>> tzwin(tzres().load_name(111))
  File "...\dateutil\tz\win.py", line 198, in __init__
    with winreg.OpenKey(handle, tzkeyname) as tzkey:
UnicodeEncodeError: 'ascii' codec can't encode characters in position 68-69: ordinal not in range(128)
```

pganssle:
Might well be the same issue. Presumably these things return `str` which require some manipulation in Python 2. I think just try to do the right thing and in case of errors fall back to the original behavior. This only fails if you try to output / print the answer, right?

pganssle:
Oh, sorry, I was thinking of the other, very similar PR on parsing. In this case I'm pretty sure I do *not* want `tzname()` returning `unicode` in Python 2. I'm 90% sure that would break a bunch of stuff downstream.

AndreLouisCaron:
>  This only fails if you try to output / print the answer, right?

No, `datetime.strftime('%Z')` fails because it uses `tzinfo.tzname()`.  This happens even if I don't try to print anything.

AndreLouisCaron:
> In this case I'm pretty sure I do not want tzname() returning unicode in Python 2. I'm 90% sure that would break a bunch of stuff downstream.

I agree.

I did a bit more digging to try to see a bit more clear.  It seems like `tzlocal` and `tzwinlocal` have different problems even if the symptoms are the same.

`tzlocal` fetches its info from `time.tzname` which is a byte string containing non-ASCII data in CP1252 (because it uses a legacy ANSI API).  The problem comes from our attempt to `.encode('ascii')` this value.  

We have a few options for `tzlocal`:
1) pass it on as is and let the caller deal with it 
2) try to coerce it to ASCII (e.g. using `Unidecode` or other lossy algorithm).

`tzwinlocal` fetches its info from the `SYSTEM\CurrentControlSet\Control\TimeZoneInformation` registry key.  It (correctly) reads the `StandardName` and `DaylightName` as Unicode strings.  The problem is it has non-ASCII data.

We have a more options options for obtaining a byte string from `tzwininfo`:
1) pass it on as is (creates situations where `tzinfo.tzname()` returns Unicode -- inconsistent at best);
2) switch to ANSI versions of registry API (possibly lossy, but consistent with `time.tzname`);
3) encode as UTF-8 (lossless, but inconsistent with `time.tzname`).

I can't think of anything else at the moment.  Let me know what you prefer.

AndreLouisCaron:
I started playing around with an alternative approach.  The idea is basically:

1. Change `adjust_encoding()` to `.encode()` only Unicode objects.  This allows `tzname()` implementations to perform their own conversions to byte string when they know the encoding.
2. Let `tzlocal.tzname()` return the raw byte string found in `time.tzinfo` and let the caller deal with the result as they would if the used `time.tzname` directly.
3. Change `tzwinlocal` to convert the string to UTF-8 (or `locale.getpreferredencoding()` at your discretion).

I don't see any regression here since the code currently fails with an exception in cases where this would change something.  Of course, it's possible that this will reveal latent Unicode handling issues in applications,  but failing with a `UnicodeDecodeError` in `strftime()` or in their own code is likely to only change the stack trace in those cases.

Thoughts?

Issue #417, opened 2017-07-14
dateutil.parse.parse() rejects bytearray as input


uckelman:
```
>>> import dateutil.parser
>>> dateutil.parser.parse(b'17 Jan 2017')
datetime.datetime(2017, 1, 17, 0, 0)
>>> dateutil.parser.parse(bytearray(b'17 Jan 2017'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 1168, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 556, in parse
    res, skipped_tokens = self._parse(timestr, **kwargs)
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 675, in _parse
    l = _timelex.split(timestr)         # Splits the timestr into tokens
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 192, in split
    return list(cls(s))
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 61, in __init__
    '{itype}'.format(itype=instream.__class__.__name__))
TypeError: Parser must be a string or character stream, not bytearray
```

Any function which can operate on a `bytes` ought to be able to operate on a `bytearray`---unless something has specifically been done to prevent that, as in `dateutil.parser.parse()`. When `bytes` works, it's unexpected that `bytearray` would fail. It would be better to accept `bytearray`s as well.

uckelman:
A one line change to `datetime.parser._timelex.__init__()` would fix this:
```
if isinstance(instream, (binary_type, bytearray)):
```
instead of
```
if isinstance(instream, binary_type):
```
However, letting duck typing work would be even better:
```
        if not hasattr(instream, 'read'):
            if hasattr(instream, 'decode'):
                instream = instream.decode()
            
            if isinstance(instream, text_type):
                instream = StringIO(instream)

            if not hasattr(instream, 'read'):
                raise TypeError('Parser must be a string or character stream, not '
                                '{itype}'.format(itype=instream.__class__.__name__))
```

pganssle:
I wouldn't say that support for `bytesarray` is particularly important here, but I am generally in favor of duck typing. I think that when we did the mypy typing for `typeshed`, we decided that the supported inputs would be `bytes`, `str` and "anything with a `.read()`".

[...]

PR #418, opened 2017-07-17
Make parse() accept anything which has a nullary decode() method

uckelman:
This change fixes Issue #417, by making `parse()` use duck typing rather than explicitly checking whether it's been given a `bytes`. Now any object which has a nullary `decode()` method can be passed to `parse()`.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -59,8 +59,10 @@ class _timelex(object):
 
             if not hasattr(instream, 'read'):
                 raise TypeError(
-                    'Parser must be a string or character stream, not '
-                    '{itype}'.format(itype=instream.__class__.__name__)
+                    'Parser must be a character stream, a string, or '
+                    'decodable to a string, not {itype}'.format(
+                        itype=instream.__class__.__name__
+                    )
                 )
 
         self.instream = instream
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -37,7 +37,7 @@ import re
 from io import StringIO
 from calendar import monthrange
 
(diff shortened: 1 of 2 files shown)


dateutil/parser.py
```diff
@@ -50,15 +50,20 @@ class _timelex(object):
...
-        if isinstance(instream, binary_type):
-            instream = instream.decode()
-
-        if isinstance(instream, text_type):
-            instream = StringIO(instream)
-
-        if getattr(instream, 'read', None) is None:
-            raise TypeError('Parser must be a string or character stream, not '
-                            '{itype}'.format(itype=instream.__class__.__name__))
+        if not hasattr(instream, 'read'):
```
pganssle: I do prefer the `getattr(instream, 'read', None) is None` idiom over `hasattr`, at least until Python 2.7 support is dropped, based on [this article](https://hynek.me/articles/hasattr/).

Also, why is this being checked at all? Just to avoid streams that have a `.decode()` method?

pganssle:
I'm not really sure how to properly ducktype `bytes`-like objects in a Python 2.7/3.3-compatible way. I can't actually even find an obvious way to duck-type `bytes`-like objects in Python 2.7, since both `str` and `unicode` have `encode()` and `decode()` methods.

I suppose it's not the worst thing in the world to just explicitly add support for `bytearray`.

pganssle:
I'll pull this and try to rebase it against master. I think maybe the right approach here is to use proper duck-typing in Python 3, and use `isinstance` in Python 2.

pganssle:
Fixed in #514. Thank you for your contribution.

Issue #414, opened 2017-07-10
DISC: state storage in parsing


jbrockmendel:
With #409 merged, the usage of `mstridx` is now uniform.  There are three places with:

```
ymd.append(value)
assert mstridx == -1
mstridx = len(ymd)-1
```

I propose that `mstridx` should be an attribute of `_ymd`.  My preferred implementation is a new `_ymd` method that gets called in some cases instead of `append`:

```
   def set_month(self, value, idx):
           assert self.mstridx is None # or -1 in in the current usage
           self.append(value)
           self.mstridx = len(self) - 1
           self._midx = idx  # This is new, will be discussed below
```

Then in `_parse`, the append + assert + set-mstridx becomes `ymd.set_month(value, i)`.  Advantages:

- One fewer state variable to carry around the `_parse` loop. This will make it simpler to refactor that into smaller functions.
- `_ymd.resolve_ymd` no longer needs `mstridx` as an argument.
- Analogous methods for `set_year` and `set_day` allow for similar validation (e.g. calling `ymd.append(info.convertyear(...))` to implicitly specify `ystridx`)

The new attribute `_midx`, and the analogous `_yidx, _didx` are there so that we can recover information later.  Keeping track of absolute position in addition to relative position may be important in cases where we need to back-track for disambiguation (see #394).  This may also be helpful for addressing #125.

Related: instead of carrying around `skipped_tokens` and `last_skipped_token_i`, use a set `skipped_indices`.  The re-combining can be done at the end of the parsing, `last_skipped_token_i` is easy to get, and we retain information that can be useful in potential post-processing.

pganssle:
@jbrockmendel I haven't had time to review this, but it sounds good to me. IIRC, you're interested in helping to break up the parser, right? I'd like to get started on that soon-ish (next few weeks maybe, depending on other workload). If you think it would be easier to have a real-time conversation (voice call or something), shoot me an e-mail and we can arrange a time.

PR #419, opened 2017-07-17
Less of a PR than an exposition/discussion

jbrockmendel:
I don't really expect/intend this to be merged, just want to show the places that I think make for easy refactoring wins.  These will be easiest to view commit-by-commit.

The heuristic for refactoring candidates was
- Can this be made a function without passing `res` or `ymd`?
- Is it obvious what this function would be named?"

[_build_tzinfo](https://github.com/jbrockmendel/dateutil/commit/ad1bff2157c5b4f12e9e3d0dfcdf0d08bb350f2a) and [_ampm_validity](https://github.com/jbrockmendel/dateutil/commit/9c107b9325a9438ae8cff4c8ca268ca915a9f287) just cut/paste a chunk of code from a method into a function.

[_adjust_ampm](https://github.com/jbrockmendel/dateutil/commit/ee75ff0a50e9fd73fc05941799f5af5464517558) and [_parse_min_sec](https://github.com/jbrockmendel/dateutil/commit/54deb3325d3aa28bf2822dad380ceb7b02015c07) are each only ~4 lines, but appear a several times each.

Commits 5 and 8 are just small cleanups.

The first commit [deprecates](https://github.com/jbrockmendel/dateutil/commit/3b22f0d9a4eef23ceabdc466691caa948ee259cc) `_skip_token`.  Instead of carrying around `skipped_tokens` and `last_skipped_index_i`, this is replaced with a set `skipped_idxs`, and at the end of the parsing `_recombine_skipped` is called to create `skipped_tokens`.

The second commit makes `mstridx` an attribute of `_ymd` and moves validation into `ymd.append`.  This was discussed briefly in #414.

Those were all of the candidates I saw that didn't alter any of the flow, i.e. should be fairly uncontroversial.  The next goals I'll suggest are

- Reducing the number of places where `i` is incremented (currently 51).
- Moving away from the `while True` block.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -1274,13 +1274,6 @@ class InvalidDatetimeError(ValueError): pass
 class InvalidDateError(InvalidDatetimeError): pass
 class InvalidTimeError(InvalidDatetimeError): pass
 
-class ProgrammingError(AssertionError):
-    """ProgrammingError is for if we reach a state that we *think* should
-    be unreachable and would otherwise be inclined to assert is impossible.
-    Since an AssertionError is useful a user, a ProgrammingError clarifies
-    that they should file a bug report.
-    """
-
 
 def _parse_hms(i, l, info, res):
     # TODO: This is still a mess.  Can we make this recursive instead of
@@ -1335,7 +1328,7 @@ def _parse_hms(i, l, info, res):
     return i
 
 
-# TODO: requre len(token) >= 3 like we do for the between-parens version?
+# TODO: require len(token) >= 3 like we do for the between-parens version?
 # do some other validation here instead of putting it off?  As of now, "Q"
 # will be accepted as a timezone...
 def _could_be_tzname(hour, tzname, tzoffset, token):
@@ -1416,7 +1409,7 @@ def _recombine_skipped(tokens, skipped_idxs):
 
     """
     skipped_tokens = []
-    for idx in sorted(list(skipped_idxs)):
+    for idx in sorted(skipped_idxs):
         if idx-1 in skipped_idxs:
             skipped_tokens[-1] = skipped_tokens[-1] + tokens[idx]
         else:
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -417,8 +417,12 @@ class _ymd(list):
         if hasattr(val, '__len__'):
             if val.isdigit() and len(val) > 2:
                 self.century_specified = True
+                assert label in [None, 'Y']
+                label = 'Y'
         elif val > 100:
             self.century_specified = True
+            assert label in [None, 'Y']
+            label = 'Y'
 
         super(self.__class__, self).append(int(val))
 
@@ -728,7 +732,7 @@ class parser(object):
                         # YYMMDD or HHMMSS[.ss]
                         s = l[i]
 
-                        if not ymd and l[i].find('.') == -1:
+                        if not ymd and '.' not in l[i]:
                             #ymd.append(info.convertyear(int(s[:2])))
 
                             ymd.append(s[:2])
@@ -744,7 +748,7 @@ class parser(object):
                     elif len_li in (8, 12, 14):
                         # YYYYMMDD
                         s = l[i]
-                        ymd.append(s[:4])
+                        ymd.append(s[:4], 'Y')
                         ymd.append(s[4:6])
                         ymd.append(s[6:8])
 
@@ -916,7 +920,7 @@ class parser(object):
                                 # TODO: not hit in tests
                             else:
                                 # Convert it here to become unambiguous
-                                ymd.append(str(info.convertyear(value)))
+                                ymd.append(str(info.convertyear(value)), 'Y')
                             i += 5
 
                         else:
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -907,15 +907,15 @@ class parser(object):
                               and info.pertain(l[i+2])):
                             # Jan of 01
                             # In this case, 01 is clearly year
-                            try:
+                            if l[i+4].isdigit():
+                                # Convert it here to become unambiguous
                                 value = int(l[i+4])
-                            except ValueError:
+                                year = str(info.convertyear(value))
+                                ymd.append(year, 'Y')
+                            else:
                                 # Wrong guess
                                 pass
                                 # TODO: not hit in tests
-                            else:
-                                # Convert it here to become unambiguous
-                                ymd.append(str(info.convertyear(value)), 'Y')
                             i += 4
 
 
@@ -957,6 +957,7 @@ class parser(object):
                     signal = (-1, 1)[l[i] == '+']
                     len_li = len(l[i+1])
 
+                    # TODO: check that l[i+1] is integer?
                     if len_li == 4:
                         # -0300
                         hour_offset = int(l[i+1][:2])
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -881,15 +881,15 @@ class parser(object):
                     continue
 
                 # Check weekday
-                value = info.weekday(l[i])
-                if value is not None:
+                elif info.weekday(l[i]) is not None:
+                    value = info.weekday(l[i])
                     res.weekday = value
                     i += 1
                     continue
 
                 # Check month name
-                value = info.month(l[i])
-                if value is not None:
+                elif info.month(l[i]) is not None:
+                    value = info.month(l[i])
                     ymd.append(value, 'M')
 
                     i += 1
@@ -923,8 +923,8 @@ class parser(object):
                     continue
 
                 # Check am/pm
-                value = info.ampm(l[i])
-                if value is not None:
+                elif info.ampm(l[i]) is not None:
+                    value = info.ampm(l[i])
                     val_is_ampm = _ampm_validity(res.hour, res.ampm, fuzzy)
 
                     if val_is_ampm:
@@ -938,7 +938,7 @@ class parser(object):
                     continue
 
                 # Check for a timezone name
-                if (res.hour is not None and len(l[i]) <= 5
+                elif (res.hour is not None and len(l[i]) <= 5
                     and res.tzname is None and res.tzoffset is None
                     and all(x in string.ascii_uppercase for x in l[i])):
                     res.tzname = l[i]
@@ -961,7 +961,7 @@ class parser(object):
                     continue
 
                 # Check for a numbered timezone
-                if res.hour is not None and l[i] in ('+', '-'):
+                elif res.hour is not None and l[i] in ('+', '-'):
                     signal = (-1, 1)[l[i] == '+']
                     len_li = len(l[i+1])
 
@@ -992,11 +992,12 @@ class parser(object):
                     continue
 
                 # Check jumps
-                if not (info.jump(l[i]) or fuzzy):
+                elif not (info.jump(l[i]) or fuzzy):
                     return None, None
 
-                skipped_idxs.add(i)
-                i += 1
+                else:
+                    skipped_idxs.add(i)
+                    i += 1
 
             # Process year/month/day
             year, month, day = ymd.resolve_ymd(yearfirst, dayfirst)
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -764,7 +764,6 @@ class parser(object):
                             i += 1
 
                         idx = info.hms(l[i+1])
-                        i += 1
 
                         while True:
                             if idx == 0:
@@ -780,12 +779,12 @@ class parser(object):
 
                             i += 1
 
-                            if i >= len_l or idx == 2:
+                            if i+1 >= len_l or idx == 2:
                                 break
 
                             # 12h00
                             try:
-                                value_repr = l[i]
+                                value_repr = l[i+1]
                                 value = float(value_repr)
                             except ValueError:
                                 break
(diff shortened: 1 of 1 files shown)


jbrockmendel:
Just pushed a bunch more commits.  Again, these are best viewed commit-by-commit.  The ones this morning were intended to be obviously-correct, easy to review, and avoid making any serious design decisions.  In contrast, these meander and in some cases border on silly.

A few highlights:
- Implement `InvalidDatetimeError(ValueError)` and a couple others.
- Decrease the set of paths we can take through The Loop
    - Raise appropriate `ValueError`s instead of returning `(None, None)`
    - Remove `continue` statements and instead change a bunch of `if`s to `elif`s.  The cost of this is a few extra dictionary lookups.
    - Delay `i += 1` wherever possible.  Fewer state changes make reasoning easier (for me at least), and reduce the number of args/returns needed as pieces of the loop are refactored out into their own functions.
- Refactor out the `while True` loop into `_hms_block`.  In this case I still don't like the function it became (and don't like passing `res` or returning `i`), but separating it gives hope that it may be de-whiled.
- The silly one: override `l.__getitem__` to return an empty string for out-of-bounds access.  That way we can drop the `len_l` checks from a bunch of things like `if i+3 < len_l and info.month(l[i+3])`

pganssle:
By the way, with regards to reducing the `i < len_l` checks to avoid `IndexError`, there's a possibly useful fact that in certain circumstances can be used to avoid index checks, which is that indexes are not ignored in slices:

```python
>>> a = 'Hello'
>>> a[0:3]
'Hel'
>>> a[5:6]
''
```

So if you want to compare a specific index of a string or something and don't want to be bothered to check if it exists:

```python
>>> a[4:5] == 'o'
True
>>> a[5:6] == 'o'
False
```

The downsides to using this are that it may be harder to read and I haven't checked what impact it would have on performance. In our case, you can imagine `if i + 1 < len_l and l[i] == ':'` might be refactored into `if l[i:i+1] == [':']`, but I think that the second statement is definitely slower than the first (it involves creating a list literal instead of a string literal, for one thing). Presumably we can check the performance of a proposed `l.__getitem__` solution against this.

jbrockmendel:
I've actually decided to roll back the `__getitem__` patch locally.  It's kind of a cute space-saver, but the amount of non-standardness it causes isn't worth the tradeoff.  After I push that change, this will be more or less a real PR.

pganssle:
@jbrockmendel Your last message was not 100% clear, is this ready for review?

jbrockmendel:
Yes.

pganssle:
I did not realize initially that these were all being moved to global scope. It's not a big deal, but all the components of the parser should stay as member functions of the parser. The final goal will be the make them public so that downstream users can tweak the fine-grained behavior of the parser by subclassing (like virtual functions in C++).

Don't worry about making the change, though. I think it might be a pain to keep this and #425 in sync. I'm still working through this, but barring anything actually conceptually wrong, we can merge this and then #425, then I'll go through and move these functions to member functions.

dateutil/parser.py
```diff
@@ -742,318 +763,215 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
...
+                        min_offset = 0
                     else:
-                        return None, None
-                    i += 1
+                        raise InvalidDatetimeError(timestr)
 
-                    res.tzoffset *= signal
+                    res.tzoffset = signal * (hour_offset*3600 + min_offset*60)
 
+                    # TODO: Are we not requiring that res.tzname be None?
```
pganssle: I think that there's a question of adjudication if you find two things that have equal claim to being a time zone in the string. Right now I guess it just takes the most recent thing that it finds that matches?

dateutil/parser.py
```diff
@@ -742,318 +763,215 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
...
-                        3 <= len(l[i+2]) <= 5 and
-                        not [x for x in l[i+2]
-                             if x not in string.ascii_uppercase]):
+                    if (i+5 < len_l and
+                        info.jump(l[i+2]) and l[i+3] == '(' and l[i+5] == ')' and
+                        3 <= len(l[i+4]) <= 5 and
+                        all(x in string.ascii_uppercase for x in l[i+4])): # TODO: merge this with _could_be_tzname
                         # -0300 (BRST)
-                        res.tzname = l[i+2]
+                        res.tzname = l[i+4]
```
pganssle: Are these differences accumulating because you're reducing the number of passes through the loop?

dateutil/parser.py
```diff
@@ -1361,6 +1270,132 @@ def parse(self, tzstr):
 def _parsetz(tzstr):
     return DEFAULTTZPARSER.parse(tzstr)
 
+class InvalidDatetimeError(ValueError): pass
+class InvalidDateError(InvalidDatetimeError): pass
+class InvalidTimeError(InvalidDatetimeError): pass
+
+class ProgrammingError(AssertionError):
```
pganssle: This seems unused, and I'm somewhat dubious of its utility.

dateutil/parser.py
```diff
@@ -1361,6 +1270,132 @@ def parse(self, tzstr):
...
+                    if newidx is not None:
+                        idx = newidx
+                i += 2
+    return i
+
+
+# TODO: requre len(token) >= 3 like we do for the between-parens version?
+# do some other validation here instead of putting it off?  As of now, "Q"
+# will be accepted as a timezone...
+def _could_be_tzname(hour, tzname, tzoffset, token):
```
pganssle: require is misspelled here.

Also, we absolutely cannot require `len(token) >= 3`, since `Z` is in the ISO8601 spec for Zulu time (UTC), unless we want to special-case this.

That said, when we fix this up, we may well be able to determine exactly the minimum and maximum length of all valid time zones. We can start by checking the length of all the keys of `tzinfos`.

dateutil/parser.py
```diff
@@ -1361,6 +1270,132 @@ def parse(self, tzstr):
...
+    # If there's already an AM/PM flag, this one isn't one.
+    if fuzzy and ampm is not None:
+        val_is_ampm = False
+
+    # If AM/PM is found and hour is not, raise a ValueError
+    if hour is None:
+        if fuzzy:
+            val_is_ampm = False
+        else:
+            raise ValueError('No hour specified with AM or PM flag.')
```
pganssle: Are these caught and re-raised in another form? Otherwise, they should be some `ValueError` subclass.

dateutil/parser.py
```diff
@@ -1370,5 +1357,21 @@ def _parsems(value):
...
+    """
+
+    >>> tokens = ["foo", " ", "bar", " ", "19June2000", "baz"]
+    >>> skipped_idxs = set([0, 1, 2, 5])
+    >>> _recombine_skipped(tokens, skipped_idxs)
+    ["foo bar", "baz"]
+
+    """
+    skipped_tokens = []
+    for idx in sorted(list(skipped_idxs)):
```
jbrockmendel: Good catch.

dateutil/parser.py
```diff
@@ -673,21 +697,16 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
 
         res = self._result()
         l = _timelex.split(timestr)         # Splits the timestr into tokens
+        tokens = l # alias that is easier to grep
 
-        # keep up with the last token skipped so we can recombine
-        # consecutively skipped tokens (-2 for when i begins at 0).
-        last_skipped_token_i = -2
-        skipped_tokens = []
+        skipped_idxs = set()
```
jbrockmendel: IIRC the main reason I used a set here was because of the `if idx-1 in skipped_idxs` check.

The secondary concern was idempotency.  I think that eventually we're going to have to do multiple passes or look-backs or something.

The version you posted doesn't work for me.   I'm assuming the `for idx in skipped_tokens:` is supposed to be `for idx in skipped_idxs:`.  If I make that change, then in py3 I get `['foo bar']` and in py2 I get an `AttributeError` because there is no `list.clear`.

dateutil/parser.py
```diff
@@ -742,318 +763,215 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
...
-                    else:
-                        i += 1
-                    continue
+                        raise InvalidDatetimeError(timestr)
+
 
                 # Check weekday
-                value = info.weekday(l[i])
-                if value is not None:
+                elif info.weekday(l[i]) is not None:
```
jbrockmendel: > Was the huge if statement being pointlessly broken up just so that info.weekday(l[i]) didn't need to be called twice?

AFAICT, yes.  I was pretty sure that the big `if` blocks were mutually exclusive, but the number of `continue` statements, `return (None, None)`, and exceptions made for enough code paths that I wanted to `(n+1)`-check.  I chose to take the performance hit of a few extra dict lookups in order to make this obvious and have only one non-raising path through the loop.

dateutil/parser.py
jbrockmendel: Agreed, removing it.

dateutil/parser.py
jbrockmendel: These are because I am delaying incrementing `i` for as long as possible.  This decreases the total number of places where `i` is incremented at all (41 down from 51).  The closer `i` is to fixed within each iteration of the loop, the easier I find this to reason about.

dateutil/parser.py
jbrockmendel: At the moment this still gets caught by the `except (IndexError, ValueError, AssertionError)` wrapping the big loop.  When we eventually get rid of that catch-all, I agree that something like `InvalidTimeError` (which subclasses `ValueError`) would be more appropriate.

jbrockmendel:
I just pushed fixes to the small things: "require" typo, get rid of `ProgrammingError`, `sorted(list` --> `sorted(`.  Other than that I want to avoid tinkering for the time being.

dateutil/parser.py
pganssle: Ah, the version I had was indeed a bit off and not Python 2.7 compatible. Here's an updated version:

```python
def _recombine_skipped_queue(tokens, skipped_idxs):
    """
    >>> tokens = ["foo", " ", "bar", " ", "19June2000", "baz"]
    >>> skipped_idxs = set([0, 1, 2, 5])
    >>> _recombine_skipped(tokens, skipped_idxs)
    ["foo bar", "baz"]

    """
    # This groups consecutive values
    skipped_tokens = []
    idx_queue = []
    for idx in skipped_idxs:
        if idx_queue and idx - 1 == [-1]:
        if len(idx_queue) and idx - 1 != idx_queue[-1]:
            skipped_tokens.append(''.join(map(tokens.__getitem__, idx_queue)))
            idx_queue = []

        idx_queue.append(idx)

    if idx_queue:
        skipped_tokens.append(''.join(map(tokens.__getitem__, idx_queue)))

    return skipped_tokens
```

Interestingly, I couldn't figure out why this was consistently slower than your version, since strings are immutable, I was expecting growing a string incrementally with `+` to be *much* slower than queuing up all the parts and combining them at the end, but lo and behold, if I switch over to using the `+=` style method, the list version is faster than your version! I believe that's because in CPython they've special-cased this sort of string extension to try to extend the string in-place. See [this StackOverflow question](https://stackoverflow.com/questions/4435169/good-way-to-append-to-a-string). Here's the faster version:

```python
def _recombine_skipped(tokens, skipped_idxs):
    """
    >>> tokens = ["foo", " ", "bar", " ", "19June2000", "baz"]
    >>> skipped_idxs = set([0, 1, 2, 5])
    >>> _recombine_skipped(tokens, skipped_idxs)
    ["foo bar", "baz"]

    """
    # This groups consecutive values
    skipped_tokens = []
    idx_queue = []
    for i, idx in enumerate(skipped_idxs):
        if i > 0 and idx - 1 == skipped_idxs[i - 1]:
            skipped_tokens[-1] = skipped_tokens[-1] + tokens[idx]
        else:
            skipped_tokens.append(tokens[idx])

    return skipped_tokens
```

Despite the fact that this is not part of the Python spec, it seems that it's been implemented in Python 2.7 and 3.6 as well as pypy2 and pypy3. Using a loop that randomly generates token strings and skipped indices to test this:

Python 2.7:
```
$ python2.7 test_recombination_speed.py 
Running extend timing test
1000 loops, 500 sets: 0.884 ms per loop, 1.767 us per set
Running queue timing test
1000 loops, 500 sets: 1.263 ms per loop, 2.526 us per set
Running set timing test
1000 loops, 500 sets: 1.192 ms per loop, 2.384 us per set
```

Python 3.6:
```
$ python3.6 test_recombination_speed.py 
Running extend timing test
1000 loops, 500 sets: 1.075 ms per loop, 2.150 us per set
Running queue timing test
1000 loops, 500 sets: 1.469 ms per loop, 2.938 us per set
Running set timing test
1000 loops, 500 sets: 1.238 ms per loop, 2.477 us per set
```

pypy2:
```
$ pypy test_recombination_speed.py 
Running extend timing test
1000 loops, 500 sets: 0.132 ms per loop, 0.263 us per set
Running queue timing test
1000 loops, 500 sets: 0.425 ms per loop, 0.851 us per set
Running set timing test
1000 loops, 500 sets: 0.313 ms per loop, 0.627 us per set
```

pypy3:
```
$ pypy3 test_recombination_speed.py 
Running extend timing test
1000 loops, 500 sets: 0.149 ms per loop, 0.297 us per set
Running queue timing test
1000 loops, 500 sets: 0.339 ms per loop, 0.677 us per set
Running set timing test
1000 loops, 500 sets: 0.236 ms per loop, 0.472 us per set
```

I'll merge as is and make a separate PR to switch back to my string-based version.  Even if we later decide that the order of the indices isn't guaranteed to be sorted, we're sorting in the `set` version *anyway*, so it's not a big deal to sort a list instead of a set. Running the same tests where the extend version sorts first gives similar speeds between list and set.

pganssle:
Merging now, I'll clean up some of the things I've mentioned here later.

dateutil/parser.py
jbrockmendel: There's some interesting stuff here.  The pypy results are impressive.

PR #420, opened 2017-07-18
Fix incorrect index check, add test that fails pre-PR

jbrockmendel:
In the status-quo, [L824-825](https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py#L824) checks:

```
                        if i < len_l and l[i] == ':':
                            res.second, res.microsecond = _parsems(l[i+1])
```

Avoiding an IndexError requires checking `i+1 < len_l`.  This is easy to miss because the try/except catches IndexError.  One way to trigger this is:

```
import dateutil.parser
dateutil.parser.IndexError = SyntaxError
dateutil.parser.parse('2017-07-17 06:21:', fuzzy=True)
```

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -821,7 +821,7 @@ class parser(object):
 
                         i += 1
 
-                        if i < len_l and l[i] == ':':
+                        if i+1 < len_l and l[i] == ':':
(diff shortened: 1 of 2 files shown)


pganssle:
Ugh, this giant try/catch block really has to go.

PR #425, opened 2017-07-22
Separate out "while True" loop for e.g. 12 h00m04s

jbrockmendel:
The `while` loop that currently handles things like "12 h00m04s" is one of the hardest pieces of the parser to reason about.  This is in part because it tries to capture all of these tokens in one pass.  This PR implements a version that allows for looking backwards, so each of the ["12", "00", "04"] gets individually processed.

This also handles the lookback previously at L793.

Last, a tox testenv is added for lightweight testing of just the parser.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+[run]
+omit =
+    .tox/*
+    .git/*
+    .hypothesis/*
+    __pycache__
+    */site-packages/*
+    */dist-packages/*
+
+[report]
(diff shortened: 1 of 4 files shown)


pganssle:
This assumes #419 is merged, correct?

jbrockmendel:
I think I wrote it separately on the assumption I'd rebase when the time came.

pganssle:
Superceded by #436

Issue #410, opened 2017-07-05
Add support of VALUE=DATE and TZID when parsing EXDATE rule


mlorant:
According to the RFC 5545, the EXDATE grammar is:

       exdate     = "EXDATE" exdtparam ":" exdtval *("," exdtval) CRLF

       exdtparam  = *(
                  ;
                  ; The following are OPTIONAL,
                  ; but MUST NOT occur more than once.
                  ;
                  (";" "VALUE" "=" ("DATE-TIME" / "DATE")) /
                  ;
                  (";" tzidparam) /
                  ;
                  ; The following is OPTIONAL,
                  ; and MAY occur more than once.
                  ;
                  (";" other-param)
                  ;
                  )

So all the following examples are valid (let's suppose some valid RRULE preceed the exception set here):

     EXDATE:20170705T143000
     EXDATE;VALUE=DATE-TIME:20170705T143000
     EXDATE;VALUE=DATE:20170705
     EXDATE;TZID=Europe/Paris:20170705T143000
     EXDATE;VALUE=DATE-TIME;TZID=Europe/Paris:20170705T143000

Yet, dateutil only supports the two first formats. Indeed, DATE and TZID params are marked as unsupported (see `_parse_rfc` in rrule.py). Digging into the code of dateutil, I don't see any use of the `parms` list except to check list of param name given in the recurrence rule. However this change might be expanded to other properties, such as DTSTART, as discussed in https://github.com/dateutil/dateutil/issues/401?

PR #429, opened 2017-07-27
Allow DTSTART parm VALUE=DATE-TIME (fixes #401)

### LINKED ISSUE #401: ValueError: unsupported DTSTART parm: VALUE=DATE-TIME
I reported this on khal:
https://github.com/pimutils/khal/issues/679
and radicale:
https://github.com/Kozea/Radicale/issues/646

When trying When trying to sync a calendar built from khal to radicale I get

File "/usr/lib/python3.4/site-packages/vobject/base.py", line 183, in transformToNati
ve
return self.behavior.transformToNative(self)
File "/usr/lib/python3.4/site-packages/vobject/icalendar.py", line 989, in transformT
oNative
obj.registerTzinfo(obj.tzinfo)
File "/usr/lib/python3.4/site-packages/vobject/icalendar.py", line 150, in gettzinfo
return tz.tzical(buffer).get()
File "/usr/lib/python3.4/site-packages/dateutil/tz/tz.py", line 1109, in init
self._parse_rfc(fobj.read())
File "/usr/lib/python3.4/site-packages/dateutil/tz/tz.py", line 1230, in _parse_rfc
cache=True)
File "/usr/lib/python3.4/site-packages/dateutil/rrule.py", line 1603, in call
return self._parse_rfc(s, **kwargs)
File "/usr/lib/python3.4/site-packages/dateutil/rrule.py", line 1562, in _parse_rfc
raise ValueError("unsupported DTSTART parm: "+parm)
ValueError: unsupported DTSTART parm: VALUE=DATE-TIME

Which is due to rrule.py not wanting any parameter to DTSTART.

The following patch is ignoring the parameter and solves the issue for me, until someone comes with a better fix. 

```
$ cat /tmp/python-dateutil-fix-dtstart.patch
diff --git a/dateutil/rrule.py b/dateutil/rrule.py
index da94351..b447600 100644
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -1559,7 +1559,8 @@ class _rrulestr(object):
                     exdatevals.append(value)
                 elif name == "DTSTART":
                     for parm in parms:
-                        raise ValueError("unsupported DTSTART parm: "+parm)
+                        if parm != "VALUE=DATE-TIME":
+                            raise ValueError("unsupported DTSTART parm: "+parm)
                     if not parser:
                         from dateutil import parser
                     dtstart = parser.parse(value, ignoretz=ignoretz,
$
```

Unrud:


pganssle:
Does this actually *fix* #401, or just stop #401 from raising an error? I agree that the RFC says that this VALUE=DATE-TIME is a valid value for the parameter, but it seems weird to just ignore it. Is it somehow superfluous information?

Unrud:
> Is it somehow superfluous information?

Yes. The type of ``DTSTART`` can be either ``DATE-TIME`` or ``DATE``. ``DATE-TIME`` is the default and supported by **dateutil**. ``DATE`` is not supported (e.g. #410).

pganssle:
OK, when I have a bit of time I'll check the RFC to double-check, but if so this is fine.

Unrud:
> OK, when I have a bit of time I'll check the RFC to double-check

[rfc5545-3.8.2.4](https://tools.ietf.org/html/rfc5545#section-3.8.2.4):

```
3.8.2.4.  Date-Time Start

   Property Name:  DTSTART
...
   Value Type:  The default value type is DATE-TIME.
...
   Format Definition:  This property is defined by the following
   ...
                  (";" "VALUE" "=" ("DATE-TIME" / "DATE")) /
                  (";" tzidparam) /
```

[rfc5545-3.6.5](https://tools.ietf.org/html/rfc5545#section-3.6.5):

```
3.6.5.  Time Zone Component

   Component Name:  VTIMEZONE
...
   Note:
   ...
      The mandatory "DTSTART" property gives the effective onset date
      and local time for the time zone sub-component definition.
      "DTSTART" in this usage MUST be specified as a date with a local
      time value.
```

pganssle:
What's interesting here is that in this specific `if`/`elif` block, *other* `name` values explicitly support `VALUE=DATE-TIME`, so whoever originally implemented this (I assume Gustavo) knew about these parameters and explicitly did not implement it for `DTSTART`.

pganssle:
Ah, I see the issue. `rrule` follows the RFC 2445 spec, which was later superceded by RFC 5545. RFC 2445 does not support `VALUE=DATE-TIME`, but RFC 5545 does.

pganssle:
OK, I've got my head around how this works now. Thanks for the PR and report @Unrud.

Unrud:
I don't think that ``VALUE=DATE`` is allowed.

> "DTSTART" in this usage MUST be specified as **a date with a local time value**.

pganssle:
@Unrud I was just going by the bit you quoted above:

```
> ";" "VALUE" "=" ("DATE-TIME" / "DATE")) 
```

Seems to say it can be a date or a datetime, and can have a TZ-ID parameter as well. I'm not sure exactly what to make of the `otherparam` part. I'm fairly confident that `DATE` is allowed, though, since in [RFC7529](https://tools.ietf.org/html/rfc7529#section-4.3.1) (an update to RFC 5545), many examples use `VALUE=DATE`.

Unrud:
> @Unrud I was just going by the bit you quoted above:
> ```
> > ";" "VALUE" "=" ("DATE-TIME" / "DATE"))
> ```

The quote is about ``DTSTART`` in general. But the same comment also contains a second quote, that is specific to ``DTSTART`` inside of ``VTIMEZONE``.

Unrud:
> and can have a TZ-ID parameter as well.

Not if it's inside of ``VTIMZEONE``:

> "DTSTART" in this usage MUST be specified as a date with a **local** time value.

pganssle:
Hm.. I think that needs to be fixed in `tz.tzical`, since `rrulestr` definitely supports it.

pganssle:
Actually, just tested it, `rrulestr` is not affected by this PR, because `tzical` fails if `DTSTART` provides any parameters:

```python
from dateutil import tz
from io import StringIO

TZICAL_EST5EDT = """
BEGIN:VTIMEZONE
TZID:US-Eastern
LAST-MODIFIED:19870101T000000Z
TZURL:http://zones.stds_r_us.net/tz/US-Eastern
BEGIN:STANDARD
DTSTART;VALUE=DATE-TIME:19671029T020000
RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=10
TZOFFSETFROM:-0400
TZOFFSETTO:-0500
TZNAME:EST
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:19870405T020000
RRULE:FREQ=YEARLY;BYDAY=1SU;BYMONTH=4
TZOFFSETFROM:-0500
TZOFFSETTO:-0400
TZNAME:EDT
END:DAYLIGHT
END:VTIMEZONE
"""

tz.tzical(StringIO(TZICAL_EST5EDT))   # Raises ValueError.
```

pganssle:
Nope, my bad, was using the release branch, on `master` this is indeed using the standard call.

Unrud:
> Hm.. I think that needs to be fixed in ``tz.tzical``, since ``rrulestr`` definitely supports it.

My bad! I thought that these functions are exclusively used to parse ``VTIMEZONE`` components, because they seem to lack support for the ``DATE`` type or ``TZID`` that would be required otherwise.

Issue #342, opened 2017-01-16
Memoize time zones


pganssle:
A huge amount of the time zone functions are called multiple times with identical inputs and expecting identical outputs, and the whole concept of a `tzinfo` is that it should be more-or-less a singleton, so there's a huge potential for memoization that is not realized.

Ideally, we'll stub out an equivalent of `functools.lrucache` so that it's backward compatible. I will update this list later, but here's a starter of things that need to be memoized:

[] `gettz`
[] `tzfile` (`utcoffset`, `dst`, `tzname`)
[] `tzical` (already done to some extent, but we should convert this over to the same way everything else is)

Likely all non-static offsets will need `utcoffset`, `dst` and `tzname` memoized.

The big concern with doing this is that it will likely increase memory consumption in downstream applications, some of which may be operating in memory-constrained environments. I'm thinking the best way to address this is to query an environment variable at import-time that sets the length of the caches to something other than default.

pganssle:
Since this changes the API, I'm pushing this to the 2.7.0 release.

pganssle:
Pushing this to a later feature release.

pganssle:
I started doing this to see if I could improve the speed of some of these time zones and I noticed that one major problem here is that all the `tzinfo` interface functions take `dt` as a parameter, and the hash function for datetime involves evaluating `utcoffset()`, which causes an infinite recursion.

What's worse, `datetime.replace` is actually quite a slow operation, so `dt.replace(tzinfo=None)` before looking up the cached value is quite expensive. Likely we'll need to refactor to find something inherently more cache-able.

Ideally, `datetime` would add a mutate-once parameter that we could use to cache its time zone index.

PR #430, opened 2017-07-27
Protect tzical cache with a lock (fixes #428)

### LINKED ISSUE #428: tzical is not thread-safe
```
...
  File "dateutil/tz/tz.py", line 1109, in utcoffset
    return self._find_comp(dt).tzoffsetto
  File "dateutil/tz/tz.py", line 1062, in _find_comp
    return self._cachecomp[self._cachedate.index((dt, self._fold(dt)))]
IndexError: list index out of range
```

``_cachedate`` and ``_cachecomp`` are not protected by a lock.

Unrud:


pganssle:
Hm.. This isn't really how I want to fix this (I am planning on ripping out a lot of the cache stuff from `tzical` anyway and replacing it), but if we get close to a new release and I still haven't followed up on #342, I'll accept this.

Unrud:
> Hm.. This isn't really how I want to fix this

This was meant as a provisional fix, until the rewrite is completed.

pganssle:
@Unrud Yeah. I don't think we're going to do any more bugfix releases for the 2.6.0 branch, but if I don't fix caches before the 2.7.0 release I'll merge this.

pganssle:
Actually, this has no changes to the public interface, so I can always just rip it out later if I need to change it. Doesn't hurt to merge it.

PR #436, opened 2017-08-05
Get rid of while-loop version of parse_hms

jbrockmendel:
This is a rebased version of (most of) #425.

Also a couple flake8 fixups.

Some pieces need docstrings.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -507,8 +507,9 @@ class _ymd(list):
                     year, day, month = self
 
             else:
+                split_tzstr = _timelex.split(self.tzstr)
                 if (self[0] > 31 or
-                    self.find_probable_year_index(_timelex.split(self.tzstr)) == 0 or
+                    self.find_probable_year_index(split_tzstr) == 0 or
                         (yearfirst and self[1] <= 12 and self[2] <= 31)):
                     # 99-01-01
                     if dayfirst and self[2] <= 12:
@@ -725,11 +726,20 @@ class parser(object):
                     # Token is a number
                     len_li = len(l[i])
 
-                    if (len(ymd) == 3 and len_li in (2, 4) and
-                        res.hour is None and
-                        (i + 1 >= len_l or
-                         (l[i + 1] != ':' and
-                          info.hms(l[i + 1]) is None))):
+                    hms_idx = _find_hms_idx(i, l, info, allow_jump=True)
+                    if hms_idx is not None:
+                        # HH[ ]h or MM[ ]m or SS[.ss][ ]s
+                        (i, hms) = _parse_hms(i, l, info, hms_idx)
+                        if hms is not None:
+                            # TODO: checking that hour/minute/second are not
+                            # already set?
+                            _assign_hms(res, value_repr, hms)
+
+                    elif (len(ymd) == 3 and len_li in (2, 4) and
+                          res.hour is None and
+                          (i + 1 >= len_l or
+                          (l[i + 1] != ':' and
+                           info.hms(l[i + 1]) is None))):
                         # 19990101T23[59]
                         s = l[i]
                         res.hour = int(s[:2])
@@ -766,26 +776,6 @@ class parser(object):
 
                             if len_li > 12:
                                 res.second = int(s[12:])
-
-                    elif ((i + 1 < len_l and info.hms(l[i + 1]) is not None) or
-                            (i + 2 < len_l and l[i + 1] == ' ' and
-                             info.hms(l[i + 2]) is not None)):
-                        # HH[ ]h or MM[ ]m or SS[.ss][ ]s
-                        i = _parse_hms(i, l, info, res)
-
-                    elif (i + 1 == len_l and l[i - 1] == ' ' and
-                          info.hms(l[i - 2]) is not None):
-                        # X h MM or X m SS
-                        idx = info.hms(l[i - 2])
-
-                        if idx == 0:               # h
-                            (res.minute, res.second) = _parse_min_sec(value)
-                        elif idx == 1:             # m
-                            res.second, res.microsecond = _parsems(value_repr)
-
-                        # We don't need to advance the tokens here because the
-                        # i == len_l call indicates that we're looking at all
-                        # the tokens already.
 
                     elif i + 2 < len_l and l[i + 1] == ':':
                         # HH:MM[:SS[.ss]]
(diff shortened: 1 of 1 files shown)


dateutil/parser.py
```diff
@@ -725,11 +726,20 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
                     # Token is a number
                     len_li = len(l[i])
 
-                    if (len(ymd) == 3 and len_li in (2, 4) and
-                        res.hour is None and
-                        (i + 1 >= len_l or
-                         (l[i + 1] != ':' and
-                          info.hms(l[i + 1]) is None))):
+                    hms_idx = _find_hms_idx(i, l, info, allow_jump=True)
+                    if hms_idx is not None:
```
pganssle: Just trying to understand the reasoning - why was the `_find_hms_idx` call moved to the beginning here?

Is it because you need to get *and store* the results of `_find_hms_idx` within the `if` call, so you've moved it to the front?

I'm not entirely sure how I feel about this. It may not be hurting anything, but I don't understand the reasoning of the original loop well enough to tell why they are ordered the way they are. At the very least, it seems possible that this will induce some performance hit in some important formats like ISO-8601.

This feels *horribly* un-pythonic to me, but one possible work-around for "need to check *and store* the result" would be some sort of mutable object modified in the function, e.g.:

```python
class MutableReturn(object):
    def __init__(self, val):
        self.return_value = val

def f(x, mr):
    rv = x - 7

    if rv > 2:
        rv = None

    mr.return_value = rv

    return rv

if __name__ == "__main__":
    mr = MutableReturn(None)

    for v in [8, 10]:
        if f(v, mr) is not None:
            print('Value was {}'.format(mr.return_value))
```

Another option that is effectively the same thing but may be saner in this case is to have `_find_hms_idx` cache its last return value as a mutable property, e.g.:

```python
def f(x):
    rv = x - 7

    if rv > 2:
        rv = None

    f.last_return = rv
    return rv

if __name__ == "__main__":
    mr = MutableReturn(None)

    for v in [8, 10]:
        if f(v) is not None:
            print('Value was {}'.format(f.last_return))
```

I think the second one's a bit more elegant, but the first one is thread safe since it doesn't manipulate any globals.

Of course, we'll make `_find_hms_idx` a member function of `parser` once this PR is wrapped up (or as part of it if you are making changes anyway), so we could always use the argument that you can get thread safety by spawning a new `parser` object for each thread - though that's kinda thin to avoid creating a mutable return object.

dateutil/parser.py
```diff
@@ -507,8 +507,9 @@ def resolve_ymd(self, yearfirst, dayfirst):
                     year, day, month = self
 
             else:
+                split_tzstr = _timelex.split(self.tzstr)
```
jbrockmendel: Yes, this was just a PEP8 thing.  Happy to revert it.

dateutil/parser.py
```diff
@@ -725,11 +726,20 @@ def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
                     # Token is a number
                     len_li = len(l[i])
 
-                    if (len(ymd) == 3 and len_li in (2, 4) and
-                        res.hour is None and
-                        (i + 1 >= len_l or
-                         (l[i + 1] != ':' and
-                          info.hms(l[i + 1]) is None))):
+                    hms_idx = _find_hms_idx(i, l, info, allow_jump=True)
```
jbrockmendel: I think that was an option used in an earlier implementation and can now be removed.

In the loop-based implementation being replaced, only the first step of this check is allowed to look past a space for a h/m/s label.  This implementation should make that irrelevant.

dateutil/parser.py
```diff
@@ -1282,56 +1272,73 @@ class InvalidTimeError(InvalidDatetimeError):
...
+        # e.g. the "04" in "12h04"
+        hms_idx = idx-1
 
-        idx = info.hms(l[i + 1])
-        while True:
-            if idx == 0:
-                res.hour = int(value)
-                if value % 1:
-                    res.minute = int(60 * (value % 1))
+    elif (1 < idx == len(tokens)-1 and tokens[idx-1] == ' ' and
```
jbrockmendel: Yes, good catch.

dateutil/parser.py
jbrockmendel: > Is it because you need to get and store the results of _find_hms_idx within the if call, so you've moved it to the front?

Yes.  I don't especially care about the ordering, but I do really like having all of these cases be `elif`s.  This is easy to move back to its original location.  I think the much simpler solution is to be willing to call `_find_hms_idx` twice.  It can be moved back down to after the `elif len_li in (8, 12, 14):` block and changed to:

```
if _find_hms_idx(i, l, info, allow_jump=True) is not None:
    hms_idx = _find_hms_idx(i, l, info, allow_jump=True)
    [...]
```

This is a pretty cheap function call.  Let's at least do some profiling before worrying about caching (much less mutability and thread-safety).

pganssle: Yeah, I'm fine with calling it twice.

Honestly, I think if we implement #125 and add format caching, we'll see some dramatic speedup in nearly all real-world use-cases, so little stuff won't really "add up" in any real sense.

jbrockmendel:
Just pushed two commits.  One of them adds a lightweight testenv for parser (with profiling).  The other addresses the comments/suggestions above.

pganssle:
I'm not overly familiar with `tox` - it seems like the `parser` testenv runs automatically when you run `tox`, but the only difference is that it runs a subset of the tests, with coverage and profiling. Is there a way to make it so that it's separate from (or maybe even on top of) the standard `tox` run?

I could see it being very useful to have module-specific testenvs set up for rapid development, but I'd think half the value of having lighter testenvs would be so you can quickly run them on all the supported environments, but this just uses `python`.

Is there anything specific about `parser` that justifies having a separate lighter test environment just for that? If not, it might be worth making a separate, generic helper script that you can run like `tox subset parser` to get one for parser, `tox subset relativedelta` for `relativedelta` dev, etc.

Either way, maybe we can do that as a separate PR.

jbrockmendel:
> I could see it being very useful to have module-specific testenvs set up for rapid development

That's all this is, yes.

Removing "parser" from the "envlist" will leave it available under "tox -e parser" without having it run with just "tox".

> Either way, maybe we can do that as a separate PR.

Sure.

PR #439, opened 2017-08-06
Improve RFC5545 support in tzical, tweak tests.

pganssle:
Per [these comments on #429](https://github.com/dateutil/dateutil/pull/429#issuecomment-320507666) according to [RFC 5544 3.6.5](https://tools.ietf.org/html/rfc5545#section-3.6.5), `VTIMEZONE` only supports the optional `VALUE=DATE-TIME` parameter,  so anything else is filtered out.


pganssle:
CC: @Unrud 

Looks good?

Unrud:
> Looks good?

Yes.

PR #441, opened 2017-08-09
%b-%Y-%d

jbrockmendel:
Apparently all the cool kids are transitioning to pytest.  To get a feel for it I took a few of the more easily-parametrized cases from test_parser.py and implemented them in test_parser_parametrized.py.  Explicitly looping over all the valid formats exposed a handful that are not currently handled correctly (and clearly, not currently tested).  Namely:

```
from dateutil import parser
>>> parser.parse('Sept 2003 25')
  File "/usr/local/lib/python2.7/site-packages/dateutil/parser.py", line 1182, in parse
  File "/usr/local/lib/python2.7/site-packages/dateutil/parser.py", line 581, in parse
    ret = default.replace(**repl)
ValueError: day is out of range for month
```

This happens because `resolve_ymd` never considers the possibility (month, year, day).  This PR adds that particular check.


dateutil/test/test_parser.py
```diff
@@ -10,6 +11,35 @@
...
+
+# If we have a 4-digit year, a non-numeric month (abbreviated or not),
+# and a day (1 or 2 digits), then there is no ambiguity as to which
+# token is a year/month/day.  This holds regardless of what order the
+# terms are in and for each of the separators below.
+seps = ['-', ' ', '/', '.']
+token_opts = [['%Y'], ['%b', '%B'], ['%d', '%-d']]
+
+prods = itertools.product(*token_opts)
+perms = [y for x in prods for y in itertools.permutations(x)]
```
pganssle: Not sure if this is more readable per se, but I believe this is equivalent to:

`list(itertools.chain.from_iterables(map(itertools.permutations, itertools.prod(*token_opts))))`

or possibly better:

`list(itertools.chain(*(itertools.permutations(x) for x in itertools.prod(*token_opts))))`

jbrockmendel: I find the two-line version more readable, but am pretty close to indifferent on this.

jbrockmendel:
pytest doesn't allow for yield-based tests, right?  I tried having this test run for (much) more than just 2003-09-25 and it looks like it created the tests as a list rather than a generator.

pganssle:
I can't reproduce this test failure on my local machine. Is this a travis thing?

jbrockmendel:
The Travis errors look like they are coming from `sys.modules.values()` changing size in iteration. Is this showing up elsewhere?  If not, I'd guess that pytest isn't playing nicely with... something in py 3.2, 3.3, 3.4.

pganssle:
Yeah, that was what I figured, but I'm not sure what because my machine doesn't do it when I spin up a Python 3.4 conda environment. I was thinking I'd maybe try removing the profiling stuff, then the coverage stuff sequentially and see if either of those fixes it, at least to narrow it down.

jbrockmendel:
I just pushed a commit that uses regular nose-style tests instead of pytest.  We'll see if that was the issue with the Travis error.

pganssle:
@jbrockmendel What was it you changed that fixed this? I just found out about debug mode on TravisCI, so I was thinking I'd try and debug it that way.

jbrockmendel:
I just removed pytest and changed the new test to yield nose-style.

pganssle:
@jbrockmendel I'll try and look at this in more detail this weekend, I'm not sure I understand how these `yield` tests work, and it seems like you are still actually *using* `pytest` here. You didn't squash the old broken commit, right? I'm going to try and figure out why it was broken in the first place.

jbrockmendel:
I think you're right that I didn't completely remove `pytest`.  I can make a new PR that just fixes the bug and adds the tests, completely stripping out the other changes.  Let me know your preference.

jbrockmendel:
I'd like to wrap up at least the bug-fix part of this.  Is the pytest problem still a problem?  If so, I'll make a new PR that uses the older test setup.

pganssle:
This is the most infuriating thing. I managed to open an interactive shell, but I *still* can't replicate the problem.

Anyway, yeah, I think just drop both all the pytest stuff including the yield tests. Switch that from using `assert` to using `self.assertEqual`. Normally I don't like a single test that handles a bunch of test cases, but as a temporary matter before we switch over to pytest and properly parametrized tests, we can just generate the tests inside the test function.

pganssle:
Superceded by #481

PR #446, opened 2017-08-16
Improve fold backport compatibility.

pganssle:
Turns out that pypy3's implementation of `datetime.replace` doesn't return `self.__class__`, it always returns `datetime`. This fixes that issue and makes pypy3 a required pass.


jbrockmendel:
Great.  I vaguely recall having issues with pypy and datetime long ago.  Good to see you're on top of it.

Issue #431, opened 2017-07-29
Proposal: Merge datestuff library into dateutil


justanr:
I have a small library of date/datetime helpers (creatively called [datestuff](https://github.com/justanr/datestuff) that I'd like to donate to this library. I often find myself installing both side by side and I feel the tools in datestuff are a good fit for dateutils' goal of "Useful extensions to the standard Python datetime features"

A quick rundown:

* `RelativeDate` and `RelativeDateTime` provide "unfixed" dates/datetimes. What this means is that when the class is created, it's not tied to a specific instance in time but rather to an offset. Allowing a user to create a date that is always 1 week from now is as easy as `RelativeDateTime(offset=timedelta(days=7))` rather than needing to deal with `datetime.now() + timedelta(days=7)` all over the place, the added benefit is that these can be used where ever regular date/datetime objects are. They also accept a factory if a different date/datetime object is desired. `RelativeDate(Time)` is already compatible with `relativedelta` as well.

* `DateRange` is a range of date/datetimes modeled after Python 3's range so it is lazy, slicing it results in a new range rather than a list and has fast path membership testing. I've never found a helper than scratched this itch for me. The existing `rrule` is nice, but seems more suitable for complex calendar scheduling than a range of dates. Where as `DateRange` is simple to use: Just give it a start date, an optional end date, and a timedelta to step by. I will say allowing both fixed and infinite ranges from the same object was probably a mistake and it does complicate the code some.

* `within_delta` is helper I wrote because I got tired of normalizing datetimes before comparing them.

If any copyright/license stuff needs to be signed over, I'm happy to do that (never done something like before).

pganssle:
@justanr Definitely willing to accept this sort of thing, though I'd have to look at the details.

One overall point, which is a hard circle to square, is that your class names comply with PEP8 and `dateutil`'s, unfortunately, do not. (Hence `relativedelta` instead of `RelativeDelta`). That's a discussion that can be had in the later stages of adoption, but it seems weird to have a mix of CamelCase and lowercase classes, but it also seems somewhat improper to introduce *new* modules that don't comply with the code standards.

[...]

PR #454, opened 2017-08-30
relativedelta parser

waynenilsen:
I have to deal with these things we call tenors in finance a lot and I want this library to be better at handling it. I believe that this is useful to many others. I was going to open an issue to discuss it but it was so easy to get started with implementation that I decided to just submit a PR and see what your thoughts were. 

Thanks

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -34,11 +34,12 @@ import datetime
 import re
 import string
 import time
+import sys
 
 from calendar import monthrange
 from io import StringIO
 
-from six import binary_type, integer_types, text_type
+from six import binary_type, integer_types, text_type, reraise
 
 from . import relativedelta
 from . import tz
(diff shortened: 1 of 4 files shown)


jbrockmendel:
Two quick thoughts/questions:

It looks like both upper and lower case `m` get mapped to Months.  Are minutes not relevant here?

I know @pganssle is wary of expanding the API.  This might be a good fit for `bdateutil`, or `pandas.tseries.offsets.BusinessMixin`.

waynenilsen:
I made a similar issue with `lubridate` see [this comment](https://github.com/tidyverse/lubridate/issues/459#issuecomment-240823991) quoted below

> Unfortunately, in the case of tenors M is used for month, the tenors are always assumed to be a day or greater and generally interpreted in a case-insensitive way. H and S are fair game. A popular example is LIBOR 3M the three month London interbank offered rate. There is also LIBOR 6M and EURIBOR 6M for example.

The maintainer there was keen on making some pretty big additions. I could definitely see this in `bdateutil` whatever seems preferable.

pganssle:
I don't have any real problem with this, but I think I'd like to roll it into a `parser.RelativeDeltaParser` class, not as an alternate constructor for `relativedelta`.

If it's going to be more general than tenors, though, the same issues that are dogging `parser.parser` come up - how to be open for enhancement such that users can expand if necessary.

I suggest the following thing:

1. Create a new `parser.RelativeDeltaParser` class.
2. Add a `parser._RelativeDeltaParserInfo` class that contains parser info, including a dictionary element mapping unit tokens to their appropriate unit.
3. Move this logic into a `RelativeDeltaParser.parse()` method, and utilize the dictionary mapping stuff in there. I think start with it being case-sensitive because that's what makes the most sense. You can always move the `.upper()` call to the input of the parser until we have a better handle on how to handle extension.

I'm not sure how the extension interface should look, which is why I think `_RelativeDeltaParserInfo` should be a private class and an instance of it attached only by default. So long as the public interface only exposes "call_some_func(some_string) -> relativedelta" and the tests only reflect things we *know* we already want to support, I don't see any problem expanding the API this little bit.

waynenilsen:
I will probably work on this tomorrow morning and get some new commits in. Thanks for the direction.

waynenilsen:
Ok, sorry about the noise with the commits. The diff looks good. I used the names that @pganssle suggested but they're a bit inconsistent with the lowercase naming of classes in this module otherwise. I also added a function to be consistent with the `parse` function that is included. Tests are also included.

I am not sure on the formatting for the doc strings. I tried to follow along with what was already there and looked up some things in the sphinx docs. 

Thank you for reviewing this PR

pganssle:
@waynenilsen Thanks for your excellent work on this. I'll look at it soon, but it looks good to my initial once-over. 

With regards to the capitalization convention, there were recently some discussions in #431 about the convention - PEP 8 suggests CapWords convention for classes, but we're using lowercase. It doesn't help that CPython uses lowercase for most standard library classes (including all those related to dates). I think there's some movement behind the idea "we don't have to be inconsistent" and "we'll start a slow process of moving towards CapWords", but this project doesn't really get a lot of attention so it's not like there was a spirited debate.

Before this gets released on PyPI (but not necessarily blocking merging into msater), I'd like to get a general sense of the conventions people use in representing durations like this. I'm a bit surprised that you chose M for minutes and m for months. Parsers should probably be normative about these sorts of things, not positive, and I would be kinda surprised if the convention for expressing durations goes with m == Months, M == minutes... I note that [ISO8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) durations uses `M` for *both* and encodes their meaning positionally.

Are tenors ever expressed out of-order? I think maybe the best thing to do would be to impose a big-endian constraint on strings to be parsed, so that we can resolve ambiguities in that fashion (i.e. you will never see a token for a larger unit of time than one you've already seen). We'll still need some sort of token mapping, so that users can specify whether a string like `1.5M` means "1.5 months" or "1.5 minutes", but it will be less important.

waynenilsen:
>  I'd like to get a general sense of the conventions people use in representing durations like this. I'm a bit surprised that you chose M for minutes and m for months.

I pulled the month/minute case convention from the [strftime docs](https://docs.python.org/3/library/time.html#time.strftime). The positional encoding makes sense in 8601 because it is fully specified whereas this parser is meant as something of a practical convenience where it is not necessary to fully specify every token. 

> Are tenors ever expressed out of-order?

I have never seen a tenor expressed as more than one token. What we have now is a `relativedelta` parser which goes beyond just tenors. I think enforcing big-endian makes sense but unfortunately doesn't resolve the ambiguity of the string `1m` for example but it does create meaning for `1m 1m`. 

The question then is still what is the meaning of `1m`? I think falling back on the meaning defined by `strftime` is ok, not sure what other conventions there are that are context-free. 

> We'll still need some sort of token mapping, so that users can specify whether a string like 1.5M means "1.5 months" or "1.5 minutes", but it will be less important.

For convenience I did add a [token mapping argument](https://github.com/dateutil/dateutil/pull/454/files#diff-ed798069162e7ab7db9a45da6c4a076fR1520) to the function version of the parser with a default value.

pganssle:
> The question then is still what is the meaning of 1m? I think falling back on the meaning defined by strftime is ok, not sure what other conventions there are that are context-free.

The problem with this is that the `m` in strftime is actually *replaced by values*, not used as an indicator of what the value it just printed *was*, so it's a convention for formatting strings, not a convention for string formats (generally string formats include this sort of thing positionally).

I think it's probably worth finding examples of people actually formatting these things to see how we can best get the "most likely right thing", and design the parser ahead of time such that it's not amazingly complicated to specify the context.

I'm actually less worried about the `1m` situation and more worried about the `1y1m` situation. If I saw `1y1m`, I'd think that means "1 year, 1 month", but it could also mean "1 year, 1 minute", and it only becomes ambiguous when it becomes either "1y1m1w" or "1y1d1m".

waynenilsen:
Perhaps the thing to do here is limit the scope of this PR then accept PRs for additional work as use cases come up?  I would send screenshots but the applications are proprietary so I made up some data. It is quite common to see tables such as below. This happens with in the Rates as well as FX domains. We could start with just pulling in the single token parser with an optional mapping dict. Again, if you think this is better suited to `bdateutil` or something else, I would understand.

tenor | rate
-- | --
1m | 0.01
3m | 0.0102
6m | 0.010404
1y | 0.010612
18m | 0.010824
2y | 0.011041
3y | 0.011262
5y | 0.011487
7y | 0.011717
10y | 0.011951
20y | 0.01219
30y | 0.012434
50y | 0.012682

pganssle:
@waynenilsen I think `bdateutil` is mostly for dealing with business days. We need a `relativedelta` and/or `timedelta` parser *anyway*, the main issue is that if we release something, it would be a backwards-incompatible change to *not* support the things the earlier version supports, so we want to start with a minimal subset that, crucially, won't conflict with common use cases (because the default should "just work" in most contexts).

PR #458, opened 2017-09-12
updated parse() docstring to reflect tzinfos integer offset units and example 

abrugh:
Doc says UTC offset is in minutes but the example snippet uses seconds. Actually trying it would seem to confirm seconds is correct.

The example snippet also shows BRST to be -3 hrs, https://www.timeanddate.com/time/zones/brst shows it to be -2 hours so i changed it to be -7200


pganssle:
I wonder why it was 3 hours. Maybe Brazil switched their base time zone?

Issue #459, opened 2017-09-14
weeks property doesn't handle negative days count


souliane:
Hallo,

the weeks property is defined like that:
```
    @property
    def weeks(self):
        return self.days // 7

```
If you initialize with relativedelta(days=-1), the weeks property would return -1 instead of 0...

pganssle:
Great catch! A PR fixing this would be welcome, otherwise I'll get on it as soon as I remember which rounding function has the right properties 😉

pganssle:
Fixed by #460

PR #460, opened 2017-09-14
fix weeks property getter and add some tests

souliane:
fix for https://github.com/dateutil/dateutil/issues/459

### DIFF
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -249,8 +249,7 @@ class relativedelta(object):
 
     @property
     def weeks(self):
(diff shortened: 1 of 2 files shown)


dateutil/relativedelta.py
```diff
@@ -249,7 +249,8 @@ def _fix(self):
 
     @property
     def weeks(self):
-        return self.days // 7
+        weeks = abs(self.days) // 7
```
pganssle: I think replace this with `return int(weeks/7.0)`, that will give the right zero-symmetrical floor-div behavior without the if statement.

PR #466, opened 2017-10-05
Change license information in setup.py

### LINKED ISSUE #464: License Mismatch with PyPi and Git Repo
Hi there,

I was wondering what license this project is using, as the information is contradicting.

On PyPi it says
License: Simplified BSD (which would be the 2-clause BSD license)

The LICENSE file in your git repository suggests that it is a 3-clause BSD license.

Could you fix your setup.py such that it contains the correct license information?

pganssle:
Fixes #464.

I'm not 100% on what the right thing to put in this `license` line here. Most people seem to use "BSD" to refer to either the 2-, 3- or 4- clause versions and there doesn't seem to be a single canonical name for the 3-clause license, [according to Wikipedia](https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_.28.22BSD_License_2.0.22.2C_.22Revised_BSD_License.22.2C_.22New_BSD_License.22.2C_or_.22Modified_BSD_License.22.29). I picked the most official-sounding one that seemed unambiguous, but maybe "BSD 3-clause" or "BSD (3-clause)" would be best?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(name="python-dateutil",
       author="Paul Ganssle",
       author_email="dateutil@python.org",
       url="https://dateutil.readthedocs.io",
-      license="Simplified BSD",
+      license="BSD 3-Clause",
       long_description="""
 The dateutil module provides powerful extensions to the
 datetime module available in the Python standard library.


jbrockmendel:
Does the name in setup.py have to use the version number?  Since "N-Clause BSD" is colloquial it might be worth just listing it that way.  A comment to that effect couldn't hurt.

pganssle:
OK, I've switched it over to BSD 3-Clause. That seems fairly consistent with https://www.spdx.org's nomenclature. Since there's no obvious standard, anyone scraping PyPI for information about this will have to deal with inconsistencies anyway.

PR #468, opened 2017-10-08
Refactor new parser functions to keep them in the parser object

pganssle:
As suggested in my earlier review, this moves all the parsing-related functions back into the `parser` object.


pganssle:
@jbrockmendel The reason I wanted them all methods is that they are not really reusable outside of the context of the class.

There's a case to be made that they might work better as static or class methods, since they don't rely on any state of the instance, but as you said they are private, so the details are not amazingly important.

PR #470, opened 2017-10-08
Testing BYD pytest branch

pganssle:
Another test to see if Travis will start failing again.


jbrockmendel:
Standing offer to make this same PR but with just the bug fix; can use nose-style tests in the interim.

pganssle:
@jbrockmendel Not much to worry about, I want to move us over to pytest anyway. I'm in fact in the middle of refactoring the `parser` tests with parametrized test cases, but I realized that the semantics of partial dates would be *very* helpful there, so I'm trying to get that finished first. I'm going to shoot an e-mail to Travis support to see if we can get debug mode working so I can understand what's actually happening here. If it becomes pressing I'll just merge the version that doesn't have the travis issue.

pganssle:
Closing. I have been defeated by this heisenbug.

Issue #467, opened 2017-10-05
lt gt does not work after +,- operates


crackevil:
see case with python 2.7:

a=relativedelta(minutes=10)+relativedelta(minutes=5)
b=relativedelta(minutes=15)
c=relativedelta(minutes=50)
print b<c

comment "a" out, can get a different result

jbrockmendel:
It looks like `relativedelta` doesn't define `__le__` at all, so this comparison is not well-defined.  It may be that it should raise (well, maybe not in this case since they are comparable, but in the more general case).

pganssle:
@crackevil See #350

pganssle:
By the way, since `relativedelta` does not implement `__lt__` or any other comparator, the default behavior is that `a < b` if `id(a) < id(b)`, so those comparison operators are arbitrary. Those objects are ordered by their location in memory, not their values.

crackevil:
That's fine. I think it's why timedelta of datetime package limit its parameters. But relativedelta has normalize function if I did not get wrong. Can it be the same logic to deal with month and year by means under lt & gt?

pganssle:
@crackevil `normalized` just converts *fractional* well-defined quantities into the equivalent combination of integer sub-units, e.g. 0.5 days -> 12 hours, etc. There isn't a well-defined conversion from months->days, since "month" generally means something like "until the same day in the next calendar month", which depends on the left-hand-side of the equation, so `relativedelta(months=1)` and `relativedelta(days=30)` don't have a fixed ordering, since `(datetime(2017, 1, 15) + relativedelta(months=1)) > (datetime(2017, 1, 5) + relativedelta(days=30))`, but `(datetime(2017, 2, 15) + relativedelta(months=1)) < (datetime(2017, 2, 15) + relativedelta(days=30)`.

You are free to subclass `relativedelta` and implement `__lt__` and `__gt`__ as desired on the subclass if you want a specific ordering. Alternatively, if you don't care about the details and don't want to subclass, you could just add both relativedeltas to an arbitrary date (maybe the unix epoch `datetime(1970, 1, 1)`) and compare the results, like I did above. So long as the date is consistent, it should have more-or-less consistent results.

crackevil:
@pganssle thanks, your approach is smart and helpful

Issue #350, opened 2017-03-01
relativedelta does not implement __lt__


binnisb:
`relativedelta(months=+2) < relativedelta(months=+6)`
should return be True but gets evaluated to False.

I see that relative delta does not implement __lt__. Is there any reason for that?

pganssle:
I'm not sure I agree that comparison should be implemented such that it returns `True` or `False` in that way, since I don't think it really makes sense to compare to `relativedelta` objects like that. Consider, what should be the answers to the following comparisons?

```python
relativedelta(months=1) < relativedelta(days=30)
relativedelta(months=5) < relativedelta(day=9, hour=2)
relativedelta(weekday=MO) < relativedelta(weekday=FR)
```

Basically, because `relativedelta` doesn't really map to any fixed time period, it doesn't make much sense to directly compare them.

@binnisb What version of `dateutil` and `python` are you using, specifically? I would expect `relativedelta` to raise a `TypeError`, not return `False`, and that's what I get with Python 3.6.0, dateutil 2.6.0:

```python
>>> from dateutil import __version__
>>> __version__
'2.6.0'
>>> from dateutil.relativedelta import relativedelta
>>> relativedelta(months=3) < relativedelta(months=4)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-cf9dd5f49f4d> in <module>()
----> 1 relativedelta(months=3) < relativedelta(months=4)

TypeError: '<' not supported between instances of 'relativedelta' and 'relativedelta'
```

binnisb:
My version of dateutil is a bit older:
```
In[35]: from dateutil import __version__
In[36]: __version__
Out[37]: '2.4.1'
```
Using Python Anaconda version 2.7.13

@pganssle I still feel you should be able to compare them. First, you can compare datetime.timedelta, second you can do:
```
  ref_date = datetime.date(2001,1,1)
  if ref_date + relativedelta(dt1=dates[-2], dt2=dates[-1]) < ref_date + self.period:
    print "yay"
  else:
    print "nay"
```
where self.period is a relativedelta you have a comparison, and mathematically you should be able to cancel out the ref_date to compare. Therefore there must be an ordering to the relativedeltas? Or am I misunderstanding something? Maybe I am missing the point since I only use the relative deltas as a better datetime.timedelta :)

[...]

PR #471, opened 2017-10-10
WIP: comparison methods for subset of values

jbrockmendel:
Ref #467, comparison methods should be consistent across py2/py3.  In some cases we can make a well-defined comparison, and in others we should raise or return NotImplemented.

To get the ball rolling, this looks at the subset of relativedelta values that are equivalent to timedelta objects.

### DIFF
--- a/dateutil/relativedelta.py
+++ b/dateutil/relativedelta.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import datetime
+from datetime import timedelta
 import calendar
 
 import operator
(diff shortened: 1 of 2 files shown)


pganssle:
See my reasoning in #350. I don't think this is a good idea. The problem with these things is that they only *sometimes* work, which leads to much harder-to-diagnose problems than if they *always* fail. Also, if you want to do some operations on things that are entirely `timedelta`-like, why not just use a `timedelta` in the first place?

It's not a problem for users to subclass `relativedelta` to have any sort of ordering semantics they want, and that *is* something we'll support going forward.

Additionally, they can easily define an "order" function like this:

```python
def rd_lt(rd1, rd2):
    dt = datetime(1970, 1, 1)
    return (dt + rd1) < (dt + rd2)
```

In any case, I think `relativedelta` is a bit of a weird class in itself, and is probably widely misused. I'm in the middle of refactoring it into `calendardelta` and `partialdate`. I'm thinking that `calendardelta` might have a `to_timedelta()` method, with some mechanism for specifying how ambiguous deltas are handled, which should be sufficient for ordering purposes.

pganssle:
By the way, on the subject of consistency, I don't think it really matters if the behavior is consistent between python 2 and 3 *too* much, because the existence of a default comparison operator is one of the well-known differences between Python 2 and Python 3. If we *did* want it to be consistent (and I'm not convinced of the value of this), the best way is probably to just do:

```python
    def __cmp__(self, other):
        raise TypeError('Comparison is not supported between types {} and {}'.format(type(self), type(other)))
```

PR #472, opened 2017-10-10
Add __abs__ to relativedelta

### LINKED ISSUE #350: relativedelta does not implement __lt__

pganssle:
Should resolve #350.


pganssle:
Ping @pferreir.

pferreir:
Thanks!

PR #474, opened 2017-10-14
Add dateutil.utils.today

pganssle:
This seems like a convenience function that would be useful, particularly for passing in to `parser.parse` as `default`.

### DIFF
--- a/dateutil/utils.py
+++ b/dateutil/utils.py
@@ -1,6 +1,24 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import datetime, time
+
+
+def today(tzinfo=None):
+    """
+    Returns a :py:class:`datetime` representing the current day at midnight
+
+    :param tzinfo:
+        The time zone to attach (also used to determine the current day).
+
+    :return:
+        A :py:class:`datetime.datetime` object representing the current day
+        at midnight.
+    """
+
+    dt = datetime.now(tzinfo)
+    return datetime.combine(dt.date(), time(0, tzinfo=tzinfo))
+
 
 def within_delta(dt1, dt2, delta):
     """
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ datetime module available in the Python standard library.
       package_data={"dateutil.zoneinfo": ["dateutil-zoneinfo.tar.gz"]},
       zip_safe=True,
       requires=["six"],
+      tests_require=["freezegun"],
       install_requires=["six >=1.5"],  # XXX fix when packaging is sane again
       classifiers=[
           'Development Status :: 5 - Production/Stable',
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,9 @@ matrix:
 before_install:
   # Install codecov
   - if [[ $TRAVIS_PYTHON_VERSION == '3.2' ]]; then pip install coverage==3.7.1; fi
+  - if [[ $TRAVIS_PYTHON_VERSION < '3.0' ]]; then pip install mock; fi
   - pip install codecov
+  - pip install freezegun
 
 install:
   - pip install six
--- a/dateutil/test/test_utils.py
+++ b/dateutil/test/test_utils.py
@@ -1,10 +1,32 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 from datetime import timedelta, datetime
-from dateutil.utils import within_delta
+
 import unittest
 
+from dateutil import tz
+from dateutil.utils import within_delta, today
+
+from freezegun import freeze_time
+
+UTC = tz.tzutc()
+NYC = tz.gettz("America/New_York")
+
+
 class UtilsTest(unittest.TestCase):
+    @freeze_time(datetime(2014, 12, 15, 1, 21, 33, 4003))
+    def testToday(self):
+        self.assertEqual(today(), datetime(2014, 12, 15, 0, 0, 0))
+
+    @freeze_time(datetime(2014, 12, 15, 12), tz_offset=5)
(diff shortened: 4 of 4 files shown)


dateutil/utils.py
```diff
@@ -1,6 +1,24 @@
...
+    :param tzinfo:
+        The time zone to attach (also used to determine the current day).
+
+    :return:
+        A :py:class:`datetime.datetime` object representing the current day
+        at midnight.
+    """
+
+    dt = datetime.now(tzinfo)
+    return datetime.combine(dt.date(), time(0, tzinfo=tzinfo))
```
pganssle: Responding to [this comment](https://github.com/dateutil/dateutil/pull/475#discussion_r144706735) by @jbrockmendel:

> +1 for today(tzinfo), but I'm very wary of implementing behavior that differs from same-named stdlib behavior. btw:

> ```python
> >>> tz = pytz.timezone('US/Pacific')
> >>> pd.Timestamp.today(tz)
> Timestamp('2017-10-14 17:19:04.081315-0700', tz='US/Pacific')
> ```

My reason for calling it `today` was because, frankly, I think that this is close to `datetime.datetime.today()` *should* do, and I think it violates principle of least surprise to find that `datetime.today()` is basically just an alias for `datetime.now()`. That said, the name overloading *is* my main concern about this function.

I think there are a few reasonable ways forward:

1. Leave as is because this is what `datetime.today()` *should have been* and people are more likely to be shocked that `datetime.datetime.today()` does something else than that `dateutil.utils.today()` does what it does.
2. Change the signature of this so that `utils.today()` takes an optional `time` argument in addition to an optional `tzinfo` argument. This makes it more clear that what it does is give you "today at this time, defaulting to midnight". The reason I don't like this is that there could be a conflict between `time` (which may have a `tzinfo` of its own) and `tzinfo`, and also the interface isn't nearly as clean.
3. Change the name to something like `today_midnight()` or `today_datetime()` or something.

I think I'm inclined to leave it as is, considering `pendulum` has a similar real distinction between its `now` and `today` behavior:

```python
>>> import pendulum
>>> print(pendulum.now())
2017-10-15T17:30:33.887480-04:00
>>>> print(pendulum.today())
2017-10-15T00:00:00-04:00
```

setup.py
```diff
@@ -26,6 +26,7 @@
       package_data={"dateutil.zoneinfo": ["dateutil-zoneinfo.tar.gz"]},
       zip_safe=True,
       requires=["six"],
+      tests_require=["freezegun"],
```
pganssle: I should note that I'm not sure that this is the most foolproof way of installing test-specific dependencies, as the only thing that cares about this particular line is, AFAIK, `python setup.py test`, and if we're going to move to `pytest`, we'll either want our own `setup.py` test runner that uses `pytest`, and/or we should have a separate script that declares and installs the test dependencies independent of `python setup.py test`. Not sure what most people do.

jbrockmendel:
I'm inclined to agree about the ideal behavior of `datetime.datetime.today`, but pushing back against that is like fighting the tides.

> that datetime.today() is basically just an alias for datetime.now()

Note there is a small difference: `now` will accept a tzinfo argument.

If this is intended to be user-facing, then the name should avoid adding to the morass here.  If it is intended to be internal, I would be fine with `_today`.

pganssle:
@jbrockmendel It's definitely intended to be external-facing.

I honestly don't see a big problem with diverging from `datetime.datetime.today` - `datetime.datetime.today` is an *alternate constructor* for `datetime`. This is a utility that gives you today at midnight. To me, the main problem with the name is mostly that it doesn't convey that it's going to give you a `datetime` back just from the name. I think it will be far less confusing than Python's current behavior.

> Note there is a small difference: now will accept a tzinfo argument.

Wow, I didn't even realize this. So it's basically a *worse* version of `now()`. I don't think we should be emulating this API at all.

jbrockmendel:
The downside of name overlap strikes me as much bigger than the upside from a one-liner.  Since we'll be stuck with the API indefinitely, a distinct name seems like the way to go.  That said, its not worth arguing about, so do your thing.

pganssle:
I don't really think there's a serious downside from the name overlap. It's not even possible to import `datetime.datetime.today` as a bare function, so the semantics of a bare function and an alternate constructor are completely different. The real problem I see is that the semantics of the two would be reversed - the default value for unspecified time components of `datetime` is 0, which is why it's bizarre that `datetime.today` (which you would expect to be the equivalent of calling `date.today()` and combining it with an empty time component) is a shittier version of `datetime.now()`, whereas a bare function called `today()` seems like it *should* return a `date`.

I'm really torn between:

- `today()` - seems like the most natural and most people will *probably* intuitively get what it does
- `today_datetime()` - This is explicit about what it does, but more verbose than I prefer.
- `datetime_today()` - Again, explicit, but this actually *might* get confused for `datetime.today()`

So far I'm still liking `today()` most. If I've learned anything from fielding issues on this repo and on the SO `python-dateutil` tag for a few years is that people don't read documentation, and they'll be very confused if something doesn't do what it seems like it would do. People will probably not complain if `today()` returns a `datetime`, because I really get the impression that people don't use the `date` class much at all and are expecting full `datetime` objects every time. They would *also* likely complain if `today()` returned what `datetime.today()` returns, since it's confusing and counter-intuitive. Likely they wouldn't complain about `today_datetime()`, but they'd be less likely to use it, because anything that's not a super concise one-liner seems not worth it to most people I guess.

pganssle:
OK, I'm going to go ahead and merge this. I doubt anyone will be too confused that `today()` returns a `datetime` (very few people care about the difference between `date` and `datetime`), and I could always justify it a bit with an `at` keyword argument that lets you specify the time as well.

Issue #94, opened 2015-06-15
How to specify a default time zone for parse(), and nothing else


nealmcb:
I'd like to be able to provide a default timezone when parsing, to be used only if the dates I'm parsing don't specify one.  That way my code would work in an ideal world, where all the dates provided to me have clear timezones, and also in the real world, where I have to figure out what time zone they expect me to assume, and tell the parser that.  Sort of the opposite of ignoring the timezone via `ignoretz`.

That is exactly what `tzinfos` seemed to offer, based on the 2.4.2 documentation.  But since that's not the case (as noted in #80), it would be great to document a clean way to do it.

This works pretty well:

    >>> print dateutil.parser.parse("00:00", default=dateutil.parser.parse("00:00Z"))
    2015-06-15 00:00:00+00:00

But I worry that it is providing more defaults than I really want to, like hours, minutes, and it is even more unwieldy when I want a default other than UTC.

So I'd prefer a more explicit and efficient shorthand, like `parse("00:00", defaulttz=tzutc())`

Note also, at https://dateutil.readthedocs.org/en/latest/parser.html: _"If a time zone is omitted, it is assumed to be UTC"_.  That seems to be wrong: I get tzinfo of None when parsing a string with no time zone:

    >>> print dateutil.parser.parse("2000-01-01T00:00").tzinfo
    None

Even so, I love dateutil.  Thank you!

pganssle:
The documentation should be updated now, that assumption of UTC was incorrect on my part - I went through and wrote that documentation aspirationally (e.g. I wrote documentation that I felt matched my intuitions about default values, *then* wrote code for it), but then forgot to remove that part when I realized that if a timezone is specified then it's best to return timezone-naive objects, not UTC objects.

As for allowing the default timezone to be specified, that might be reasonable, but I'm a bit cautious about it because the argument parsing logic is already decently extensive and complicated, I'd rather pruning that back rather than moving it forward.

Depending on how unweildy or quick you want to make your code, there are a few ways to accomplish this, but I think that the easiest thing to do is to just write a wrapper that adds a tzinfo file to a date if none exists:

```python
>>> from dateutil.parser import parse
>>> from dateutil import tz

>>> add_default_tz = lambda x, tzinfo: x.replace(tzinfo=x.tzinfo or tzinfo)

>>> EST = tz.gettz("America/New_York")
>>> parsed_dt = add_default_tz(parse("0:00"), EST)

>>> parsed_dt
[...]

PR #475, opened 2017-10-14
Add default tzinfo

pganssle:
As a partial and possibly more general fix to #94, this adds a `tzinfo` to any `datetime` that doesn't have one, and otherwise leaves it alone.

### DIFF
--- a/dateutil/test/test_utils.py
+++ b/dateutil/test/test_utils.py
@@ -5,7 +5,8 @@ from datetime import timedelta, datetime
 import unittest
 
 from dateutil import tz
-from dateutil.utils import within_delta, today
+from dateutil import utils
+from dateutil.utils import within_delta
 
 from freezegun import freeze_time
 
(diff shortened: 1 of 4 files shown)


pganssle:
Ping @nealmcb

nealmcb:
Thanks for working on this!

So my proposal was implementing something like this:

```python
>>> parse("00:00", defaulttz=tzutc())
```
And it seems that this would be the proposed implementation from this PR https://github.com/dateutil/dateutil/pull/475/commits/a6aae6780c200267e66bf09948442a7764451074:

```python
>>> from dateutil.utils import default_tzinfo
>>> default_tzinfo(parse("0:00"), tzutc())
```

I guess that works, and if people get their date objects from somewhere other than `parse`, it is indeed more general. But it is less discoverable for people looking at the `parse` documentation. Would it make sense to put an example there?

nealmcb:
I guess a practical example would be something that takes two strings to be parsed from some list of meeting times, in which one string has timezone info, and the other doesn't. Both could be parsed, with a default of the `NYC` timezone provided, and the results printed out.

dateutil/utils.py
```diff
@@ -1,6 +1,44 @@
...
+    :param tzinfo:
+        The time zone to attach (also used to determine the current day).
+
+    :return:
+        A :py:class:`datetime.datetime` object representing the current day
+        at midnight.
+    """
+
+    dt = datetime.now(tzinfo)
+    return datetime.combine(dt.date(), time(0, tzinfo=tzinfo))
```
jbrockmendel: +1 for `today(tzinfo)`, but I'm _very_ wary of implementing behavior that differs from same-named stdlib behavior.  btw:

```
>>> tz = pytz.timezone('US/Pacific')
>>> pd.Timestamp.today(tz)
Timestamp('2017-10-14 17:19:04.081315-0700', tz='US/Pacific')
```

dateutil/utils.py
```diff
@@ -1,6 +1,44 @@
...
+        The :py:class:`datetime.tzinfo` subclass instance to assign to
+        ``dt`` if (and only if) it is naive.
+
+    :return:
+        Returns an aware :py:class:`datetime.datetime`.
+    """
+    if dt.tzinfo is not None:
+        return dt
+    else:
+        return dt.replace(tzinfo=tzinfo)
```
jbrockmendel: lgtm

dateutil/test/test_utils.py
```diff
@@ -1,10 +1,43 @@
...
 from datetime import timedelta, datetime
-from dateutil.utils import within_delta
+
 import unittest
 
+from dateutil import tz
+from dateutil import utils
+from dateutil.utils import within_delta
+
+from freezegun import freeze_time
```
jbrockmendel: +1

pganssle:
@nealmcb Yes, that's not a bad idea. After we settle on the final interface here, I can add it in to the documentation for the parser.

dateutil/utils.py
pganssle: @jbrockmendel Sorry, I miscalculated what would be shown in the diff here, this discussion *should* be taking place on #474, so I've responded [here](https://github.com/dateutil/dateutil/pull/474/files#r144732646)

pganssle:
@nealmcb Added an example in the `default_tzinfo` documentation. Look good?

jbrockmendel:
LGTM

PR #481, opened 2017-10-26
Add fix for b-Y-d cases; add tests; remove redundant tests

jbrockmendel:
Replaces #441

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -39,7 +39,7 @@ class TestFormat(unittest.TestCase):
                 for fmt in unambig_fmts:
                     dstr = actual.strftime(fmt)
                     res = parse(dstr, default=datetime(year, month, day))
-                    assert res == actual
+                    self.assertEqual(res, actual)
 
 
 class ParserTest(unittest.TestCase):
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -11,6 +11,12 @@ from dateutil.parser import parse, parserinfo
 from six import assertRaisesRegex, PY3
 from six.moves import StringIO
 
+try:
+    datetime.now().strftime('%-d')
+    PLATFORM_HAS_DASH_D = True
+except ValueError:
+    PLATFORM_HAS_DASH_D = False
+
 
 class TestFormat(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


dateutil/test/test_parser.py
```diff
@@ -11,6 +12,36 @@
 from six.moves import StringIO
 
 
+def test_Ybd_permutations():
```
pganssle: This doesn't actually test anything unless you run it with `pytest` (and even then it's a deprecated way to do this). The function should be moved into the `TestCase` object until we move over to `pytest`, at which point we'll just switch to the parametrized version.

pganssle: Note that by "moved to" I mean that it should not `yield` anything, and the `assert` statement should be switched to a `self.assertEqual()` call.

jbrockmendel:
I forgot we're not using `nose`, will fix this afternoon.

jbrockmendel:
Just pushed; added a new test class in the hopes that we can avoid the one-giant-class pattern.

It occurs to me: these tests should be robust to "fuzzy", "dayfirst", and "yearfirst".  I may follow-up by extending the test.  For now let's just get this bugfix in.

pganssle:
I think the appveyor build failure might have been a temporary failure on the AppVeyor side, kicked off a new build.

dateutil/test/test_parser.py
```diff
@@ -11,6 +12,36 @@
...
+class TestFormat(unittest.TestCase):
+
+    def test_ybd(self):
+        # If we have a 4-digit year, a non-numeric month (abbreviated or not),
+        # and a day (1 or 2 digits), then there is no ambiguity as to which
+        # token is a year/month/day.  This holds regardless of what order the
+        # terms are in and for each of the separators below.
+
+        seps = ['-', ' ', '/', '.']
+        token_opts = [['%Y'], ['%b', '%B'], ['%d', '%-d']]
```
pganssle: I think `%-d` might not be supported on Windows.

pganssle:
[strftime.org](strftime.org) says that `%-d` is platform specific, and I think that Windows doesn't support it, which was causing the AppVeyor failures that were happening *after* I fixed the unrelated TLS issue.

I've updated it now so that '%-d' tests are only run if the platform supports them, and your innermost `for` loop was at the wrong level (though this made no difference).

I removed the nested for loops for year, month and day because they were not necessary (we'll get back to that when we're using pytest), and also dropped the `default` spec, because I think it's actually dangerous to make the correct answer the default value if nothing is found for a given token.

jbrockmendel:
> which was causing the AppVeyor failures that were happening after I fixed the unrelated TLS issue.

Nice catch!

PR #483, opened 2017-10-27
add day check; with test; closes #360

### LINKED ISSUE #360: Parser fails to parse 'DDmonYYYY'
```python
>>> dateutil.parser.parse('13NOV2017', fuzzy=True)
datetime.datetime(2017, 4, 13, 0, 0)  # --- today's date

>>> dateutil.parser.parse('13NOV2017', fuzzy=True, dayfirst=True)
datetime.datetime(2017, 4, 13, 0, 0)  # --- today's date
```
Without fuzzy, it raises an error.

These should parse to the proper dates. I am not sure how the tokenization is done, but a possible solution could be to use an alphabet character (in addition to what I think is used - any non-alphanum character)

jbrockmendel:
Will evidently need rebasing after review.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -398,6 +398,17 @@ class _ymd(list):
     def has_day(self):
         return self.dstridx is not None
 
+    def could_be_day(self, value):
+        if self.has_day:
+            return False
+        elif not self.has_month:
+            return 1 <= value <= 31
+        elif not self.has_year:
+            # Be permissive, assume leapyear
+            return 1 <= value <= monthrange(2000, self.month)[1]
+        else:
+            return 1 <= value <= monthrange(self.year, self.month)[1]
+
     @staticmethod
     def token_could_be_year(token, year):
         try:
@@ -715,8 +726,8 @@ class parser(object):
             while i < len_l:
 
                 # Check if it's a number
+                value_repr = l[i]
                 try:
-                    value_repr = l[i]
                     value = float(value_repr)
                 except ValueError:
                     value = None
@@ -830,11 +841,15 @@ class parser(object):
                             ymd.append(value)
                         i += 1
 
+
                     elif info.ampm(l[i + 1]) is not None:
                         # 12am
                         hour = int(value)
                         res.hour = self._adjust_ampm(hour, info.ampm(l[i + 1]))
                         i += 1
+
+                    elif ymd.could_be_day(value):
+                        ymd.append(value)
 
                     elif not fuzzy:
                         raise InvalidDatetimeError(timestr)
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -405,9 +405,12 @@ class _ymd(list):
             return 1 <= value <= 31
         elif not self.has_year:
             # Be permissive, assume leapyear
-            return 1 <= value <= monthrange(2000, self.month)[1]
+            month = self[self.mstridx]
+            return 1 <= value <= monthrange(2000, month)[1]
         else:
-            return 1 <= value <= monthrange(self.year, self.month)[1]
+            month = self[self.mstridx]
+            year = self[self.ystridx]
+            return 1 <= value <= monthrange(year, month)[1]
 
     @staticmethod
     def token_could_be_year(token, year):
@@ -840,7 +843,6 @@ class parser(object):
                             # Year, month or day
                             ymd.append(value)
                         i += 1
-
 
                     elif info.ampm(l[i + 1]) is not None:
                         # 12am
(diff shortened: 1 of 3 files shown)


pganssle:
LGTM.

dateutil/parser.py
```diff
@@ -398,6 +398,17 @@ def has_month(self):
...
         return self.dstridx is not None
 
+    def could_be_day(self, value):
+        if self.has_day:
+            return False
+        elif not self.has_month:
+            return 1 <= value <= 31
+        elif not self.has_year:
+            # Be permissive, assume leapyear
+            return 1 <= value <= monthrange(2000, self.month)[1]
```
pganssle: Weird choice of an exemplary leap year, since this is one of the weird leap years.

Is it meaningful to add a test that will hit this branch for leap and non-leap years?

pganssle:
Hm.. I'm noticing there's not a codecov hook anymore, but actually this seems to be introducing some uncovered lines [see here](https://codecov.io/gh/dateutil/dateutil/pull/483/diff#D1-403). @jbrockmendel Do you think you can come up with some strings that will hit those other cases?

dateutil/parser.py
jbrockmendel: Fair point.  I'm happy to use 1996.

> Is it meaningful to add a test that will hit this branch for leap and non-leap years?

My expectation is that this will turn out to be fragile in ways unrelated to leap vs non-leap.  My preference would be to keep an eye out for from-the-wild cases that hit these branches rather than try to concoct them for the sake of coverage stats.

jbrockmendel: That said, I do have a couple of PRs coming up that are just adding tests, many of which will currently be `xfail`s.

pganssle: I don't think  1996 vs 2000 matters. Just gave me a start because I was like, "Wait, that's not supposed to be a leap year because it's a century, actually, wait, the century is divisible by 4..." Doesn't matter to a computer and the comment is necessary to make it clear why you chose 2000 anyway.

Regarding the tests, I think it might be a good idea to come up with some tests that will hit those branches. I can do it if you didn't have anything in mind.

jbrockmendel: Go for it.  Anything you don't get covered I'll circle back to get next week.

jbrockmendel:
I haven't brainstormed any particularly good examples for the other cases.  Got any in mind?

pganssle:
I couldn't think of anything so I tried fuzzing the parser a bit with `python-afl`, and in fact found a bug in those lines!

```python
>>> parse('02:17NOV2017BCE')
...
AttributeError: '_ymd' has no attribute 'month'
```

jbrockmendel:
Good catch.  I'll have to take a look at that fuzzer.

Pushing a fix shortly.

Though the trailing "BCE" is a whole other can of worms.

pganssle:
It also works with a trailing `CE`, which *would* be a valid date in Python. I haven't found any examples yet where it would actually parse to a real date, though.

jbrockmendel:
Yah, I'm running the fuzzer now to get a feel for how it works.

I vote for closing out this bug fix and not cargo-culting coverage.

pganssle:
@jbrockmendel I wouldn't worry too much about the fuzzing, but I think that my insistence on coverage was not at all cargo culting, since in this case the code you've written is fundamentally broken. Anything that reaches the uncovered branches will fail with `AttributeError`. I'm not sure if there's an obvious fix to it, since `_ymd` no longer actually stores year, month or day, so either the approach to `could_be_day` needs to be fixed or `_ymd` needs to take a different approach as a container.

Worst case scenario, we can add tests that directly call `_ymd` (and maybe that's a good start when doing the development), but I'd be more comfortable with that if we could come up with some reasoning to explain why no valid datetime strings would ever reach those branches in the current implementation.

jbrockmendel:
>  since in this case the code you've written is fundamentally broken

Fair point.  (Huh.  The fix I pushed isn't showing up here... gotta see where that ended up)

> [...] if we could come up with some reasoning to explain why no valid datetime strings would ever reach those branches in the current implementation.

I have no such reasoning, and I like the idea of using fuzzing to try to find them.  All I'm getting at is that if we can't find such cases after [whatever the appropriate amount of effort is] we shouldn't let that stop us from fixing the bug we know about.

pganssle:
I like this approach with `test_internals.py`. Fuzzing with 5 parallel tasks for most of a day didn't find anything useful here, so we'll replace the relevant `test_internals.py` calls if we come up with any thing that naturally hits these lines (I'm guessing that's more likely to happen due to refactoring than fuzzing or something).

Issue #487, opened 2017-10-30
datetime exception while parsing unambiguous string


WeatherGod:
```
>>> date_str
'1237 PM EDT Mon Oct 30 2017'
>>> dateutil.parser.parse(date_str)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/dateutil/parser.py", line 1182, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/dateutil/parser.py", line 581, in parse
    ret = default.replace(**repl)
ValueError: hour must be in 0..23
```

As a side-note, is there ever going to be an option for end-users to specify a format string for parsing? In this case, `dateutil` is really the only good option for easily handling arbitrary timezones (I don't know ahead of time which timezone a string will have), but I have always been antsy about the lack of control over the parser, because in this case, I know that my date string will always follow this particular form.

pganssle:
This one is a bit weird partially because apparently it's choosing the wrong thing for the hour, but I think at the moment it would be quite hard to support this kind of string, mostly because we don't necessarily have a way to distinguish between `1237` and `2017` as the year (though in this case we *could* theoretically distinguish it because October 30, 1237 was a Thursday).

In this case, since you know the format, I recommend doing something like this:

```python
from datetime import datetime
from dateutil import tz

def parse_specific_string(dtstr):
    dt = datetime.strptime(dtstr, '%H%M %p %Z %a %b %d %Y')
    tzname = dtstr[8:].split(' ', 1)[0]
    if tzname:
        tzi = tz.gettz(tzname)
        dt = dt.replace(tzinfo=tzi)

    return dt
```

I would then take the person who decided to use `'%H%M %p %Z %a %b %d %Y'` as a string format and give them a stern talking-to. I think `%Z` may only be "supported" for Python 3, and even then it really just ignores that section of the string. No matter how you slice it, though, it's risky business using the 3-letter time zone name abbreviations to actually get a time zone name.

jbrockmendel:
This is a class of cases that I intend to fix, but its not high on the priority list.  As @pganssle mentioned, its a re-write as opposed to a bug-fix.

One way you could help keep this from drifting down the todo list would be to submit a PR creating a test for this case marked as `xfail`.

pganssle:
[...]

Issue #394, opened 2017-06-04
New test cases that will currently fail


jbrockmendel:
I have a handful of cases found in the wild that are not currently supported, or in some cases give incorrect results.  In each case I have a fix ready, but some of them required non-trivial refactoring.  I'll make PRs for any that you want to support.

```
dstr = "File Creation Time: 0513201421:33"
parse(dstr, fuzzy=True)
# desired: datetime(2014, 5, 13, 21, 33)
# actual: ValueError: hour must be in 0..23

dstr = "1991041310:19:24"
parse(dstr, fuzzy=True)
# desired: datetime(1991, 4, 13, 10, 19, 24)
# actual: ValueError: hour must be in 0..23

# Infer day/year from the trailing "th"
dstr = "06 June 11th"
parse(dstr, fuzzy=True)
# desired: datetime(2006, 6, 11)
# actual: datetime.datetime(2011, 6, 6, 0, 0)

# The "94" is getting identified as a year at the beginning, then
# it balks at the "1994"
dstr = "SMITH R &  WEISS D 94 CHILD TR FBO M W SMITH UDT 12/1/1994"
parse(dstr, fuzzy=True)
# desired: datetime(1994, 12, 1)
# actual: ValueError: Unknown string format


# The 2011 is identified as the year, and so the "2012" is slotted into
# the un-filled hour/minute
dstr = "2011 MARTIN CHILDREN'S IRREVOCABLE TRUST u/a/d NOVEMBER 7, 2012"
parse(dstr, fuzzy=True)
# desired: datetime(2012, 11, 7)
# actual: datetime(2011, 11, 7, 20, 12)
```

pganssle:
This one is certainly unambiguous and should probably not have the `fuzzy` keyword:

'''python
# Infer day/year from the trailing "th"
dstr = "06 June 11th"
parse(dstr, fuzzy=True)
# desired: datetime(2006, 6, 11)
# actual: datetime.datetime(2011, 6, 6, 0, 0)
```

The others I think can be supported at some point, but I'd like to put it off until after we've got a new model for how to customize how the parser works. This week some point I'm going to put together a general document and open it for comments on how to do that, but I think that it's not a good idea to paint ourselves into a corner by introducing new improvements that we'll be implicitly promising to support going forward.

PR #488, opened 2017-10-31
Adding an expected fail test case for the parser.

WeatherGod:
* Issue reported in #487

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -17,6 +17,7 @@ try:
 except ValueError:
     PLATFORM_HAS_DASH_D = False
 
+import pytest
 
 class TestFormat(unittest.TestCase):
 
@@ -890,7 +891,7 @@ class ParserTest(unittest.TestCase):
         with self.assertRaises(ValueError):
             parse(invalid)
 
-    @unittest.skip("Known failure")
+    @pytest.mark.xfail()
     def test_somewhat_ambiguous_string(self):
         # Ref: github issue #487
         # The parser is choosing the wrong part for hour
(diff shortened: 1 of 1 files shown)


pganssle:
Hm.. I've avoided "known failures" in the test suite for a while because I feel like it's a bit weird to test *that* something fails unless the failure is a deliberate behavior. Thinking about it now, I guess it's a useful sentinel to let you know when you've accidentally fixed an issue. I guess when we switch to pytest we can change these to explicit `xfail`s.

@jbrockmendel Since you advocated this, can you point me at anything that goes into the pros and cons of using `xfail`?

jbrockmendel:
>  Since you advocated this, can you point me at anything that goes into the pros and cons of using xfail?

I think of it as a TODO list, kind of a TDD thing.  It helps us understand where the parser's weaknesses are.  It's something we can point new contributors to for good first-PRs.  And it's a natural place to park stuff like #394.

dateutil/test/test_parser.py
```diff
@@ -877,3 +877,14 @@ def test_validate_hour(self):
         invalid = "201A-01-01T23:58:39.239769+03:00"
         with self.assertRaises(ValueError):
             parse(invalid)
+
+    @unittest.skip("Known failure")
```
jbrockmendel: For the purposes of treating this like a TODO list, I'll follow up to put these into a separate class.

pganssle: Once we have `pytest`, we should refactor these as `@pytest.mark.xfail`, since that acts as a sentinel for any PRs that might accidentally fix it in the future.

Issue #424, opened 2017-07-21
Add isoparse() function.


pganssle:
There is currently no inverse to the `isoformat()` function. If you *know* you've got an iso8601 date, the logic is actually *much* simpler than the current parser logic, but because `isoformat()` outputs a format not compatible with even Python 3.6's `strptime()` (shamefully), it's some pretty low hanging fruit.

jbrockmendel:
One solution here would be to adapt the numpy/pandas iso8601 parser.  I think there are a couple corner cases we would want to fix up, but it's generally in pretty good shape.

The down-side is that it is in a C extension.  My impression is that for this format in particular, people are going to want that performance bump.  But dateutil being pure-python is really pleasant, so I'm ambivalent on this one.

pganssle:
@jbrockmendel Yeah, I've been very hesitant to do anything with C extensions because it can add deployment headaches downstream. I have an idea of how to do the iso8601 parser, that I believe will be quite fast (for pure python, anyway). I believe that most if not all of the iso8601 formats have a fixed length, so it may be possible to just look up the format string by the length of the input, with some corner cases handled.

wagner-certat:
A c extension could be optional, so downstream can enable/use it voluntarily.

pganssle:
Done in #489.

Issue #449, opened 2017-08-16
Refactoring "partialdatetime" out of relativedelta


pganssle:
There are a number of reasons why it might make sense to separate the absolute and relative components of `relativedelta.relativedelta`:

1. People are often confused by the fact that the arguments have such similar names - `month` vs. `months` - easy to miss in code review, etc.
2. The absolute arguments make it way less meaningful to compare these things, and the differences around which components change if you add/subtract things are not intuitive.
3. There is a meaningful abstraction for the absolute components - a "partial date" - e.g. "month=2, day=1".

I suggest an implementation where all the absolute components are refactored out of `relativedelta` and into a `partialdatetime` object - which itself would be composed of a `partialdate` and `partialtime` object. These would have the following properties:

* They can be fully empty, and empty `partial*` objects are falsy.
* They can be used to update existing dates - the interface for this is unclear, but we'll want a simple method for using them to update `datetime objects, some options:
    - Implement them as a subclass of something that can be expanded with double-stars (`datetime.replace(**p_dt)`)
    - A method that returns a `dict` (`datetime.replace(**p_dt.asdict())`)
    - A method on `partialdate` that updates the `datetime` (`partialdate.update(dt)`) (not great)
    - Override `__add__` to mimic the existing `relativedelta` interface with `datetime` (also not great)

The following are still open questions:
* Mutability - should they be immutable like `datetime`, or mutable? I'm leaning towards immutable.
* Binary operations - how should they be defined? What is `partialdate(year=2017, month=2) < partialdate(year=2017, day=14)`? How about `partialdate(year=2017, month=2) - partialdate(year=2016)`?

I'd say that they should be added as a keyword argument and main implementation for `relativedelta` relatively soon, and then the absolute arguments should be deprecated in the following release. Ideally we'd factor out the non-absolute components into a separate object at the same time, but I think the correct name for that object is `relativedelta`, so I'm not sure what we would want to call the super-object in that case.

jbrockmendel:
I'll look at this more closely later, but want to give a big +1 for immutability.

jbrockmendel:
`partialdatetime` seems less `timedelta`-like than `datetime`-like.  Am I getting that wrong?

We've discussed the possibility of having `parser.parse` return something like a `partialdatetime`.  Would you want to use the same class for both features?

[...]

PR #489, opened 2017-11-01
Implement dedicated isoparser

pganssle:
Here's a first pass at a dedicated iso parser per #424.

This involves a reorganization of the code for `parser` into its own little submodule, to avoid the explosion of `parser.py`. I think there will likely be some problems for downstream users who are using private interfaces, but I'm not too worried about that.

I've got a rough sketch for a Cython version of this that is ~20x faster, but it seems like even this pure-python version beats a `datetime.strptime` call:

```python
In [1]: from dateutil.parser import isoparse
   ...: from datetime import datetime
   ...: 
   ...: dt = '2014-02-28T14:25:03.446990'
   ...: 

In [2]: %timeit isoparse(dt)
10000 loops, best of 3: 20.7 µs per loop

In [3]: %timeit datetime.strptime(dt, '%Y-%m-%dT%H:%M:%S.%f')
The slowest run took 142.82 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 34 µs per loop
```

I may have gone a bit overboard with the test parametrization. More tests need to be added as well - I'm not, for example, testing any failure conditions.

**Remaining items to do**:
- [x] Explicitly convert to ASCII bytestring (there is no unicode in valid ISO-8601, and this assumption will make the Cython implementation faster)
- [x] Add tests for the `parse_isodate` and `parse_isotime` functions.
- [x] Decide what to do about the mix of classmethods and methods.

Deferred:
- Clean up documentation so that the primary interface is more explicitly `isoparse`.

Ping @jbrockmendel @wagner-certat @mariocj89


jbrockmendel:
I'll go over this sometime in the next couple days.

If we go down this path (which I'm generally positive on), we should consider adapting the [numpy](https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/datetime_strings.)/[pandas](https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/src/datetime/np_datetime_strings.c) implementation.

.travis.yml
```diff
@@ -21,13 +21,14 @@ before_install:
...
   - pip install freezegun
+  - pip install pytest
 
 install:
   - pip install six
   - ./ci_tools/retry.sh python updatezinfo.py
 
 script:
-  - coverage run --omit=setup.py,dateutil/test/* setup.py test
+  - coverage run --omit=setup.py,dateutil/test/* -m pytest
```
jbrockmendel: Let's separate this out, as it has caused heisenbugs in the past.

pganssle: Separate it out how?

pganssle:
> If we go down this path (which I'm generally positive on), we should consider adapting the numpy/pandas implementation.

I don't think this would be particularly useful. It's hard to directly compare what they are doing with what we are doing, but I don't think it would be terribly difficult to get something as fast as they have if we drop down to C, *and* they don't actually fully support ISO strings. Consider the following edge cases:

```python
import numpy as np

iso_strings = ['2014-W03', '2014-W03-3',   # ISO week
               '2014-224',                 # Ordinal day
               '2014-01-01T00-12:15',      # Timezone aware
               '20140301T24:00'            # Midnight as 24:00
               ]

for iso_string in iso_strings:
    try:
        np.datetime64(iso_string)
        print('Sucess: {}'.format(iso_string))
    except ValueError:
        pass
```

The only one that parses correctly is the timezone aware one, and that's apparently deprecated:

```
/home/c0zj/anaconda3/bin/ipython:11: DeprecationWarning: parsing timezone aware datetimes is deprecated; this will raise an error in the future
Sucess: 2014-01-01T00-12:15
```

I don't think it should be terribly difficult to eventually add a C implementation here that's quite fast, but even the pure python version I've got so far (with no profiling to look for bottlenecks, mind you), is already faster than `strftime`, which is saying something.

.travis.yml
pganssle: Also, if by "heisenbugs" you mean that issue with your other PR, I think that *may* have been to do with the `assert` method of creating tests.

pganssle:
The items that will be left for a separate PR:

1. Add `hypothesis` tests here (should be easy to test the hypothesis that dates, times and datetimes should be round-trippable with `.isoformat()`)
2. Add a fast C version for `isoparse`
3. Add support for offsets and periods.

.travis.yml
jbrockmendel: > Separate it out how?

I'm suggesting the move to pytest should be done in an independent PR (I think you may have already updated this below)  

> may have been to do with the yield method of creating tests

Good to know, thanks.

jbrockmendel:
> I don't think this would be particularly useful. It's hard to directly compare what they are doing with what we are doing, but I don't think it would be terribly difficult to get something as fast as they have if we drop down to C, and they don't actually fully support ISO strings

Fair enough.  I mention it because I've been spending a lot of time recently working on pandas._libs.tslibs so am pretty familiar with it.  If we implement something strictly better so that we can take that off their plate, that would be a big win.  Also note that pandas altered the numpy version to support tz-aware returns.

jbrockmendel:
> Add support for offsets and periods.

Can you clarify what you mean by offsets and periods? Again my train of thought goes to `pd.offsets` and `pd.Period`.

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,384 @@
...
+from datetime import datetime, timedelta, time, date
+import calendar
+from dateutil import tz
+
+import re
+
+__all__ = ["isoparse", "Isoparser"]
+
+
+class Isoparser(object):
```
jbrockmendel: Is this something you expect users to subclass regularly?

dateutil/parser/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+from ._parser import parse, parser, parserinfo
+from ._parser import DEFAULTPARSER, DEFAULTTZPARSER
+from .isoparser import Isoparser, isoparse
+
+__all__ = ['parse', 'parser', 'parserinfo',
+           'isoparse', 'Isoparser',
+           'InvalidDatetimeError', 'InvalidDateError', 'InvalidTimeError']
```
mariocj89: You need to bring this to this namespace as well, don't you? It is neither a module nor a local variable.

I think doing this will break `from dateutil.parser import *`

mariocj89: You might want to add some tests to verify current behaviour on the ` import * ` and general imports before moving it :X

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,384 @@
...
+    def __init__(self, sep='T', default_year=None):
+        """
+        :param sep:
+            A single character that separates date and time portions
+
+        :param default_year:
+            The default year to be used as the basis for parsing the uncommon
```
mariocj89: You might want to document the default is the current year

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,384 @@
...
+            no-year date formats.
+        """
+        if (len(sep.encode('utf-8')) != 1 or
```
mariocj89: You might not need to do encode at all.

You can just check the length, and if you are worried on unicode characters you are already using `ord` that will give you the unicode point.

```
>>> ord(u'语')
35821
```

Raising it as in python2 you will be calling encode in bytestrings (should not cause any issue unless using unicode characters, but quite weird)

.travis.yml
pganssle: OK, yeah, maybe I'll cherry-pick the pytest commit and rebase this on master after we do that.

dateutil/parser/__init__.py
pganssle: Good catch! I'll add that and add some tests.

dateutil/parser/isoparser.py
pganssle: Good point. I put the default year documentation in `isoparse`, but I should also have it here.

dateutil/parser/isoparser.py
pganssle: This is kinda one of the open questions, honestly. I've been a bit haphazard about the design of this class because I keep vacillating between "users will create an instance" and "users will create a subclass", which is why half the methods are `classmethod` and half are instance methods.

I think I'm partially being biased by the parser, where my hope was that downstream users would subclass `parser` to allow customization of behavior, but even in the parser, I'm thinking that might not be the right design for this particular sort of task.

After giving it a bit of thought, I believe that since ISO-8601 has a very clear set of options it supports, it might be best if it we refactored this to be more like `parser`/`parserinfo`, where `isoparserinfo` would have some limited feature flags like `ordinal_dates` and `weeks` (so that there's not this artificial division between "common" and "uncommon").

That said, in terms of subclasses, I think that when #449 is done, I'll probably want an internal subclass that returns a `PartialDatetime` instead of `datetime` (that way the `default_year` logic will be unnecessary).

dateutil/parser/isoparser.py
pganssle: Ah, good call. I added the `ord` check after the `encode` portion and didn't realize it obviated the need to encode.

dateutil/parser/isoparser.py
jbrockmendel: > [...] but even in the parser, I'm thinking that might not be the right design for this particular sort of task.

Yah, I've gone along with the make-everything-a-method approach since there isn't a _ton_ of downside, but I think YAGNI probably applies unless we have good reason to believe users actually are subclassing.

(This may be a discussion that belongs elsewhere, don't want to derail this PR's discussion)

Especially for e.g. `parser._build_tzinfo`, I have a hard time imagining the use case where it is easier to subclass than it is to pass a callable for `tzinfos`.  So I might start pushing to be more selective about does-this-really-need-to-be-a-method.

.travis.yml
jbrockmendel: If you're doing cherry-picking for small fairly-immediate changes, making `parser` a directory is worthwhile.  It'll make for much smaller diffs in other PRs.

dateutil/parser/isoparser.py
jbrockmendel: > In any case, that doesn't mean I want bare functions sitting around.

OK, I evidently mis-interpreted your previous comment.  Never mind.

.travis.yml
pganssle: Hm.. Probably a good idea, but I think I'll just wrap up the feature as is and do one big rebase at the end.

pganssle:
@jbrockmendel Yes. Technically this code already has one - if you parse with `common=True`, you won't hit the `YYYY-?Www-?D?`, `--MM-DD` or `YYYY-DDD` branches. I'm thinking that "common" and "uncommon" are not descriptive or fine-grained enough, though, so I was thinking of introducing an `isoparserinfo` class containing feature flags, with some pre-configured constants for the equivalent of `common` and `uncommon`.

I think the final interface will be partially driven by how easy it is to make these feature flags performant in both Python and Cython (as I intend for `isoparser` to be the first module with an API-compatible C, Cython and/or Rust backend).

jbrockmendel:
That seems like a reasonable approach.

>  (as I intend for isoparser to be the first module with an API-compatible C, Cython and/or Rust backend).

I think this may also bear on my feeling of rushedness mentioned over in the pandas thread.  This represents a pretty big shift in terms of build/distribution.

That said, the rust bit in particular would be very neat.

pganssle:
@jbrockmendel To be very clear, I am *not* planning on releasing a compiled backend as part of the 2.7.0 release, but I have been using this as a test module for doing that. I've implemented parts of the algorithm in Rust and in Cython for some basic profiling and I'm using those results to try to guide what the public part of the API should look like in Python.

I foresee a compiled backend release happening in a few stages, either I'll start with a separate `cdateutil` package or some sort of build / install options where the compiled backend is *entirely* optional. The end goal in the near term would be to have one pure python release and then pre-build binaries for a number of architectures, that way anyone on an unsupported architecture could always fall back to the pure python version.

With regards to the use of Rust, I haven't got the Rust part binding to Python quite yet, but I'm not optimistic about its performance, at least partially because I don't see an obvious way to have Rust call the C `datetime` constructor (there's one version that takes `PyObject`s, and then another version that takes a bunch of `int`s - the second version is *much* faster than the first), and even without that it's not considerably faster than my lightly-optimized Cython implementation.

Rust also has a few other strikes against it. First, it will be harder to accept PRs that would have an effect in the Rust code, since contributors will likely find it harder to work with than Cython or C. Second, there's not nearly the same support for Rust<->Python bindings that there is for C<->Python bindings. Third (and this may be similar to the second reason), building this thing cross-platform is going to be hard enough, adding in a dependency on rust/cargo might make things even more complicated.

pganssle:
From profiling, it seems like one of the biggest slowdowns in parsing time zone aware zones is constructing `tzoffset`. I think making that a singleton will improve that markedly for all real world scenarios.

pganssle:
For the last bullet point, I think I'm going to drop all class methods and go with all instance methods, then add aliases for `parse_isodate`, `parse_isotime` and `parse_isotzstr` in the main namespace, to start with.

Unfortunately, there's still one big work item, which is that the `isoparse_common` and `isoparse_uncommon` distinction is weird and unpleasant to me, so I need to refactor that with feature flags.

That said, this is becoming a pretty long-lived feature branch. I'm thinking maybe in this branch we'll drop the `common`/`uncommon` distinction entirely (except as an implementation detail) so that we can go ahead and merge a minimal-features version of this, then add a separate branch for the feature flag refactoring.

jbrockmendel:
I said several days ago I would review this promptly and then didn't.  Double-plus will do so today.

dateutil/parser/__init__.py
```diff
@@ -3,5 +3,8 @@
 from ._parser import DEFAULTPARSER, DEFAULTTZPARSER
 from ._parser import InvalidDateError, InvalidDatetimeError, InvalidTimeError
 
+from .isoparser import Isoparser, isoparse
+
 __all__ = ['parse', 'parser', 'parserinfo',
+           'isoparse', 'Isoparser',
```
jbrockmendel: ISOParser?

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+
+        :param zero_as_utc:
+            Whether to return :class:`dateutil.tz.tzutc` for zero-offset zones
+
+        :return:
+            Returns :class:`dateutil.tz.tzoffset` for offsets and
+            :class:`dateutil.tz.tzutc` for ``Z`` and (if ``zero_as_utc`` is
+            specified) offsets equivalent to UTC.
+        """
+        return cls._parse_tzstr(tzstr, zero_as_utc=zero_as_utc)
```
jbrockmendel: This strikes me as redundant.  What's the benefit of having both of these?

dateutil/parser/__init__.py
pganssle: I looked into this particularly for the question of how to handle CapWords for TZ and it seems like the consensus is that you don't capitalize all the letters of acronyms, just treat them as any other word (so `IsoParser` instead of `ISOParser`, `TzUtc` instead of `TZUTC`).

That said, despite the fact that Python is pretty weirdly inconsistent on this point, when I actually read PEP 8 recently, they pretty strongly discourage changing an established library style just to be consistent with PEP 8, so I'm thinking we should just embrace `isoparser`.

dateutil/parser/isoparser.py
pganssle: `parse_tzstr` is a public function, `_parse_tzstr` is the private implementation.

In this case, it's useful because `parse_tzstr` has the `@_takes_ascii` decorator and `_parse_tzstr` doesn't (because it is only called from other functions that already have the `@_takes_ascii` decorator, and as such it prevents an unnecessary call).

jbrockmendel:
My main thought so far:

I know you like using methods where functions might otherwise work.  In most cases I cringe a little but recognize it doesn't really hurt anything.  This is a case where I'll ask you to consider using functions where possible.

There is eventually going to be a C/Cython version of this.  Because perf is critical here, I'm guessing we'll end up going full-on C.  Once there are two versions, we'll need to be super-careful to keep them in sync.  Two things we can do to keep them in sync are 1) fine-grained unit tests to ensure their behaviors are the same and 2) keep the code structure parallel in each version.  Using functions instead of methods makes this _much_ easier, especially in the raw-C case.

Why not take the private versions of these methods and turn them into functions?  The instance/class attributes that are accessed can be passed as args.

pganssle:
> I know you like using methods where functions might otherwise work. In most cases I cringe a little but recognize it doesn't really hurt anything. This is a case where I'll ask you to consider using functions where possible.

I don't know why you cringe at this, having a mix of instance methods and bare functions useful *only* for a given class breaks encapsulation. If it were a single function or a bunch of functions that were useful for other things, then I'd be more willing to use bare functions.

There is some case to be made that `Isoparser` is too big, and that it should be `IsoDateParser`, `IsoTimeParser` and `IsoTzParser`, with an `IsoParser` that contains one of each. I think there's basically no shared logic between the three.

The one method that I think may be useful to throw in `dateutil.utils` would be `Isoparser._calculate_weekdate`. The main reason I'm not doing that is that I'm not entirely sure what the right scope is for `dateutil.utils`. It's clearly a "miscellaneous bare functions" category, I'm wondering if it would be best broken down into smaller categories. I think a function that is the inverse of `date.isocalendar()` is useful outside of its utility in parsing ISO datetimes.

> There is eventually going to be a C/Cython version of this. Because perf is critical here, I'm guessing we'll end up going full-on C. Once there are two versions, we'll need to be super-careful to keep them in sync. Two things we can do to keep them in sync are 1) fine-grained unit tests to ensure their behaviors are the same and 2) keep the code structure parallel in each version. Using functions instead of methods makes this much easier, especially in the raw-C case.

In fact, I think one of the best reasons for encapsulation is that it makes it much easier to handle drop-in and partial replacement functions. Depending on exactly how we end up doing it, we can either have a `cdef` Cython class that calls either C or Cython `cdef` methods and is a drop-in replacement for `Isoparser`, or possibly we can have a `CIsoparser` subclass of `Isoparser` where one or more of the speed-critical functions has been overridden to call `C` functions, but the remainder of the logic is shared. This would be much more difficult if we were using a bunch of bare functions.

Likely we'll start out with `isoparse` being the only part of the official public interface, or possibly `isoparse` plus some bare functions for parsing dates-only, times-only and tz strings only. The fact that they are methods on some constant instance will be considered an implementation detail.

jbrockmendel:
You make some fair points.  I'll give your comments some thought and revisit tomorrow(ish).

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+        elif tzstr[0:1] == b'+':
+            mult = 1
+        else:
+            raise ValueError('Time zone offset requires sign')
+
+        hours = int(tzstr[1:3])
+        if len(tzstr) == 3:
+            minutes = 0
+        else:
+            minutes = int(tzstr[(4 if tzstr[3:4] == cls._TIME_SEP else 3):])
```
jbrockmendel: By this point you know that `len(tzstr)>3` so the `tzstr[3:4]` trick is unnecessary, can just do `tzstr[3]`.

The checks below are probably sufficient, but I would advocate explicitly checking here

```
elif len(tzstr) == 5:
    if not tzstr[-2:].isdigit(): --> raise
elif len(tzstr) == 6:
    if not tzstr[-2:].isdigit() or tzstr[3] != cls._TIME_SEP --> raise
```

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+        week_1 = jan_4 - timedelta(days=jan_4.isocalendar()[2] - 1)
+
+        # Now add the specific number of weeks and days to get what we want
+        week_offset = (week - 1) * 7 + (day - 1)
+        return week_1 + timedelta(days=week_offset)
+
+    @classmethod
+    def _parse_isotime(cls, timestr):
+        len_str = len(timestr)
+        components = [0, 0, 0, 0, None]
```
jbrockmendel: I'd advocate using named variables and putting them all into a tuple at the end

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+
+            if comp < 3:
+                # Hour, minute, second
+                components[comp] = int(timestr[pos:pos + 2])
+                pos += 2
+                if (has_sep and pos < len_str and
+                        timestr[pos:pos + 1] == cls._TIME_SEP):
+                    pos += 1
+
+            if comp == 3:
```
jbrockmendel: elif

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+        return components, pos + 2
+
+    def _parse_isodate_uncommon(self, dt_str):
+        if len(dt_str) < 4:
+            raise ValueError('ISO string too short')
+
+        if dt_str[0:2] == b'--':
+            # --MM-DD or --MMDD
+            month = int(dt_str[2:4])
+            pos = 4 + (dt_str[4:5] == self._DATE_SEP)
```
jbrockmendel: Why is `_DATE_SEP` a instance attribute while `_TIME_SEP` is a class attribute?

dateutil/parser/isoparser.py
pganssle: That's actually not why I'm doing `tzstr[3:4]`. In Python 3, the somewhat foolish decision was made that `b'x'[0] != b'x'`. The result if indexing `bytes` is an integer for some insane reason, not a single-character `bytes` or some other thing. If you take a single-character slice, it gives you `b'x'` in both Python 2 and Python 3.

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+            day = int(dt_str[pos:pos + 2])
+            year = self._default_year
+
+            if month == 2 and day == 29:
+                # Calcualtes the latest leap year
```
jbrockmendel: Calcualtes --> Calculates

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
...
+        if len_str - pos < 2:
+            raise ValueError('Invalid common day')
+        components[2] = int(dt_str[pos:pos + 2])
+        return components, pos + 2
+
+    def _parse_isodate_uncommon(self, dt_str):
+        if len(dt_str) < 4:
+            raise ValueError('ISO string too short')
+
+        if dt_str[0:2] == b'--':
```
jbrockmendel: Maybe break this condition, the one on line 272, and the one on 290 into three small methods that each handle one format?  That would leave flexibility to tinker with what constitutes "common".

dateutil/parser/isoparser.py
pganssle: That was my original thought, but it makes it much more complicated to exit early from the parsing. The three options you have are: 

1. have a `components` list which you index in a `while` loop that you can `break` from (what I've done here)
2. take a mutable `components` object (basically a list again) as an input and `return` when you're done parsing the string (something I'm strongly considering doing, actually, for speed reasons)
3. use some other complicated method like increasingly nested values.

I think the first one is relatively readable and performant, but came up with the second one while working on the Rust implementation, so I thought I'd try that out and see if it looks any better.

dateutil/parser/isoparser.py
pganssle: Yes, that's the plan for the feature flags.

dateutil/parser/isoparser.py
pganssle: They're technically both class attributes, but `_parse_isodate` is an instance method and `_parse_isotime` is a class method because I started with most things as class methods and only moved to instance methods as necessary. I think I'll move everything to instance method or class methods in the end.

The reason why dates need to be instance methods is that I've made `default_year` an instance attribute.

dateutil/parser/isoparser.py
pganssle: Fair.

dateutil/test/test_isoparser.py
```diff
@@ -0,0 +1,489 @@
...
+
+    return list(it.chain.from_iterable(outputs))
+
+
+@pytest.mark.parametrize('time_val,time_fmt', __make_time_examples())
+@pytest.mark.parametrize('as_bytes', [True, False])
+def test_isotime(time_val, time_fmt, as_bytes):
+    tstr = time_val.strftime(time_fmt)
+    if isinstance(time_val, six.text_type) and as_bytes:
+        tstr = tstr.encode('ascii')
```
pganssle: It's a bit concerning that this line is never hit.

dateutil/parser/isoparser.py
jbrockmendel: The `_takes_ascii` point is compelling; the private vs public doesn't seem like a meaningful distinction in this case.

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,424 @@
+# -*- coding: utf-8 -*-
```
jbrockmendel: I haven't looked over the tests file, but this file looks generally OK to me.

dateutil/parser/isoparser.py
pganssle: Private means it's an implementation detail. You can't count on it existing, but the current implementation requires it. If we later find a different way to do this, the only thing we're guaranteeing is that the public interface has a `parse_tzstr` function that takes something ascii-like.

jbrockmendel:
> having a mix of instance methods and bare functions useful only for a given class breaks encapsulation

I'm thinking of the level of encapsulation as being the module.  Especially if its behind a `dateutil._parser` layer.

Here's a thought: my cringing is related to implicit statefulness.  A classmethod or an instance method carries the risk that a function/method's implementation may not be fully isolated (which I like particularly for testing purposes).  But a staticmethod wouldn't have that problem.  Would using staticmethods for some of these things also fulfill your encapsulation criteria?

jbrockmendel:
>>> where one or more of the speed-critical functions has been overridden to call C functions, but the remainder of the logic is shared. This would be much more difficult if we were using a bunch of bare functions.

It really, really wouldn't.  Either way its an import and an override.

dateutil/parser/isoparser.py
jbrockmendel: Oh, wow.  Yikes.

pganssle:
> I'm thinking of the level of encapsulation as being the module. Especially if its behind a dateutil._parser layer.

It would not be, this is in `dateutil.parser.isoparser`.

> Here's a thought: my cringing is related to implicit statefulness. A classmethod or an instance method carries the risk that a function/method's implementation may not be fully isolated (which I like particularly for testing purposes). But a staticmethod wouldn't have that problem. Would using staticmethods for some of these things also fulfill your encapsulation criteria?

If I'm understanding correctly, neither class nor instance methods imply any sort of state. These classes in particular should be considered immutable (though I'm not sure the easiest way to actually *enforce* this), meaning that you have some data (the configuration) and something that acts on it (the methods).

I tend to avoid `staticmethod` because it's more or less pointless. Class methods get access to class attributes and AFAIK there's no real overhead to including them. In this case, all the class methods actually *use* the fact that they are class methods, since they are retrieving class attributes.

I again fail to see the downside here. None of these things are any harder to test by virtue of the fact that they are essentially namespaced in a class instance, and in fact they may be *easier* to test because you can inherit from them to selectively modify behavior if you want. Encapsulating all these related methods along with their configuration in a class makes it much easier to spawn off variants even if the class were private, since we could privately inherit it to create isoparser variants.

I'm not saying it's outside the realm of possibility to use bare functions, but I don't see the point - it seems like all downside to me.

jbrockmendel:
> If I'm understanding correctly, neither class nor instance methods imply any sort of state.

They don't _imply_ state, but they carry the possibility.  Fundamentally I don't trust future-us to not be dumb.  Functional-style code makes that somewhat easier (for me at least).

> These classes in particular should be considered immutable (though I'm not sure the easiest way to actually enforce this),

I've been working on a _very_ similar problem recently (for `relativedelta`-like objects); let's discuss in a separate thread.  The short answer is "do `__init__` in cython, and pray you never have to worry about pickling".

> I tend to avoid staticmethod because it's more or less pointless.

Yah, they behave like module-level functions but are in the class namespace and can be overriden at the instance/class level.  It's mostly a organizational/encapsulation/style thing to loudly declare "these functions go together".

> In this case, all the class methods actually use the fact that they are class methods, since they are retrieving class attributes.

In e.g. `_parse_tzstr`, `cls._TIME_SEP` is used.  In the function-based approach I advocate, that would be passed as an argument (probably a kwarg with a default).

> Class methods get access to class attributes and AFAIK there's no real overhead to including them

I think that's accurate in python-land, not so sure about cython-land.  Let's not worry about this micro-optimization just yet.

> since we could privately inherit it to create isoparser variants.

For methods that are likely to be overridden I'm totally on board.

I think we've beaten this horse to death (again).  Yah?

jbrockmendel:
I'll take a close look at the tests in the next day or two.

TimSC:
As far as I can tell, this branch has the #531 issues, which would be nice to resolve. (or I could be confusing myself when using git!)

TimSC:
This PR addresses the first two issues in #531 but it fails to parse the string "2010-02-18T16.2", which I think is valid. Perhaps check out the tests at https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/ and add them to the automatic tests?

pganssle:
@TimSC All of #531 apparently applies to `dateutil.parser.parse`, so *none* of the "issues" in that PR are addressed by this parser (though this parser has different behavior, so it has a different default date and it has support for 24:00 as an alias for midnight).

You'll note that the final format is explicitly *not* supported by this version of the parser. See [this line in the documentation](https://github.com/pganssle/dateutil/blob/1c73e342e56ab313d60feef4e098cf052ce2533c/dateutil/parser/isoparser.py#L100). When the other issues with this are worked out, fractional time component support will likely be added.

pganssle:
@jbrockmendel @mariocj89 I've implemented a "minimum feature set" that I think will be *mostly* uncontroversial, such that we'll be free to add features in the future, but we won't have any particular need to break backwards compatibility. I want to just merge this so people can start testing it out and so other contributors can help with the cleanup more easily.

The two things I'm uneasy about but think I'll change in a later PR:

1. I'm not entirely sure that I want to expose `isoparser` publicly, but I don't feel strongly about it.
2. Currently we support the `--MM-DD` format, but I guess that that was only a feature of the 2000 version of the standard. Since 2004 that has been dropped, and it will make the interface much simpler if we drop it.

dateutil/parser/__init__.py
```diff
@@ -3,7 +3,10 @@
 from ._parser import DEFAULTPARSER, DEFAULTTZPARSER
 from ._parser import InvalidDateError, InvalidDatetimeError, InvalidTimeError
 
+from .isoparser import isoparser, isoparse
+
 __all__ = ['parse', 'parser', 'parserinfo',
+           'isoparse', 'Isoparser',
```
mariocj89: this does not match your local variables

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,380 @@
...
+        # If it's a stream, read the whole thing
+        str_in = getattr(str_in, 'read', lambda: str_in)()
+
+        # If it's unicode, turn it into bytes, since ISO-8601 only covers ASCII
+        if isinstance(str_in, six.text_type):
+            # ASCII is the same in UTF-8
+            try:
+                str_in = str_in.encode('ascii')
+            except UnicodeEncodeError as e:
+                msg = 'ISO-8601 strings contain only ASCII characters'
```
mariocj89: I might say something like "should contain" from the error it might not be clear that the issue is the input string containing "invalid" characters

mariocj89: OK, saw the raise from, still would change "contain" to should contain if string refers to the input string

dateutil/parser/isoparser.py
```diff
@@ -0,0 +1,380 @@
...
+    def __init__(self, sep='T'):
+        """
+        :param sep:
+            A single character that separates date and time portions
+        """
+        if (len(sep) != 1 or ord(sep) >= 128 or sep in '0123456789'):
+            raise ValueError('Separator must be a single, non-numeric '
+                             'ASCII character')
+
+        self._sep = sep.encode('ascii')
```
mariocj89: What if sep is a bytestring? In py 2 it'll technically work but it is quite not correct

pganssle: Good point. Adding an xfail test for this.

dateutil/parser/isoparser.py
pganssle: It refers to ISO-8601 strings. All ISO-8601 strings consist of only ASCII characters, even though there are invalid things *intending* to be ISO-8601 strings that contain non-ascii characters.

dateutil/test/test_isoparser.py
```diff
@@ -0,0 +1,446 @@
...
+        datetime(2014, 2, 4, 12, 30, 15, 224000,
+                 tzinfo=tz.tzoffset(None, timedelta(hours=5))))])
+def test_bytes(isostr, dt):
+    assert isoparse(isostr) == dt
+
+
+###
+# Invalid ISO strings
+@pytest.mark.parametrize('isostr,exception', [
+    ('201', ValueError),                        # ISO string too short
```
mariocj89: Seems it fails nicely but would be good to get a test case for dates as:
YYYY-MM-D as 2016-12-1

For future development

pganssle: I don't think `YYYY-MM-D` is a supported format, is it?

jbrockmendel:
Little bit late, but LGTM.  I'd like to implement robust profiling before going too much further down this road (i.e. into cython-land).

The numpy/pandas iso8601-esque parsers catch "today" and "now".  Are those part of the spec?

pganssle:
> The numpy/pandas iso8601-esque parsers catch "today" and "now". Are those part of the spec?

Most decidedly not. The spec is about very precise specifications of dates and times. `now` and `today` are among the least precise ways to serialize a date and time.

PR #490, opened 2017-11-02
Refactor _parse_numeric_token out from the big parsing loop

jbrockmendel:
Cut/paste/de-indent.

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -726,122 +726,8 @@ class parser(object):
                     value = None
 
                 if value is not None:
-                    # Token is a number
-                    len_li = len(l[i])
-
-                    if (len(ymd) == 3 and len_li in (2, 4) and
-                          res.hour is None and
-                          (i + 1 >= len_l or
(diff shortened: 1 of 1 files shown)


pganssle:
I'm not totally convinced of the point of this. Now instead of a huge loop body you have a huge function body? Is that really an improvement?

jbrockmendel:
The idea is that it would be really helpful if each step in the loop was just one function call, so we could step through it interactively.  Instead of doing that directly, this is just taking one step in that direction by separating out a big chunk of it.

We've got a few different threads going in which we're not 100% on the same page.  Let's put this on the backburner until we have some concensus-momentum going.

pganssle:
@jbrockmendel Hm... I think I may have screwed this up by doing a rebase? There was a merge conflict that I fixed and did a force push, and it broke the `dBY` test. I'm not sure why. Maybe best if you force push the original version back up here, then do the rebase again on your side and fix the merge conflicts.

PR #491, opened 2017-11-02
port tokenization from downstream, implement first test case

jbrockmendel:
The `tokenization` module here is taken from a creatively-named `dateutil2` I forked a while back.  I think this part can be made useful here, largely to simplify/robustify long `elif ....` clauses (many of which still don't do as much validation as they could/should.

For now the only check this actually uses is a more robust float check (along with a test case that fails under the status quo).

The new classes are pretty solid and have doctests.  The docstrings will need to be updated to a grown-up format.

### DIFF
--- a/dateutil/tokenization.py
+++ b/dateutil/tokenization.py
@@ -1,0 +1,399 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import string
+import re
+import calendar
+
+from six import text_type as unicode
+
+
+class Token:
+    @staticmethod
+    def is_valid_day(year, month, day):
+        """is_valid_day(year, month, day)
+
+        >>> Token.is_valid_day(None, None, 14)
+        True
+
+        >>> Token.is_valid_day(None, 11, 30)
+        True
+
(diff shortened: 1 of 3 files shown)


dateutil/tokenization.py
```diff
@@ -0,0 +1,399 @@
...
+            # Note that ''.isdigit() returns False, so this handles
+            # the case of a trailing "." or ","
+            return False
+
+        return True
+
+
+_inf = float('INF')
+_minf = float('-INF')
+def check_float_like(item):
```
pganssle: This is some non-date-related utility function. Definitely not a good fit for dateutil.

pganssle:
I'm not sure how good a fit this is. What exactly is it for? If it's just to be used for internal logic for the parser, it definitely shouldn't be its own free-standing public module.

dateutil/tokenization.py
jbrockmendel: This func is used for the added test case `test_misleading_float`.

jbrockmendel:
> it definitely shouldn't be its own free-standing public module.

I'm not opposed to an underscore.

PR #493, opened 2017-11-04
Proof of concept _constructor property

jbrockmendel:
This is meant as a demonstration more than anything else.

There are some attributes/methods of the `parser` class that thematically don't really belong there.  In particular I'm thinking of `_result` and `_build_tzinfo`.  They are currently part of `parser` for subclassing reasons.

The demonstration here is to separate these out into a post-parsing class here called `ParseResult`.  The most important thing here is the `_res_class` property.  It provides a hook for subclassing.  So if downstream someone wants to post-process results differently, they implement a different _resultbase subclass and change the _res_class property.

The pattern could also be applied to allow hooks for `_ymd`, `_timelex`or `parserinfo` (though parserinfo already has a mechanism for passing a custom object).

### DIFF
--- a/dateutil/parser.py
+++ b/dateutil/parser.py
@@ -547,18 +547,39 @@ class ParseResult(_resultbase):
                  "hour", "minute", "second", "microsecond",
                  "tzname", "tzoffset", "ampm"]
 
-    def _build_tzaware(self, tzinfos, result, naive):
-        if callable(tzinfos) or (tzinfos and result.tzname in tzinfos):
+    def _build_naive(self, default):
+        repl = {}
+        for attr in ("year", "month", "day", "hour",
+                     "minute", "second", "microsecond"):
+            value = getattr(self, attr)
+            if value is not None:
+                repl[attr] = value
+
+        if 'day' not in repl:
+            # If the default day exceeds the last day of the month, fall back
+            # to the end of the month.
+            cyear = default.year if self.year is None else self.year
+            cmonth = default.month if self.month is None else self.month
+            cday = default.day if self.day is None else self.day
+
+            if cday > monthrange(cyear, cmonth)[1]:
+                repl['day'] = monthrange(cyear, cmonth)[1]
+
+        ret = default.replace(**repl)
+        return ret
+
+    def _build_tzaware(self, tzinfos, naive):
+        if callable(tzinfos) or (tzinfos and self.tzname in tzinfos):
             tzinfo = self._build_tzinfo(tzinfos,
-                                        result.tzname, result.tzoffset)
+                                        self.tzname, self.tzoffset)
             aware = naive.replace(tzinfo=tzinfo)
-        elif result.tzname and result.tzname in time.tzname:
+        elif self.tzname and self.tzname in time.tzname:
             aware = naive.replace(tzinfo=tz.tzlocal())
-        elif result.tzoffset == 0:
+        elif self.tzoffset == 0:
             aware = naive.replace(tzinfo=tz.tzutc())
-        elif result.tzoffset:
-            aware = naive.replace(tzinfo=tz.tzoffset(result.tzname,
-                                                     result.tzoffset))
+        elif self.tzoffset:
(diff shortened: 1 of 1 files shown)


pganssle:
I'm pretty sure that `_result` is in `parser` for namespacing reasons, not for any subclassing reason. `_build_tzinfo` definitely is.

I think this is not an unreasonable way to break things out, but I actually think that `_result` can be pretty easily replaced with `relativedelta`, or possibly `partialdatetime` when #449 is complete, which would be very useful, since then passing a `relativedelta` or `partialdatetime` to the `default` argument would allow `relativedelta` to be *returned* so you can see which components were neither in your default nor in the parsed string.

If that were done, the parts of `_result` that are not part of `relativedelta` (`tzname`, `tzoffset`, `ampm`) would be local to the `_parse` loop.

That said, I don't think that `_build_tzinfo` is a particular outlier here. Everything else is taking strings representing datetime components and turning them into constructor inputs, why should `tzinfo` be any different?

In addition to kinda going "too far", I also think this doesn't go far enough, in the sense that I'd like an intermediate result class that is *reusable*, so that we can cache the results. If the parser infers a given format, I'd like to know *where* it found each token. Might be that we first construct an abstract tree that *can* contain ambiguities, like this:

```
"20140304" ->
         %Y-()
           / \
        %m%d %d%m
```

Or something of that nature (likely we'd want to create some reasonable date parsing DSL that's more robust than `strptime`'s). Downstream consumers who want unusual control over their datetime parsing can manipulate the intermediate syntax tree, and we can cache it so that later parses are fast.

pganssle:
I'll also note that if we make `tzparser` public and give it some public-facing functions that advertise what kinds of strings it will accept (`min_len`, `max_len`, something like that), we can decompose the `_build_tzinfo` function into a single `tzparser()` call.

jbrockmendel:
> I'm pretty sure that _result is in parser for namespacing reasons, not for any subclassing reason. _build_tzinfo definitely is.

namespacing as in not wanting it to be in the module namespace?  I was under the impression that the reason you wanted to move `_recombine_skipped`, `_build_tzinfo`, etc from being functions to being methods was for the over-arching "easier to subclass" agenda.  Am I more confused than previously thought?

> [...] when #449 is complete

That may be true, but that definitely involves some pretty big design+API decisions, and AFAIK is likely a ways off.  I'll read over this more carefully shortly.  If there's WIP I'll be happy to take a look at that.

> If that were done, the parts of _result that are not part of relativedelta (tzname, tzoffset, ampm) would be local to the _parse loop.

That's an interesting point; I'll have to give some thought to how that information could be put to use.

> That said, I don't think that _build_tzinfo is a particular outlier here

`_build_tzinfo` is more of an implementation detail than it is parsing logic.  In fact the methods in `ParseResult` are some of the few that I imagine people _would_ frequently want to subclass/override.

> If the parser infers a given format, I'd like to know where it found each token.

I've given some thought to this, will put together a demonstration before long.  I've been using just a dict+set to represent that during the parsing phase, haven't put much thought into what form it would be returned as.

> Might be that we first construct an abstract tree that can contain ambiguities

Happy to discuss this, but I see it as orthogonal to the suggestion here.

> [...] we can decompose the _build_tzinfo function into a single tzparser() call

That might be a nice separation.

jbrockmendel:
A few thoughts:

1) I'm increasingly liking the idea of a ParseResult-type class as the object being returned from `parser._parse`.  `parser` can be focused on the actual string processing, while `ParseResult` can focus on what to do with the intermediate representation.

2) The `_res_class`-style hook could also be useful for `_timelex`, since there is currently no nice way to use a subclassed _timelex.

3) Related to the partialdatetime discussed elsewhere: there is nothing requiring parser._parse to only identify YmdHMS.  In particular, I think we should consider recognizing a) nanosecond precision, b) BC/BCE years, and c) quarters.  Most users wouldn't be affected, but those who want access to the extra info could get it via either parser._parse or subclassing ParseResult.

PR #494, opened 2017-11-05
Define __all__ in the top level package

### LINKED ISSUE #406: __all__ not defined on root package
Hi, I don't know if this is a problem or intentionally planed, but because of namespace conflicts, we typically import the base package and not deep into it. For example:
```
import dateutil

dateutil.parser.parse('Fri Jun 30 12:00:56 EDT 2017')
```

However this appears to fail with the error:
> AttributeError: 'module' object has no attribute 'parser'

The solution appears to be: `from dateutil import parser` and then call parser, I'm guessing this is because the `__all__` in the root `__init__.py` doesn't exist thus not listing any packages. Is this intended behavior? Should I just use my above workaround?
Thanks!

mariocj89:
Even if using import * is discouraged in general and considered a bad practice, it is considered by convention to define the public API of a module. That combined with some users asking for it lead to this commit.

Closes #406

### DIFF
--- a/dateutil/__init__.py
+++ b/dateutil/__init__.py
@@ -1,2 +1,5 @@
 # -*- coding: utf-8 -*-
 from ._version import VERSION as __version__
+
+__all__ = ['easter', 'parser', 'relativedelta', 'rrule', 'tz',
+           'utils', 'zoneinfo']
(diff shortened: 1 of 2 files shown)


pganssle:
Thanks, Mario!

tebriel:
Thanks!

PR #495, opened 2017-11-05
Update test runner to use pytest

pganssle:
This changes the test runner over to pytest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,14 @@ before_install:
   - if [[ $TRAVIS_PYTHON_VERSION < '3.0' ]]; then pip install mock; fi
   - pip install codecov
   - pip install freezegun
+  - pip install pytest
 
 install:
   - pip install six
   - ./ci_tools/retry.sh python updatezinfo.py
 
 script:
-  - coverage run --omit=setup.py,dateutil/test/* setup.py test
+  - coverage run --omit=setup.py,dateutil/test/* -m pytest
 
 after_success:
   - codecov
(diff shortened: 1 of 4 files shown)


pganssle:
@jbrockmendel @mariocj89 I think this is fine, any comments? I'd like to merge ASAP so that everyone else can start using pytest.

mariocj89:
Looks great! I'd personally add a -vvvv for pytest to get the history of test run on the builds

pganssle:
@mariocj89 Ah, good call, I didn't notice that that was missing.

pganssle:
What is the difference between `-v` and `-vvvv`?

mariocj89:
Indeed, Nothing for pytest, just bad habits brought from old days with nose 😆

mariocj89:
Just realized, it actually changes the way some diffs are reported.

-v will add the test name in the reports. having multiple v just increases the level of verbosity, which adds extra information on some assertion failures.

Example, if you assert two diffs which are different but big, adding extra verbosity will print the whole dict whilst just -v won't.
You are probably fine with just -v here

pganssle:
OK, we can change it later if `-vvv` is better.

.travis.yml
```diff
@@ -21,13 +21,14 @@ before_install:
...
   - pip install freezegun
+  - pip install pytest
 
 install:
   - pip install six
   - ./ci_tools/retry.sh python updatezinfo.py
 
 script:
-  - coverage run --omit=setup.py,dateutil/test/* setup.py test
+  - coverage run --omit=setup.py,dateutil/test/* -m pytest -v
```
jbrockmendel: There is a pytest-cov plugin, not sure to what extent there is a "preferred" usage of that vs this usage.

jbrockmendel:
Little late, but LGTM.

PR #497, opened 2017-11-05
Return a singleton from tzutc()

pganssle:
Python assumes that `tzinfo` objects representing the same zone are effectively singletons, which is why intra- vs inter-zone semantics apply based on `dt1.tzinfo is dt2.tzinfo`, not `dt1.tzinfo == dt2.tzinfo`.

`tzutc()` is the simplest case, since it only ever represents one zone. I've also added a convenience constant to the top level `tz/__init__.py`, so people can just use it as a constant.

Eventually the other classes will store their representations in caches (though this is more complicated and has more implications for the API long term).

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -596,6 +596,12 @@ class TzWinFoldMixin(object):
 ###
 # Test Cases
 class TzUTCTest(unittest.TestCase):
+    def testSingleton(self):
+        UTC_0 = tz.tzutc()
+        UTC_1 = tz.tzutc()
+
+        self.assertIs(UTC_0, UTC_1)
+
     def testOffset(self):
         ct = datetime(2009, 4, 1, 12, 11, 13, tzinfo=tz.tzutc())
 
@@ -614,7 +620,6 @@ class TzUTCTest(unittest.TestCase):
         UTC0 = tz.tzutc()
         UTC1 = tz.tzutc()
 
-        self.assertIsNot(UTC0, UTC1)
         self.assertEqual(UTC0, UTC1)
 
     def testInequality(self):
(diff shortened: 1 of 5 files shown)


dateutil/test/test_tz.py
```diff
@@ -614,7 +620,6 @@ def testEquality(self):
         UTC0 = tz.tzutc()
         UTC1 = tz.tzutc()
 
-        self.assertIsNot(UTC0, UTC1)
```
pganssle: I don't know why I was testing this.

jbrockmendel:
To what extent does this complement vs duplicate pytz.utc?

pganssle:
@jbrockmendel `pytz` has a whole different thing going on, but UTC is so simple that basically all implementations will have the same properties. This version has some fast paths for `dateutil`-specific stuff (like checking ambiguities), but will work the same as `pytz.utc`. Of course, there's no way to make `pytz.utc is dateutil.tz.UTC` or `dateutil.tz.UTC is datetime.timezone.UTC`, but that's fine. This will mostly make it slightly cheaper to use UTC from `dateutil` (no need to construct a new instance or even call the constructor now), and move us down the road to a one-object-per-zone model.

pganssle:
@abalkin Any thoughts on this? Not sure who else would understand the implications of this well enough to review.

abalkin:
This looks like a no-brainer optimization.  There are several code paths in the datetime module where having the same tzinfo rather then just equal TZ objects saves several calls to utcoffset() methods.

pganssle:
@abalkin Thanks. That's pretty much what I thought. This is the least controversial of the singletons anyway. I'm going to try and come up with an interface for the others that's not totally nuts but ports these "fast path" qualities to the other types (particularly `tzfile`), I imagine the details of that one will be more important.

PR #499, opened 2017-11-06
Add test requirements

pganssle:
Switch over test dependency management to a more centralized `test_requirements.txt`.

HT @mariocj89

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,6 @@ datetime module available in the Python standard library.
       package_data={"dateutil.zoneinfo": ["dateutil-zoneinfo.tar.gz"]},
       zip_safe=True,
       requires=["six"],
-      tests_require=["freezegun", "pytest"],
       install_requires=["six >=1.5"],  # XXX fix when packaging is sane again
       classifiers=[
           'Development Status :: 5 - Production/Stable',
--- a/tox.ini
+++ b/tox.ini
@@ -11,5 +11,4 @@ envlist =
 
 [testenv]
 commands = pytest {posargs}
-deps =
-    six,pytest,freezegun
+deps = -rtest_requirements.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,10 @@ before_install:
   # Install codecov
   - if [[ $TRAVIS_PYTHON_VERSION == '3.2' ]]; then pip install coverage==3.7.1; fi
   - if [[ $TRAVIS_PYTHON_VERSION < '3.0' ]]; then pip install mock; fi
+  - pip install -r test_requirements.txt
   - pip install codecov
-  - pip install freezegun
-  - pip install pytest
 
 install:
-  - pip install six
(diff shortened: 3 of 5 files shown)


jbrockmendel:
`six` is a regular requirement.  Are test_requirements on top of regular requirements or instead of?

pganssle:
I suppose it doesn't matter much either way, since six is the *only* regular requirement.

pganssle:
We can tweak the way requirements are being imported a bit later, for now I'll just leave `six`, because I'm trying to get us to a reasonable baseline so I can rebase the Isoparser branch.

tox.ini
```diff
@@ -11,5 +11,4 @@ envlist =
 
 [testenv]
 commands = pytest {posargs}
-deps =
-    six,pytest,freezegun
+deps = -rtest_requirements.txt
```
mariocj89: I know I'm late here but you'd usually see requirements-dev.txt rather than test_requirements.

It is even "reserved" in PyPI https://pypi.python.org/pypi/requirements-dev.txt

pganssle: Definitely not too late to move it. What do you think of the suggestion of dropping `six` proposed above? We can do both changes in one PR.

mariocj89: Your requirement file should provide a "reproducible python environment". It should include all your dependencies.
What some people do if you want to split runtime/build vs test requirements you can include one from another.

Example: having `requirements.txt` with actual dependencies and `requirements-dev.txt` including requirements.txt (you can do -r)

mariocj89: What you might want though is to pin the versions in your requirements file.

mariocj89: You might be able to add mock using [markers](https://www.python.org/dev/peps/pep-0426/#environment-markers) as well here:
`mock ; python_version < '3.0'` (Try this out, I am just writing in github)

jbrockmendel:
The six thing I asked about largely because I've had tox problems when a dependency gets specified multiple times.  If it isn't causing problems here, don't spend much time worrying about it.

setup.py
```diff
@@ -24,7 +24,6 @@
       package_data={"dateutil.zoneinfo": ["dateutil-zoneinfo.tar.gz"]},
       zip_safe=True,
       requires=["six"],
-      tests_require=["freezegun", "pytest"],
```
wagner-certat: Why did you remove this list? I could still be read from the file.

pganssle: I believe this is only ever useful for `python setup.py test`, and now that we're using `pytest` (and haven't configured `setup.py test` to run the tests yet), it's useless to have it around. There's probably some utility in setting up `setup.py test` so it works properly, but my understanding is that the `test` target is considered something of a failed experiment.

wagner-certat: > and now that we're using pytest (and haven't configured `setup.py` test to run the tests yet)

Ok, I can look up how other projects call pytest. I think it's quite simple.

Issue #507, opened 2017-11-08
parser not installed in master


pganssle:
Related to #506 and discovered by @jreback in https://github.com/pandas-dev/pandas/issues/18141#issuecomment-342954600, after moving `parser.py` into `parser/`, I didn't update the package list in `setup.py`.

I'm going to switch us over to using `find_packages` and clean up the `requirements.txt` comments brought up by @mariocj89 in #499 in one PR. If that happens before we fix #506, I'll just run the tests myself appropriately.

pganssle:
Fixed by #509

PR #509, opened 2017-11-09
Fix package discovery in setup.py

pganssle:
One way to possibly address #507


pganssle:
This seems to be failing in the expected way.

pganssle:
I think this is breaking `pandas` builds, so I'm going to merge this without review. We can nitpick about how I accomplished it later.

PR #510, opened 2017-11-09
Use the self.ystridx already found for unambiguous year < 99 dates

### LINKED ISSUE #293: Inconsistent behaviour when dealing with dates close to year 0

pganssle:
Fixes #293.

@jbrockmendel Let me know if I missed something here, but it seems like all this checking for missed year tokens is unnecessary since `_ymd` is remembering which one is the year in the first place. This also allows us to drop the time string from `_ymd` (it didn't belong there anyway).

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -38,6 +38,7 @@ import time
 from calendar import monthrange
 from io import StringIO
 
+import six
 from six import binary_type, integer_types, text_type
 
 from .. import relativedelta
@@ -378,10 +379,9 @@ class parserinfo(object):
 
 
 class _ymd(list):
-    def __init__(self, tzstr, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         super(self.__class__, self).__init__(*args, **kwargs)
         self.century_specified = False
-        self.tzstr = tzstr
         self.dstridx = None
         self.mstridx = None
         self.ystridx = None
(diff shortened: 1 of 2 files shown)


dateutil/parser/_parser.py
```diff
@@ -378,10 +379,9 @@ def validate(self, res):
 
 
 class _ymd(list):
-    def __init__(self, tzstr, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
```
jbrockmendel: +1 for dropping this.

dateutil/parser/_parser.py
```diff
@@ -963,7 +939,7 @@ def _parse_numeric_token(self, tokens, idx, info, ymd, res, fuzzy):
                     if value is not None:
                         ymd.append(value, 'M')
                     else:
-                        raise InvalidDatetimeError(ymd.tzstr)
+                        raise InvalidDatetimeError()
```
jbrockmendel: Admittedly accessing the input via ymd.tzstr was a bit of a kludge, but it still seems like it should be included in the exception msg.  Maybe attach it to `res` since that is kind of a ParseState?

While we're at it, maybe truncate the string that goes in there at, say, 128 characters?

pganssle: No need for any of that. For one thing, none of these exceptions ever made it out of `_parse`, because they all got caught and raised as `ValueError` for whatever reason.

If we want to fix that as part of the cleanup, we can just catch, attach the string in the `_parse` scope, then re-raise.

pganssle:
Oops, good point.

jbrockmendel:
Contingent on a test case, this gets a "LGTM"

pganssle:
OK, test case added, so merging.

PR #513, opened 2017-11-11
Add gitter badge, remove defunct pypi badge

pganssle:
Supplants #512.

This adds a gitter for the project, in case anyone wants that.

Also, the pypi downloads badge stopped working some time ago, so we can remove it until something better comes along.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,10 @@
 dateutil - powerful extensions to datetime
 ==========================================
 
+.. image:: https://img.shields.io/pypi/v/python-dateutil.svg?style=flat-square
+    :target: https://pypi.python.org/pypi/python-dateutil/
+    :alt: pypi version
+
 .. image:: https://img.shields.io/travis/dateutil/dateutil/master.svg?style=flat-square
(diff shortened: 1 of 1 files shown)


jbrockmendel:
gitter is like a chatroom right?  What's the use case?

pganssle:
Yes, it's a chat. Just another way to contact - ask questions, discuss implementations, etc, in real time. I only added it because it's a fairly low-cost thing to do.

jbrockmendel:
OK.  Will it need to have "support staff" online in real-time to be useful?

pganssle:
I doubt that will be necessary. The flow of `python-dateutil` questions on SO is low enough that I can handle basically all of them myself in a reasonable amount of time. I don't think the gitter will be particularly popular, it's just an avenue where someone can have real-time conversations about the software if need be.

I imagine it will be most useful for sprints (where a bunch of people will be around at the same time and have some need to coordinate) or some sort of scheduled or semi-scheduled discussion about some particular implementation that needs more rapid back-and-forth than you can get with carefully composed replies on an issue tracker.

PR #515, opened 2017-11-12
Deprecate private interfaces

pganssle:
Per my comments in pandas-dev/pandas#18141, I think that it is not particularly expensive to preserve some of the private interface, with an explicit deprecation warning.

The main question I have here is whether we should use `DeprecationWarning`, which is not visible by default in Python 2.7+, or if we should use something more visible, like our own custom warning class that *is* visible, or possibly make some sort of context manager that temporarily enables `DeprecationWarning`.

Kinda unfortunate that `DeprecationWarning` was disabled by default, since that encourages libraries to do shit like what I just described, but it is what it is.

### DIFF
--- a/dateutil/parser/__init__.py
+++ b/dateutil/parser/__init__.py
@@ -5,3 +5,51 @@ from ._parser import InvalidDateError, InvalidDatetimeError, InvalidTimeError
 
 __all__ = ['parse', 'parser', 'parserinfo',
            'InvalidDatetimeError', 'InvalidDateError', 'InvalidTimeError']
+
+
+###
+# Deprecate portions of the private interface so that downstream code that
+# is improperly relying on it is given *some* notice.
+
+
+def __deprecated_private_func(f):
+    from functools import wraps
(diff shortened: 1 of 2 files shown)


jbrockmendel:
Is FutureWarning a viable alternative?

pganssle:
@jbrockmendel `FutureWarning` is intended for when the semantics of something change. I think you'd use it in a case like Python 2 integer division behavior vs. Python 3 integer division behavior. Or presumably if we thought it mattered, we could raise a `FutureWarning` when two time zones of the same type compare equal but `tz1 is not tz2`, to say that semantically equal time zones of the same type will have a different semantic relationship to one another in the future. This PR deprecates an old (private) interface, so `DeprecationWarning` is more appropriate.

In any case, avoiding `DeprecationWarning` just so it's visible isn't much different from just defining a `DeprecatedPrivateInterfaceWarning` and raising that. The question is really whether we want to circumvent the default behavior of hiding deprecation warnings or not.

PR #516, opened 2017-11-12
Fix test requirements

pganssle:
We'll need to pin `pytest` at version >= 2.0 if we want to use `skipIf`, and we'll need `skipIf` if we want to avoid calling `pytest.warns` on Python 3.2.

While I was in there I did a bit of cleanup.


jbrockmendel:
This generally looks good.  Not sure if we should still be supporting 3.2, though that decision doesn't need to be made here.

pganssle:
Yeah, I think after the 2.7.0 release we can drop 3.2 support, but we really should be cutting a new release soon and there's no urgent need to remove 3.2 support.

Issue #511, opened 2017-11-10
BLD: would be helpful to have a dev version number


jreback:
e.g. current master is fixed at 2.6.1 (hard coded). maybe use https://github.com/warner/python-versioneer

IOW I was to have in tests ``LooseVersion(dateutil.__version__) > '2.6.1'`` but this is False (both on master and the released version), should be True on master.

jreback:
xref https://github.com/pandas-dev/pandas/issues/18141

pganssle:
Agreed. Been meaning to set up some sort of automatic versioning for a while anyway. I'll take a look at versioneer when I get a moment.

pganssle:
@jreback Partially solved by #518.

I'll leave this open because versioneer or something like it would be nice to have.

PR #518, opened 2017-11-12
Bump version to 2.7.0dev0

pganssle:
This is a temporary solution to #511 until we can get a proper auto-version in place.


pganssle:
Merging without review because I doubt this will be controversial and it will hopefully unblock pandas and other dependencies who want to code against master.

Issue #285, opened 2016-07-25
Add in RFC7529 extensions to rrule.


pganssle:
Looks like [RFC7529](https://tools.ietf.org/html/rfc7529) provides some significant extensions to the RRULE, including a more satisfactory resolution to #149.

We should probably implement these extensions. I haven't read it thoroughly, but we may want to see if these extensions allow `BYEASTER` to be easily recast as a combination of other rules. This will allow us to generate compliant `rrulestr` strings specified with the `byeaster` input.

pganssle:
One thing to note is that I believe a full implementation of `RSCALE` would be a *lot* of work, so we'll limit this to supporting only the Gregorian calendar for now.

pganssle:
Note to anyone who wants to work on the `skip` portion of this, the work is >80% done by @yoch in #522. If someone wants to do the last part, please do so by forking his/her PR.

ryancausey:
> Note to anyone who wants to work on the `skip` portion of this, the work is >80% done by [@yoch](https://github.com/yoch) in [#522](https://github.com/dateutil/dateutil/pull/522). If someone wants to do the last part, please do so by forking his/her PR. (And probably best to at least ping yoch to mention it, since your work will be for naught if they update the PR before your fork is accepted).

@pganssle is this still an accurate description of the state of #522 if someone wants to try and get it moving again? I see that there is mention of you performing some refactoring work on the branch in the PR comments.

PR #522, opened 2017-11-14
implement the skip option

yoch:
Try to implement the SKIP option (which solves issue #149).

To be compatible with the RFC-7529 (#285), I also added RSCALE element because skip depends on it, but without the support of non-gregorian calendars.


yoch:
I haven't created new tests yet, but here some examples:

```pycon
>>> from dateutil.rrule import *
>>> from pprint import pprint
>>> start_date = datetime(2014, 12, 31) 
>>> rr=rrule(freq=MONTHLY, count=3, dtstart=start_date); print(rr); pprint(list(rr))
DTSTART:20141231T000000
RRULE:FREQ=MONTHLY;COUNT=3
[datetime.datetime(2014, 12, 31, 0, 0),
 datetime.datetime(2015, 1, 31, 0, 0),
 datetime.datetime(2015, 3, 31, 0, 0)]
>>> rr=rrule(freq=MONTHLY, count=3, dtstart=start_date, rscale='GREGORIAN', skip='BACKWARD'); print(rr); pprint(list(rr))
DTSTART:20141231T000000
RRULE:FREQ=MONTHLY;COUNT=3;RSCALE=GREGORIAN;SKIP=BACKWARD
[datetime.datetime(2014, 12, 31, 0, 0),
 datetime.datetime(2015, 1, 31, 0, 0),
 datetime.datetime(2015, 2, 28, 0, 0)]
>>> rr=rrule(freq=MONTHLY, count=3, dtstart=start_date, rscale='GREGORIAN', skip='FORWARD'); print(rr); pprint(list(rr))
DTSTART:20141231T000000
RRULE:FREQ=MONTHLY;COUNT=3;RSCALE=GREGORIAN;SKIP=FORWARD
[datetime.datetime(2014, 12, 31, 0, 0),
 datetime.datetime(2015, 1, 31, 0, 0),
 datetime.datetime(2015, 3, 1, 0, 0)]
>>> rr=rrule(freq=MONTHLY, count=3, dtstart=start_date, skip='FORWARD'); print(rr); pprint(list(rr))
  File "/cygdrive/d/Projets/dateutil/dateutil/rrule.py", line 676, in __init__
    assert skip is None, 'SKIP is not allowed without RESCALE'
AssertionError: SKIP is not allowed without RESCALE
```

yoch:
Important remark: this SKIP doesn't cover all the possibles cases.

The RFC give these cases : 

>    1.  The start date of the recurrence is a leap day in the specified
>        calendar system.
>    2.  The start date of the recurrence is in a leap month in the
>        specified calendar system.
>    3.  The start date of the recurrence has a day-of-month value greater
>        than the smallest day-of-month value for any month in any year in
>        the specified calendar system.
>    4.  A "BYMONTHDAY" element in an "RRULE" has a day-of-month value
>        greater than the smallest day-of-month value for any month in any
>        year in the specified calendar system.
>    5.  A "BYMONTH" element in an "RRULE" has a value corresponding to a
>        leap month in the specified calendar system.
>    6.  A combination of a "BYMONTHDAY" element and a "BYMONTH" element
>        in an "RRULE" has a value corresponding to a leap day in the
>        specified calendar system.
>    7.  A "BYYEARDAY" element in an "RRULE" has an absolute value greater
>        than the smallest number of days in any year in the specified
>        calendar system.
>    8.  A "BYWEEKNO" element in an "RRULE" has an absolute value greater
>        than the smallest number of weeks in any year in the specified
>        calendar system.

1. cannot occurs because dtstart must be valid;
2. and 5. cannot occurs in Gregorian calendar (no leap month);
3. is adressed by this PR;
4. , 6. , 7. and 8. need to be adressed.

dateutil/rrule.py
```diff
@@ -658,6 +659,23 @@ def __init__(self, freq, dtstart=None,
             self._bysecond = tuple(sorted(self._bysecond))
             self._original_rule['bysecond'] = self._bysecond
 
+        # rscale and optional skip values
+        if rscale is not None:
+            assert rscale == 'GREGORIAN', 'unsupported RSCALE'
```
pganssle: First off, I'm not comfortable with using assertions for error handling. This should raise an error of some sort unless you're just asserting some invariant of the code (consider that assertions are turned off if Python is run with `python -O`.

I'm of two minds about how to implement this. One thing we could do is have a fixed list of calendar types that are considered valid, even if unsupported, adding a class attribute like this:

```python
# List taken from https://www.unicode.org/repos/cldr/tags/latest/common/bcp47/calendar.xml
VALID_RSCALES = {'GREGORIAN', 'GREGORY',
                 'BUDDHIST', 'CHINESE', 'COPTIC', 'DANGI', ... }
```

Then change this check to:
```python
if rscale is not None:
    rscale = rscale.upper()
    if rscale in self.VALID_RSCALES:
        if rscale not in {'GREGORY', 'GREGORIAN'}:
            raise NotImplementedError('Non-gregorian calendar '
                                      '{} not supported.'.format(rscale))
    else:
        raise ValueError('Invalid RSCALE component: {}'.format(rscale))
```

Note that I also support lower and mixed-case calendar names there with the `upper` call. That said, an alternative to using string comparison at all is to simply have an enum-type class that enumerates the supported calendars. That seems like it might have some advantages here, but depending on the implementation it might make it harder for people to extend `dateutil` to support other calendars. String comparison seems like a good start.

dateutil/rrule.py
```diff
@@ -658,6 +659,23 @@ def __init__(self, freq, dtstart=None,
...
+
+            self._original_rule['rscale'] = rscale
+
+            if skip is not None:
+                self._original_rule['skip'] = skip
+                self._skip = skip
+            else:
+                self._skip = 'OMIT'
+
+            assert self._skip in ['OMIT', 'BACKWARD', 'FORWARD'], 'invalid SKIP'
```
pganssle: I think in this case the possible values for `SKIP` are probably bounded such that an enum-like structure (can't use real enums because we still support Python 2.7) would be better here. In that case, I'd find an assertion of type or value more acceptable since it's probably a superfluous check that can be omitted in optimized code.

dateutil/rrule.py
```diff
@@ -678,6 +696,14 @@ def __str__(self):
         dateutil-specific extension BYEASTER.
         """
 
+        def _fmt(obj):
```
pganssle: What is the purpose of this function? The other two branches you added are never hit.

yoch: Without this function, strings are formated uncorrectly by: `parts.append(partfmt.format(name=name, vals=(','.join(str(v) for v in value))))`. (The second branch never hit because I didn't add tests, but the last one may not be used.)

dateutil/rrule.py
```diff
@@ -798,6 +826,27 @@ def _iter(self):
...
         while True:
+            leap = False
+
+            # SKIP rule
+            if day > 28 and self._skip is not None and self._skip != 'OMIT':
+                daysinmonth = calendar.monthrange(year, month)[1]
+                if day > daysinmonth:
+                    if self._skip == 'BACKWARD':
+                        day = daysinmonth
+                        leap = True
```
pganssle: Why is this called `leap`? Shouldn't it be called `skipped`? Or `shifted` or something?

pganssle:
@yoch Are you intending to implement 4, 6, 7 and 8 in a separate PR, or is this a "WIP" PR that you plan on updating to include those cases, or is this just a partial implementation and you're not planning on completely implementing the `skip` option?

I think we can include a *full* implementation of skip (minus 1, 2 and 5) in dateutil 2.7.0, otherwise we can defer merging this until after the 2.7.0 release so that the full implementation can be ready for the 2.8.0 release.

dateutil/rrule.py
```diff
@@ -798,6 +844,27 @@ def _iter(self):
...
+                    elif self._skip == 'FORWARD':
+                        day = 1
+                        month += 1
+                        if month == 13:
+                            month = 1
```
pganssle: I believe it's not actually possible to hit this case in the Gregorian calendar when only allowing skip condition 3, since December always has 31 days.

I'd have to check the RFC rules to see if, with the new extension, you can do something like `bymonthday=32` and have it fall forward. I suspect that this is unnecessary, since that should be equivalent to a `bymonthday=1` rule.

If this case is forbidden by the RFC (or if this set of conditions is only ever reached for end-of-month skip conditions), just drop this whole `if` condition and replace it with this:

```python
# In the Gregorian calendar, December has 31 days, so we should never find
# a skip condition that would roll over to the next year
assert month <= 12, "Skip forward logic resulted in anomalous month"
```

dateutil/test/test_rrule.py
```diff
@@ -2859,6 +2889,56 @@ def testStrValueDate(self):
         self.assertEqual(list(rr), [datetime(1997, 9, 2, 0, 0, 0),
                                     datetime(1998, 9, 2, 0, 0, 0)])
 
+    def testStrSkipBackward(self):
+        rr = rrulestr("DTSTART:20141231T000000\n"
```
pganssle: I'm assuming this is a codecov bug, but for some reason this test is not being reported as reached by codecov...

yoch: It's my mistake. I used the same name for two tests...

yoch:
@pganssle With regard to the missing points, it turns out that it's more complex and probably need to modify the existing code more deeply, so I plan to make a separate PR, if I can.

yoch:
Cases 4 and 6 are now correctly handled (no idea why appveyor fails, but the errors doesn't seems related to the commit).

dateutil/test/test_rrule.py
```diff
@@ -431,6 +431,28 @@ def testMonthlySkipForward(self):
                           datetime(2015, 3, 1, 0, 0),
                           datetime(2015, 3, 31, 0, 0)])
 
+    def testMonthlyByMonthSkipBackward(self):
```
pganssle: Does the new stuff cover the `bymonthday=30` case as well? Can you add tests for that?

yoch: Not yet, in fact this was a fix for BYMONTH. But I've fixed it for BYMONTHDAY now.

Now I don't know how to deal with the negatives cases (do we also interpret FORWARD and BACKWARD in negative sense or not).

pganssle:
Appveyor failure was due to pytest-dev/py#169, seems fixed now.

pganssle:
@yoch I edited your first post to add a "to-do" list tracking the cases, because I was having a hard time remembering which cases were left to do. I wasn't sure if you would be able to update a separate post by me, so I edited your post so that you'd have permissions to update the list.

Thanks again for your hard work on this.

pganssle:
@yoch If you rebase against master, the master branch has a fix for the Appveyor issue.

pganssle:
I've started some refactoring on this, I'm not sure why the `BYMONTHDAY` and `BYYEARDAY` elements were checked off, because I think actually none of the `byxxx` rules are handled. I guess this has to be pushed to 2.9.0.

mojimi:
What can be done to help move this forward? Highly needed feature

immerrr:
@pganssle I have taken a stab at implementing the skipping in https://github.com/dateutil/dateutil/commit/2dfd4a7ac49368aebc9529130b37ab1cceaec862. It fixes the failing tests that were added to this branch in 39d51f68c82529efabc551ea4a043edbbdc1d4cd

My change is somewhat rough: it may raise errors on `bymonthday=32` and does not handle negative bymonthdays at all, but it shows the general idea. WDYT?

Issue #520, opened 2017-11-12
Appveyor (Windows) coverage not uploaded to codecov


pganssle:
It seems that it is a known issue that codecov is not updated from Appveyor (which explains why we have pathetic coverage for the `tzwin` module and anything windows-related). See codecov/support#402. There seems to be some solution available, but the latest version of `codecov-python` does not have the fix.

Looks like in twisted/twisted#831 they pin to `codecov == 2.0.5`, though in the codecov issue a `codecov-exe` is mentioned. Shouldn't be terribly difficult to get it fixed, and that should give us a better idea of what needs to be done to get to 100% coverage.

pganssle:
This may actually be done?

pganssle:
For sprinters: The first item on the list is figuring out if this is actually done from the [codecov](https://codecov.io/gh/pganssle/dateutil). If it is we can close this issue.

anthrotype:
it looks like the coverage reports _are_ being uploaded to codecov.io successfully.

See appveyor's log on the latest build from master (this is just py27, but the others are the same):
https://ci.appveyor.com/project/dateutil/dateutil/build/2.3.1406/job/otf53sc4o06a8tu4#L730

And in the codecov Uploads log page, the Appveyor builds seems to be all listed:
https://codecov.io/gh/dateutil/dateutil/commit/5e6f011c6d490ffc8842a2ea5363741b84766539/build

I think this can be closed.

pganssle:
Thanks @anthrotype!

PR #526, opened 2017-11-16
Pin codecov and py dependencies on Appveyor.

pganssle:
Short-to-medium term fix for #520,  until codecov/support#402 is solved in a release to the python `codecov` client.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,6 +42,7 @@ before_test:
   - xcopy dateutil\\test build_test\\test /E /H /K /O /X 
   - cd build_test
 test_script:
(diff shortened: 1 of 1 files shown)


pganssle:
OK, so while testing this, I realized that we're not being hit by that codecov issue. What was happening was that `codecov run -m pytest -v` was inferring that the only source was `dateutil.test` for whatever reason.

By explicitly adding `coverage run --source=dateutil -m pytest -v`, we get the full report.

Unfortunately, this doesn't actually bring up our coverage at all, because the old erroneous reports from Appveyor were counting `dateutil.test` twice, and `dateutil.test` has excellent coverage.

PR #527, opened 2017-11-16
pin py to <1.5 on py33 (1.5 onward dropped support for py26 and py33)

nicoddemus:
Ref: pytest-dev/py#169


pganssle:
I think this goes too far, it will enforce that `py` is installed in all cases, including in `tox` builds, even if one of our dependencies drops it. I have a more limited solution in #526

nicoddemus:
OK 👍

PR #528, opened 2017-11-16
Improve coverage

pganssle:
Simple changes to improve measured code coverage (through legitimate means).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
   - cd build_test
 
 script:
-  - COVERAGE_FILE=.coverage.main coverage run --source=dateutil,test -m pytest -v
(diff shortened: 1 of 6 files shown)


pganssle:
I see what the problem here is. Because we're not actually *running* the tests from `dateutil.test` it's screwing up the coverage metrics. Damn.

pganssle:
I was able to "trick" codecov into mapping the test coverage in `dateutil.test` even when it's a separate directory by moving that directory into `dateutil/test` (in the build environment), so that's nice.

pganssle:
I spoke too soon. Now it's not covering the `dateutil` package at all :(

pganssle:
Per discussion with various Python luminaries, I think a good first start would be to start using `tox` for the builds.

pganssle:
Actually, I'll open a new branch. This PR can serve as a monument to my failure.

PR #536, opened 2017-12-05
Pin pytest dependency on 3.3

pganssle:
Trying to fix whatever new issues have been created by whatever's happening with py's support for Python 3.3 (see #526).


pganssle:
See #2966.

PR #539, opened 2017-12-05
separate out build_naive and build_aware methods

jbrockmendel:
Ideally I still think these methods should be part of an intermediate results class along the lines of #525, but even in `parser` they are nicely focused.


pganssle:
No immediate objections here.

jbrockmendel:
If this goes in I can follow-up with a PR to get test coverage for the one line that is missed.

PR #547, opened 2017-12-06
WIP: Update link to RFC 2445 / RFC 5545

### LINKED ISSUE #318: RFC822 timestamp using GMT is parsed into BST if local time is UK time

### LINKED ISSUE #543: Update link to RFC 2445 / RFC 5545
Currently we link to the `.txt` version of RFC 2445, but the rich text version has better links and navigation.

I would probably change the link in [`rrule`](https://dateutil.readthedocs.io/en/latest/rrule.html) to link to https://tools.ietf.org/html/rfc5545 instead. Technically this is not quite right, since we're *mostly* based on RFC 2445, but since I'm intending to eventually support RFC 5545, we can accept that all deficiencies in our implementation are bugs, not intended behavior.

A few other places to update to refer to RFC 5545 instead of RFC 2445:
- [tz](https://github.com/dateutil/dateutil/blob/64c6c85ded28dea8a617b91258aabf269d1ca34a/dateutil/tz/tz.py#L1153)
- [rrule test](https://github.com/dateutil/dateutil/blob/2f3a1606d1021a0d673bb1c8f37dd319a6747a2a/dateutil/test/test_rrule.py#L2366)
- [rrule](https://github.com/dateutil/dateutil/blob/2f3a1606d1021a0d673bb1c8f37dd319a6747a2a/dateutil/rrule.py) Do a search and replace for 2445 -> 5545. RFC 2445 Sec. 4.3.10 is replaced by RFC 5545 [Sec. 3.3.10](https://tools.ietf.org/html/rfc5545#section-3.3.10)

To do list:
- [x] Update link to https://tools.ietf.org/html/rfc5545
- [x] Update references to 2445 to refer to 5545
- [ ] Update references to Sec. 4.3.10 to refer to Sec 3.3.10

mariocj89:
Updating the link of the RFC from 2445 to RFC 5545.

Closes #543

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -2,7 +2,7 @@
 """
 The rrule module offers a small, complete, and very fast, implementation of
 the recurrence rules documented in the
-`iCalendar RFC <http://www.ietf.org/rfc/rfc2445.txt>`_,
+`iCalendar RFC <https://tools.ietf.org/html/rfc5545>`_,
 including support for caching of results.
 """
 import itertools
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -1150,7 +1150,7 @@ class tzical(object):
         A file or stream in iCalendar format, which should be UTF-8 encoded
         with CRLF endings.
 
-    .. _`RFC 2445`: https://www.ietf.org/rfc/rfc2445.txt
+    .. _`RFC 2445`: https://tools.ietf.org/html/rfc5545
     """
     def __init__(self, fileobj):
         global rrule


pganssle:
@mariocj89 Can you update the sections as well? The RRULE section is 3.3.10 in RFC 5545, but it's 4.3.10 in RFC 2445. See the issue #318.

mariocj89:
Will be resubmitted in the evening sprint :)

Issue #519, opened 2017-11-12
Setup builds so xfail doesn't count towards code coverage


pganssle:
We've lately added quite a few `xfail` tests as a sort of "to do", but I realized that this may cause the `codecov` metrics to be less meaningful than one might hope. The reason being that `xfail` tests can fail for *any* reason, including that they went down some random buggy code path whose *proper* behavior is not actually covered, and that buggy code path threw an actual bug we should be aware of. As such, I'd rather code executed as the result of `xfail` tests *not* contribute to the code coverage. (They should still be *run* because I'd like to know if they turn from xfail to xpass).

If there's not just some standard `pytest-coverage` or `coverage` option that does this, I think it's simple enough to set up by just running `pytest` with all `xfail` tests skipped with `coverage` on, then running it a second time with coverage off and *only* `xfail` tests included.

pganssle:
Fixed in #563 (though I haven't checked this...)

PR #549, opened 2017-12-06
move CI to tox

gaborbernat:


gaborbernat:
Resolves #519.

pganssle:
@gaborbernat Hrm.. This has a similar problem to one of my earlier configurations - tests are not covered. I thought we were going to set it up so that the regression was that we're not doing `changedir` anymore?

I'll figure it out later I guess.

gaborbernat:
I explicitly excluded tests from coverage, I can enable it if you want?

pganssle:
@gaborbernat Yes, I've caught several bugs when I realized that parts of my test suite was not getting executed.

gaborbernat:
added back via config flag in ``tox.ini`` in #568 (yeah was lazy to create another PR for it, sorry)

PR #552, opened 2017-12-06
Adds python_require

### LINKED ISSUE #537: Add python_requires to setup.py
Right now we're just using the trove classifiers, so I think people may end up doing some complicated pinning, which is less than ideal (see #536 and #526). What we need is the `Requires-Python:` metadata in pip. I think this is added through `setup.py`. Here are some examples of people using it:

- [pytest](https://github.com/pytest-dev/pytest/blob/master/setup.py#L86)
- [ipython](https://github.com/ipython/ipython/blob/master/setup.py#L235)

@carreau gave [this talk at PyBay](https://youtu.be/3i6n1RwqQCo?t=660) (link goes to the relevant time - 11 minutes in) where he discusses the `python_requires` option ([slides are here](https://speakerdeck.com/pybay/2017-building-bridges-stopping-python-2-without-damages)), probably best to listen to what he's talking about and avoid or document the pitfalls here.

jakec-github:
This closes #537 

Adds python_require to setup() in setup.py. Only works for users installing with pip 9 or above. Will warn if setuptools version is <=24.3.

Also help from Stefan Bonchev

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -31,6 +31,7 @@ Apache 2.0 and BSD 3-clause. In the list below, anyone whose name is marked with
 - Holger Joukl <holger.joukl@MASKED> (gh: @hjoukl)
 - Igor <mrigor83@MASKED>
 - Ionuț Ciocîrlan <jdxlark@MASKED>
+- Jake Chorley (gh: @jakec-github)
 - Jan Studený <jendas1@MASKED>
 - Jitesh <jitesh@MASKED>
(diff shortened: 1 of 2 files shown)


pganssle:
The `setuptools.__version__` check should use `distutils.version.LooseVersion` (or `StrictVersion`):

Here's a link to explain how it works: https://stackoverflow.com/questions/11887762/compare-version-strings-in-python

jakec-github:
I used LooseVersion to compare setuptools versions since it should work a little more reliably

Issue #541, opened 2017-12-05
Prepocessing results of timelex.split


jbrockmendel:
Looking at the test case `timestr = '5:50 A.M. on June 13, 1990'` which incorrect identifies the 'M' in 'A.M.' as a tzname, I'm thinking it might make sense to avoid this particular problem within timelex.split so that "A.M." is identified as a single token.  A few similar cases come to mind: "P.M.", "A.D.", "B.C.".  Thoughts?

jbrockmendel:
@pganssle I'm thinking about implementing something for this, want to run it by you first since it may run up against your aversion to hard-coding strings.  There are two cases I'd like to catch at tokenization time:

1) [AP].M.
2) "[A-Za-z]/[A-Za-z]" --> retain "US/Eastern" instead of ["US", "/", "Eastern"].

Thoughts?

pganssle:
@jbrockmendel The problem with this and many of your other attempts at structural changes to the parser is that they are piecemeal and tweak around the edges, which is exactly what I *do not want*. I will need some time to learn more about the general theory of parsing before I'm ready to make any decisions about how the parser will change, unfortunately.

Unless you want to present a fully-featured parser as a *fait accompli* with an interface I would be happy with, I imagine most of these suggestions and PRs will languish for a while. I would probably not recommend this (unless you prefer to just write a separate parser for reasons other than getting it merged into `dateutil`), as we seem to have markedly different sensibilities about what makes for a good API.

I am, generally speaking, wary of doing parsing in the lexer. I think part of the problem is that `timelex` seems to actually just be a tokenizer, not a proper lexer (e.g. a tokenizer that attaches some additional context to the tokens). I don't understand the various techniques of lexing and parsing and when to apply them well enough to know what the right solution is here.

Also, what exactly is the point of pre-processing these like this? Both the tokens and the separators are available to the parser, and it frequently re-combines them as necessary. Why is this case special such that we should violate this separation of concerns?

pganssle:
>I think part of the problem is that timelex seems to actually just be a tokenizer, not a proper lexer (e.g. a tokenizer that attaches some additional context to the tokens). 

[...]

PR #554, opened 2017-12-06
fix bug with incorrectly picking out 1-character tznames

jbrockmendel:
I think with this fix, the errors exposed by #540 should be fixed.

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -1045,6 +1045,9 @@ class parser(object):
             (res.second, res.microsecond) = self._parsems(value_repr)
 
     def _could_be_tzname(self, hour, tzname, tzoffset, token):
+        if len(token) <= 1 and token != 'Z':
+            # see GH#540
+            return False
         return (hour is not None and
                 tzname is None and
                 tzoffset is None and
--- a/dateutil/test/test_internals.py
+++ b/dateutil/test/test_internals.py
@@ -12,9 +12,21 @@ import sys
 
 import pytest
 
-from dateutil.parser._parser import _ymd
+from dateutil.parser._parser import _ymd, parser
 
 IS_PY32 = sys.version_info[0:2] == (3, 2)
+
+
+class TestParserPrivate(object):
+    def test_single_character_tzname(self):
+        # See GH#540
+        dstr = '5:50 A.M. on June 13, 1990'
+        res = parser()._parse(dstr)[0]
+        assert res.tzname is None
+
+        dstr = 'Jan 29, 1945 14:45 AM I going to see you there?'
+        res = parser()._parse(dstr, fuzzy=True)[0]
+        assert res.tzname is None
 
 
 class TestYMD(unittest.TestCase):


pganssle:
Hm... This is *probably* OK, except that `Z` is actually an artifact of a specific way of specifying time zone offsets that's broader and always uses a single character.

See [Alpha Time](https://www.timeanddate.com/time/zones/a) from [this list](https://www.timeanddate.com/time/zones/). I'm not entirely sure I want to foreclose these as a general rule. Plus, eventually we'll want to allow `tzinfo` strategies to somehow advertise the valid time zone name ranges, and we don't want these other things to break if you supply some 1-length time zone abbreviations. Probably best to do this on the tokenization step maybe?

jbrockmendel:
> Probably best to do this on the tokenization step maybe?

I've been giving this some thought (#541); catching e.g. "A.M." in the tokenization step would be really nice, not sure it would handle the other broken case here.

Also if in the future we want to identify e.g. "US/Eastern" as a tzname, that may be easier to catch pre-tokenization since that would become `["US", "/", "Eastern"]`.

> eventually we'll want to allow tzinfo strategies to somehow advertise the valid time zone name ranges, and we don't want these other things to break if you supply some 1-length time zone abbreviations

That's reasonable.  Maybe if we pass `tzinfos` through so that the appropriate check can be done in `_could_be_tzname `?

pganssle:
@jbrockmendel Maybe take a look at how that last part is done in #227, since AFAICT that actually does work.

I do want to apologize that mostly my attitude towards this parser stuff is to sorta reject a lot of these incremental changes. It's a weird situation because I have only a general sense about general techniques in parsing, and I think this sort of heavily contextualized parsing with ambiguous constraints and such is one of the harder kinds of parsing (compared to well-defined grammars). I haven't found the time to like.. read a book on parsing, so my answer to a lot of these strategy proposals is more or less, "I'm not sure how I'll feel about that *after* I have a reasonable framework to understand parsing problems in general, so it's probably not worth commenting on now."

jbrockmendel:
> my attitude towards this parser stuff is to sorta reject a lot of these incremental changes

I totally understand where you're coming from, but appreciate that you get that it can be frustrating on this end.

#227 looks like a reasonable approach (though parts of it would be easier if we caught some strings at tokenization-time).  Is it high enough in the queue that I should wait for it to fix this bug?

In it's current condition, this PR basically trades one bug for another.  It avoids incorrectly identifying 1-character timezones, but will incorrectly reject Alpha Time in the (probably rare) case that a user chooses that.  I think the thing to do is to pass `tzinfos` to `_parse` and then to `_could_be_tzname` and Do It Right.  Thoughts?

pganssle:
As a general rule, when the bugs are both at least somewhat rare, I tend to put a good deal of weight on the status quo, but also reversibility (i.e. I'm more comfortable changing the behavior when you can get the old behavior by jumping through some simple hoops).

Considering that I don't see any obvious seam where a user can reverse this behavior and we have some general, vague plans to fix this the *right* way later, I don't see any reason to make this particular trade-off.

jbrockmendel:
>  I tend to put a good deal of weight on the status quo

Absolutely agree.  I'm currently working on updating this PR to fix this bug without introducing the new one.

dateutil/test/test_internals.py
```diff
@@ -12,11 +12,40 @@
...
+        dstr = '2017-12-07 10:27:15B'
+        tzinfos = {'B': dateutil.tz.tzoffset('Beta', 120)}
+        res = parser().parse(dstr, tzinfos=tzinfos)
+        assert res.tzinfo is tzinfos['B']
+
+    @pytest.mark.xfail
+    def test_alpha_tzname_ampm(self):
+        # If the Alpha-style timezone is "A" it gets incorrectly
+        # identified as part of AM/PM
+        dstr = '2017-12-07 10:27:15A'
```
pganssle: Hm.. I would guess that `10:27:15A` means AM/PM more often than it means `Alpha`, though I guess they did explicitly set the `tzinfos`...

jbrockmendel: Yah I'm not 100% sure what to do in this case, leaning towards agreeing on the "well if they were explicit...".  Also I think "AM" and "A.M." are way more common than "A" so that should be reasonably safe.

dateutil/parser/_parser.py
```diff
@@ -1044,12 +1045,26 @@ def _assign_hms(self, res, value_repr, hms):
...
+        elif isinstance(tzinfos, collections.Callable):
+            # We can't validate at this point so we let it through and validate
+            # when constructing the final return value
+            return True
+        else:
+            return token in tzinfos or (hour is not None and
+                    tzname is None and
+                    tzoffset is None and
+                    len(token) <= 5 and
+                    all(x in string.ascii_uppercase for x in token))
```
jbrockmendel: Not sure if this is the right thing to do.  If the user _does_ pass a `tzinfos`, should we be strict in this branch and only accept `token in tzinfos` (and maybe UTC, GMT, Z or the parserinfo equivalents)?  Should the `token in tzinfo` check also require the `hour is not None and tzname is None and tzoffset is None`?

PR #563, opened 2017-12-06
xfails do not count towards code coverage

gaborbernat:
but still fail the test suite if they start passing


gaborbernat:
resolves #519

Issue #339, opened 2016-12-31
Add resolve_imaginary function to `tz`


pganssle:
It occurs to me from looking at crsmithdev/arrow#72 that it might be useful to have a `resolve_imaginary` function built in to `dateutil`, similar to this one that I wrote as an example:

```python
from dateutil.tz import datetime_exists
from datetime import timedelta

def resolve_imaginary(dt):
    if dt.tzinfo is not None and not datetime_exists(dt):
        # You need to shift this by the difference between the current offset
        # and the previous one. Historically, there has never been a time zone
        # shift of more than 24 hours, nor has there been two changes to a time
        # zone within 24 hours, so if we take the "wall time" - 24 hours, we
        # should be OK, barring any insane time zone behavior in the future.
        curr_offset = dt.utcoffset()
        old_offset = (dt - timedelta(hours=24)).utcoffset()

        dt += curr_offset - old_offset

    return dt
```

I'm fairly certain that an imaginary date you always want to "slide forward" - it will never be that you accidentally wanted to go back.

We have a particular advantage here because, like with `datetime_ambiguous` and `datetime_exists`, we can define an API in the `tzinfo` objects themselves that lets us avoid the "look back 24 hours" heuristic if at all possible. This works for all known transitions in the Olson database, but we could get a sudden breakage if, for example, a specific zone changes their DST and then changes back within 24 hours, or something like that.

PR #565, opened 2017-12-06
London Sprint - imaginaryDateTest

Cheukting:
London Sprint #339 

Pushing function resolve_imaginary in tz, created 4 test cases which 3 of them pass. the only one that fail is testMonroviaForward which is dealing with + 44mins 30seconds time zone changes. Further testing suggest that utcoffset ignore the 30seconds so it 0:30 on 7 Jan 1972 will become 1:14, not 1:14:30 on the same day.

### DIFF
--- a/.cache/v/cache/lastfailed
+++ b/.cache/v/cache/lastfailed
@@ -1,0 +1,6 @@
+{
+  "dateutil/test/test_tz.py::TzWinTest::testTzResNameFromString": true,
+  "dateutil/test/test_tz.py::TzWinTest::testTzwinName": true,
+  "dateutil/test/test_tz.py::imaginaryDateTest::testMonroviaForward": true,
+  "dateutil/test/test_utils.py": true
+}
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -2283,8 +2283,7 @@ class imaginaryDateTest(unittest.TestCase):
         self.assertEqual(dt,datetime(1995, 1, 2, 2, 30, tzinfo=timezone))
         
     def testMonroviaForward(self):
-        #import pdb; pdb.set_trace()
         timezone = tz.gettz('Africa/Monrovia')
-        dt = datetime(1972, 1, 7, 0, 30, 0, tzinfo=timezone)
+        dt = datetime(1972, 1, 7, hour = 0, minute = 30, second = 0, tzinfo=timezone)
         dt = tz.resolve_imaginary(dt)
-        self.assertEqual(dt,datetime(1972, 1, 7, 1, 14, 30, tzinfo=timezone))
+        self.assertEqual(dt,datetime(1972, 1, 7, hour = 1, minute =  14,  second = 30, tzinfo=timezone))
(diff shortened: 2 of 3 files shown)


.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,6 @@
+{
+  "dateutil/test/test_tz.py::TzWinTest::testTzResNameFromString": true,
+  "dateutil/test/test_tz.py::TzWinTest::testTzwinName": true,
+  "dateutil/test/test_tz.py::imaginaryDateTest::testMonroviaForward": true,
+  "dateutil/test/test_utils.py": true
+}
```
jbrockmendel: These cache files generally don’t belong in the commit.  Is this here intentionally?

dateutil/tz/tz.py
```diff
@@ -1531,4 +1531,21 @@ def __enter__(self):
     def __exit__(*args, **kwargs):
         pass
 
+def resolve_imaginary(dt):
+    import pdb;pdb.set_trace()
```
jbrockmendel: Is the Pdb needed outside of testing?

Cheukting: Sorry I was finishing in a rush, should be for testing only, I can tidy it up and push it again.

.cache/v/cache/lastfailed
Cheukting: I think they are automatically generated during testing, can be deleted. I will have to tidy things up and push it again.

pganssle:
@Cheukting FYI, it's generally not a great idea to work in the master branch of your fork. It's fine in this case, but it will make it harder to work in the future.

Cheukting:
Thanks @pganssle for the tip. I am new in term of developing packages so still need to learn what's the best practice. Will create a new branch next time. :smiley:

dateutil/test/test_tz.py
```diff
@@ -2262,3 +2262,28 @@ def testExitFold(self):
...
         self.assertEqual(getattr(dt, 'fold', 0), 0)
+
+class imaginaryDateTest(unittest.TestCase):
+    def testCanberraForward(self):
+        timezone = tz.gettz('Australia/Canberra')
+        dt = datetime(2018, 10, 7, 2, 30, tzinfo=timezone)
+        dt = tz.resolve_imaginary(dt)
+        self.assertEqual(dt,datetime(2018, 10, 7, 3, 30, tzinfo=timezone))
+        
+    def testSydneyBack(self):
```
pganssle: I didn't realize this originally, but `datetime(2018, 4, 1, 3, 30)` is very close to, but not actually an ambiguous time. It might be best to use a more obviously unambiguous time like `datetime(2018, 6, 2, 12, 0)`.

Also maybe rename this to be like `testSydneyExisting` and/or add a comment like `# Tests that resolve_imaginary has no effect on normal datetimes`.

pganssle:
@Cheukting No matter what you end up doing, can you please add yourself to the AUTHORS.md file? Even if you make no other changes, your work on this so far deserves recognition.

pganssle:
Consolidated into #607.

PR #568, opened 2017-12-07
add document generation to tox

gaborbernat:
and trigger automatically at CI (also check links in documentation to avoid having dangling pointers to the internet)

### DIFF
--- a/dateutil/zoneinfo/__init__.py
+++ b/dateutil/zoneinfo/__init__.py
@@ -8,7 +8,7 @@ from io import BytesIO
 
 from dateutil.tz import tzfile as _tzfile
 
-__all__ = ["get_zonefile_instance", "gettz", "gettz_db_metadata", "rebuild"]
+__all__ = ["get_zonefile_instance", "gettz", "gettz_db_metadata"]
 
 ZONEFILENAME = "dateutil-zoneinfo.tar.gz"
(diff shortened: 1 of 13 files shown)


dateutil/zoneinfo/__init__.py
```diff
@@ -8,7 +8,7 @@
 
 from dateutil.tz import tzfile as _tzfile
 
-__all__ = ["get_zonefile_instance", "gettz", "gettz_db_metadata", "rebuild"]
+__all__ = ["get_zonefile_instance", "gettz", "gettz_db_metadata"]
```
pganssle: I don't think we can change the public interface like this...

pganssle: Hm, actually `rebuild` is a separate file. Yeah, I see how this works. This is fine.

gaborbernat: the problem here is that this import does not work, the Sphinx could not load it, as no import is there. how is this rebuild suppose to exist?

pganssle:
This is awesome, thanks @gaborbernat. I'll merge once CI passes.

gaborbernat:
``tox -e docs`` will now build to ``.tox\docs_out`` the documentation (on any platform) 👍

PR #569, opened 2017-12-07
add readthedocs badge, PyPi versions and license

gaborbernat:
see https://github.com/gaborbernat/dateutil/blob/badge/README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,26 +1,43 @@
 dateutil - powerful extensions to datetime
 ==========================================
 
-.. image:: https://img.shields.io/pypi/v/python-dateutil.svg?style=flat-square
+|pypi| |support| |licence| 
+
+|gitter| |readthedocs|
+
+|travis| |appveyor| |coverage|
+
+.. |pypi| image:: https://img.shields.io/pypi/v/python-dateutil.svg?style=flat-square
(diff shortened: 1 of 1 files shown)


pganssle:
Hm.. License is not accurate because it's pulled from the PyPI metadata. Not a permanent situation, though.

gaborbernat:
Yeah should self-fix once the PyPI one is fixed.

gaborbernat:
@pganssle anything to do with this one?

pganssle:
Hm... It's generally fine, but I'd like it if I could get the badges to be on two lines, rather than just overflowing from one line onto the next. I tried inserting an explicit linebreak and it looks quite ugly, unfortunately.

gaborbernat:
@pganssle shall we close this then?

pganssle:
@gaborbernat I'm going to keep it open a bit longer as a reminder to try and figure out how to do layout on RST so that we *can* include more badges without it looking messy.

gaborbernat:
Fixed the merge conflict and gave it a go through a different approach :thinking: 

screenshot-github com-2018 03 29-13-09-27

The only other way would be putting it inside a table I think.

pganssle:
@gaborbernat Yeah, that looks a lot better. Thanks for doing that.

Issue #427, opened 2017-07-26
dateutil parser rounding or correctness problem?


dddmello:
Was running some tests and the below code:

    from dateutil import parser as dparser
    print((dparser.parse('5.6h') - dparser.parse('0s')).total_seconds())

prints:

20100.0

I would expect it to print 20160.0 (off my 60 seconds).

Specifically it looks like:

>>> dparser.parse('5.6h')
datetime.datetime(2017, 7, 26, 5, 35)

Should actually be returning datetime.datetime(2017, 7, 26, 5, 36)

I see this behavior with version 2.5.3 - wondering if this is expected or if there is a rounding/truncation bug somewhere?

jbrockmendel:
Looks like this is a floating point issue:

```
>>> 5.6 % 1
0.5999999999999996
>>> int(60*(5.6 % 1))
35
```

I'll put in a PR to fix this -- unless you want beat me to it.

jbrockmendel:
> Was running some tests

If you have some real-life test cases we can add to the test suite, those would be helpful.

pganssle:
I suppose the rounding issue should be fixed, though if this is indeed the issue, it seems to imply that "5.6h" only parses to hours + minutes, not hours + minutes + seconds + microseconds. I'm not really sure who is using "5.6h" as a time format rather than an interval format, but I'm somewhat surprised that to the extent we're supporting it, we truncate at minutes.

That said, this is not the right use of `dateutil.parser.parse`. "5.6h" and "0s" are intervals, not times, and unless you explicitly specify the `default`, there might be subtle bugs during certain transitions (e.g. if run at midnight), if you use them as such, since what's happening is you're essentially parsing "5.6h" into "today at 5AM + 0.6 hours" and "0s" as "today at midnight", then subtracting the two to get an actual interval. This is a *horrible* abuse of a a datetime parser.

I think the proper way to handle this would be to implement a separate interval parser that returns `timedelta` and/or `relativedelta` objects.l See #295 and #351. There are probably other issues that also boil down to "we should have an interval parser."

jbrockmendel:
I've got a "fix" for this that parses float-like strings more carefully.  Should I bother making a PR or is this format going to be handled elsewhere?

pganssle:
@jbrockmendel I think we can fix the rounding problem. This format is in the test suite so it's something we support. I'm just horrified by how it's being used here, so we should probably provide a proper alternative.

m-dz:
[...]

PR #570, opened 2017-12-07
Fix issue 427

m-dz:
Fixes issue #427 with all tests passing

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -42,6 +42,8 @@ from io import StringIO
 import six
 from six import binary_type, integer_types, text_type
 
+from decimal import Decimal
+
 from .. import relativedelta
 from .. import tz
 
@@ -873,7 +875,7 @@ class parser(object):
     def _parse_numeric_token(self, tokens, idx, info, ymd, res, fuzzy):
         # Token is a number
         value_repr = tokens[idx]
-        value = float(value_repr)
+        value = Decimal(value_repr)
         len_li = len(value_repr)
 
         len_l = len(tokens)
@@ -932,7 +934,7 @@ class parser(object):
         elif idx + 2 < len_l and tokens[idx + 1] == ':':
             # HH:MM[:SS[.ss]]
             res.hour = int(value)
-            value = float(tokens[idx + 2])  # TODO: try/except for this?
+            value = Decimal(tokens[idx + 2])  # TODO: try/except for this?
             (res.minute, res.second) = self._parse_min_sec(value)
 
             if idx + 4 < len_l and tokens[idx + 3] == ':':
@@ -1032,7 +1034,9 @@ class parser(object):
         return hms_idx
 
     def _assign_hms(self, res, value_repr, hms):
-        value = float(value_repr)
+        # See GH issue #427, fixing float rounding
+        value = Decimal(value_repr)
+
         if hms == 0:
             # Hour
             res.hour = int(value)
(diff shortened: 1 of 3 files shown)


pganssle:
@m-dz Do you want to add yourself to the Authors file, or you can just comment here with the right name / e-mail (optional) to use?

dateutil/parser/_parser.py
```diff
@@ -42,6 +42,8 @@
 import six
 from six import binary_type, integer_types, text_type
 
+from decimal import Decimal, getcontext
```
pganssle: `getcontext` is not used

dateutil/parser/_parser.py
```diff
@@ -1091,6 +1095,12 @@ def _parse_min_sec(self, value):
         # TODO: Every usage of this function sets res.second to the return
         # value. Are there any cases where second will be returned as None and
         # we *dont* want to set res.second = None?
+
+        # See GH issue #427, fixing float rounding
+        # Conversion here might be redundant as value is already passed as Decimal()
+        # but _parse_min_sec() is also called in _parse_numeric_token where it is still a float().
```
pganssle: If it's already `float`, you're not making anything better by converting it to `Decimal`.

```python
>>> from decimal import Decimal
>>> Decimal(5.6)
Decimal('5.5999999999999996447286321199499070644378662109375')
```

The problem is to avoid converting it to a `float` in the first place.

m-dz:
@pganssle , will do it, thanks!

dateutil/test/test_parser.py
```diff
@@ -917,6 +917,18 @@ def test_pre_12_year_same_month(self):
         dtstr = '0003-03-04'
         assert parse(dtstr) == datetime(3, 3, 4)
 
+    def test_rounding_float_like_strings(self):
```
pganssle: This is a mix of `pytest` and `unittest` style tests. For the `unittest` style tests, you use `self.assertEqual` and the name would be `testRoundingFloatLikeStrings`. I'm fine with you de-indenting it so it's not in a class and use `pytest`, like so:

```python
@pytest.mark.parametrize('dtstr,dt', [
    ('5.6h', datetime(2009, 9, 25, 5, 36)),
    ('5.6m', datetime(2009, 9, 25, 0, 5, 36)),
    ('5.6s', datetime(2009, 9, 25, 0, 0, 5, 600000))
]
def test_rounding_floatlike_strings(dtstr, dt):
    assert parse(dtstr, default=datetime(2003, 9, 25)) == dt
```

Though actually `'5.6s'` never had this problem (that's fine, but it might be worth a comment).

dateutil/parser/_parser.py
pganssle: If this is not causing any problems just drop the `Decimal` call.

m-dz:
@pganssle , just saw your comments, will have a look in the afternoon.

pganssle:
@m-dz I think where you put it is fine. Thanks for the PR and update!

I'm not sure that these other `Decimal` calls are *necessary* per se, but since `float` is probably the wrong data structure to use in this case anyway, and from brief profiling of the constructor it doesn't seem significantly slower to construct a `Decimal` than a `float`, I'm inclined to come down on the side of correctness.

m-dz:
@pganssle , thanks for help and useful comments! I have changed `float()` to `Decimal()` everywhere, where the modulo 1 operation was and I agree they might be redundant, but I can still see someone using e.g. `'0h5.6m'` as an input.

PR #573, opened 2017-12-08
Removed support for YYYYMM

### LINKED ISSUE #546: Drop support for YYYYMM in isoparser
According to Wikipedia [YYYYMM](https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates) is explicitly *not* supported in ISO-8601, so we should drop support for it. Hopefully this does not add undue complication into the algorithm.

See #489 for initial isoparse implementation.

kirit93:
Fixes #546

### DIFF
--- a/dateutil/parser/isoparser.py
+++ b/dateutil/parser/isoparser.py
@@ -206,7 +206,10 @@ class isoparser(object):
         pos += 2
 
         if pos >= len_str:
-            return components, pos
+            if has_sep:
+                return components, pos
+            else:
+                raise ValueError('Invalid ISO format')
 
         if has_sep:
             if dt_str[pos:pos + 1] != self._DATE_SEP:
--- a/dateutil/test/test_isoparser.py
+++ b/dateutil/test/test_isoparser.py
@@ -54,9 +54,8 @@ def test_year_only(dt):
 
 DATES += [datetime(2000, 2, 1), datetime(2017, 4, 1)]
 @pytest.mark.parametrize('dt', tuple(DATES))
-@pytest.mark.parametrize('fmt',
-    ['%Y%m', '%Y-%m'])
-def test_year_month(dt, fmt):
+def test_year_month(dt):
+    fmt   = '%Y-%m'
     dtstr = dt.strftime(fmt)
 
     assert isoparse(dtstr) == dt
@@ -228,6 +227,7 @@ def test_bytes(isostr, dt):
     ('20120425C012500', ValueError),            # Wrong time separator
     ('2001-1', ValueError),                     # YYYY-M not valid
     ('2012-04-9', ValueError),                  # YYYY-MM-D not valid
+    ('201204', ValueError),                     # YYYYMM not valid
     ('20120411T03:30+', ValueError),            # Time zone too short
     ('20120411T03:30+1234567', ValueError),     # Time zone too long
     ('20120411T03:30-25:40', ValueError),       # Time zone invalid
@@ -322,16 +322,15 @@ def test_parse_tzstr_fails(tzstr, exception):
 def __make_date_examples():
     dates_no_day = [
         date(1999, 12, 1),
-        date(2016, 2, 1),
+        date(2016, 2, 1)
     ]
 
     if six.PY3:
         # strftime does not support dates before 1900 in Python 2
         dates_no_day.append(date(1000, 11, 1))
 
-    date_no_day_fmts = ('%Y%m', '%Y-%m')
-
-    o = it.product(dates_no_day, date_no_day_fmts)
+    # Only one supported format for dates with no day
+    o = zip(dates_no_day, it.repeat('%Y-%m'))
 
     dates_w_day = [
         date(1969, 12, 31),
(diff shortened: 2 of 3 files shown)


dateutil/parser/isoparser.py
```diff
@@ -206,7 +206,7 @@ def _parse_isodate_common(self, dt_str):
         pos += 2
 
         if pos >= len_str:
-            return components, pos
+            raise ValueError('YYYYMM or YYYY-MM is an invalid ISO 8601 format')
```
pganssle: `YYYY-MM` is very valid. It is *only* `YYYYMM` that is not valid (because it creates an ambiguity between ISO-8601 dates and non-ISO8601 `YYMMDD` triplets.

dateutil/test/test_isoparser.py
```diff
@@ -342,24 +335,8 @@ def __make_date_examples():
 
     dates_w_day_fmts = ('%Y%m%d', '%Y-%m-%d')
     o = it.chain(o, it.product(dates_w_day, dates_w_day_fmts))
-
     return list(o)
 
-
-@pytest.mark.parametrize('d,dt_fmt', __make_date_examples())
```
pganssle: Why was this test removed?

dateutil/test/test_isoparser.py
```diff
@@ -368,6 +345,7 @@ def test_parse_isodate(d, dt_fmt, as_bytes):
     ('2013-02-29', ValueError),                 # Not a leap year
     ('2014/12/03', ValueError),                 # Wrong separators
     ('2014-04-19T', ValueError),                # Unknown components
+    ('2012-02', ValueError)                     # Invalid ISO 8601 format
```
pganssle: This is valid, only `201202` is invalid.

dateutil/test/test_isoparser.py
kirit93: I removed this test because it's testing date strings in the format "YYYY-MM" and I thought that format was invalid as well. I'll add it back.
Just to make sure I've understood the problem correctly, date strings in the format "YYYYMM" should raise a ```ValueError``` right?

pganssle: Yes.

dateutil/test/test_isoparser.py
```diff
@@ -55,7 +55,7 @@ def test_year_only(dt):
 DATES += [datetime(2000, 2, 1), datetime(2017, 4, 1)]
 @pytest.mark.parametrize('dt', tuple(DATES))
 @pytest.mark.parametrize('fmt',
-    ['%Y%m', '%Y-%m'])
+    ['%Y-%m', '%Y-%m'])
```
pganssle: I think this just repeats the same test twice. Since there's only one valid year-month format, I think you can actually move `fmt` into the function definition.

dateutil/test/test_isoparser.py
```diff
@@ -322,14 +323,14 @@ def test_parse_tzstr_fails(tzstr, exception):
...
-        date(2016, 2, 1),
+        date(2016, 2, 1)
     ]
 
     if six.PY3:
         # strftime does not support dates before 1900 in Python 2
         dates_no_day.append(date(1000, 11, 1))
 
-    date_no_day_fmts = ('%Y%m', '%Y-%m')
+    date_no_day_fmts = ['%Y-%m']
```
pganssle: Why the change from tuple to list? I'm not married to using a tuple, just curious. (Though also this might be a situation where we can drop this parameter entirely, I'd have to read the tests more thoroughly to see).

kirit93: In the ```__make_data_examples()``` function ```o = it.product(dates_no_day, date_no_day_fmts)``` is used to create the examples, and ```date_no_day_fmts``` is a tuple with two formats we want the date to be in. However now we need only one format so a list with a singular element is used. This is similar to line 383.

dateutil/test/test_isoparser.py
kirit93: Done.

dateutil/test/test_isoparser.py
pganssle: @kirit93 I don't understand why this would make you switch to a list, though.

```python
>>>import itertools as it
>>>from datetime import date
>>>dates_no_day = [date(1999, 12, 1), date(2016, 2, 1)]
>>>dates_no_date_fmts = ('%Y-%m',)
>>>o = it.product(dates_no_day, dates_no_date_fmts)
>>>print(list(o))
[(datetime.date(1999, 12, 1), '%Y-%m'), (datetime.date(2016, 2, 1), '%Y-%m')]

```

pganssle:
@kirit93 I just re-wrote that test anyway, since `product` is not really the appropriate itertool in this case, so tuple vs. list doesn't matter.

I've squashed a bunch of your intermediate commits and added you to the authors file. Please review the `AUTHORS.md` file before I merge and let me know if you'd prefer to change how you are credited.

dateutil/test/test_isoparser.py
kirit93: You're right, I could have left it as a tuple. I just changed it to be a list because that's how it was done in the test on line 383, no other reason.

kirit93:
```AUTHORS.md``` looks good. Thanks!

PR #574, opened 2017-12-08
Adds tz examples to tz.py

jakec-github:
This copies all the tz examples into tz.py. Note they are still in the examples section

As long as this is what you were after I will copy the rest over in a separate pull request this weekend

### DIFF
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -38,7 +38,7 @@ class tzutc(datetime.tzinfo):
     """
     This is a tzinfo object that represents the UTC time zone.
 
-    **Examples:**
+    **Example:**
 
     .. doctest::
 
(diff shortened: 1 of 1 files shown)


dateutil/tz/tz.py
```diff
@@ -189,6 +208,22 @@ def __repr__(self):
...
     A :class:`tzinfo` subclass built around the ``time`` timezone functions.
+
+    .. doctest:: tzlocal
+
+        >>> from datetime import *
+        >>> from dateutil.tz import *
+
+        >>> datetime.now(tzlocal())
+        datetime.datetime(2003, 9, 27, 10, 1, 43, 673605,
+                  tzinfo=tzlocal())
```
jbrockmendel: flake8 would complain about indentation here and * imports

pganssle: True, though to be fair he's literally just copy-pasting from [here](https://dateutil.readthedocs.io/en/latest/examples.html#tzoffset-examples). Probably wouldn't be a terrible idea to go through the examples and fix them.

The problem now, though, is that we started literally copying stuff from the examples page into the individual module pages, because I didn't want to break old URLs. I realize now that the reason I didn't do that in the first place is that now we have to maintain the example documentation in *two places*.

I'm inclined to accept this PR since it's not breaking anything *new*, but to put a halt to these documentation moves until we work out a way to transclude examples.

pganssle:
I'll have to pull this and build it to see how it looks, some of these docstrings look a bit long for my tastes, in which case it might be best to do the "examples all at the bottom of the page" thing.

Also, per my comments to @jbrockmendel, I'm realizing that we need to figure out how, in rst, to write the docs in the docstrings and include them in the examples. I think it's pretty easy but anything more than basic formatting and I have trouble with it in RST.

pganssle:
Closing in favor of #671

Issue #529, opened 2017-11-16
Cut down on Appveyor builds


pganssle:
Currently this is the Appveyor build matrix:

```yaml
  matrix:
    - PYTHON: "C:/Python27"
    - PYTHON: "C:/Python27-x64"
    - PYTHON: "C:/Python33"
    - PYTHON: "C:/Python33-x64"
    - PYTHON: "C:/Python34"
    - PYTHON: "C:/Python34-x64"
    - PYTHON: "C:/Python35"
    - PYTHON: "C:/Python35-x64"
    - PYTHON: "C:/Python36"
    - PYTHON: "C:/Python36-x64"
```

I'm not sure I can recall a time when we had a failure on x64 that didn't also fail on x86, and Appveyor is often the slow step in the CI (I think they don't do concurrent builds). I'm thinking we should probably cut down and re-organize the build matrix, like this:

```yaml
  matrix:
    - PYTHON: "C:/Python27-x64"
    - PYTHON: "C:/Python36-x64"

    - PYTHON: "C:/Python27"
    - PYTHON: "C:/Python33"
    - PYTHON: "C:/Python34"
    - PYTHON: "C:/Python35"
    - PYTHON: "C:/Python36"
```

We'd be testing on all supported versions (except 3.2, which we've droped from these builds for whatever reason ages ago) on x86, and only test the latest two builds on x64, but the latest two x64 builds would be executed *first*.

The other option is to test x64 on all versions and only test the latest two with x86. The downside there is that I think 1. x86 is more likely to have problems than x64 and 2. you're *more likely* to be using x86 if you're also using an old version of Python.

I imagine when we add a binary component, we'll want to turn the full matrix back on again at least temporarily, since that's way more likely to have architecture + version corner cases.

kirit93:
A simple fix for this could be to exclude tests we don't want. Eg. if we don't want Python33-x64 we can add 
```
matrix:
    exclude:
        - platform: x64
           PTYHON_VERSION: 33
```
And so on for the configurations we don't want to test.
Appveyor does concurrent builds based on the type of account you have. For the basic and professional it's just 1, but the premium account gives you 2 concurrent build and if there's an option for unlimited concurrent builds but that's around $300 per month.

pganssle:
@kirit93 It's not really a question of implementation so much as whether these are actually *necessary*.

[...]

PR #579, opened 2017-12-10
Changed appveyor.yml to reduce tests

kirit93:
Related to issue #529 
@gaborbernat Please let me know if the changes I've made are what is expected.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,14 +1,20 @@
 build: false
 environment:
   matrix:
-    - PYTHON_VERSION: 27
-    - PYTHON_VERSION: 33
-    - PYTHON_VERSION: 34
-    - PYTHON_VERSION: 35
-    - PYTHON_VERSION: 36
-platform:
-  - x86
-  - x64
+    - PLATFORM : x86
+      PYTHON_VERSION: 27
+    - PLATFORM : x86
+      PYTHON_VERSION: 36
+    - PLATFORM : x64
+      PYTHON_VERSION: 27
+    - PLATFORM : x64
+      PYTHON_VERSION: 33
+    - PLATFORM : x64
+      PYTHON_VERSION: 34
+    - PLATFORM : x64
+      PYTHON_VERSION: 35
+    - PLATFORM : x64
+      PYTHON_VERSION: 36
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


pganssle:
Hmm.. One thing to consider here is the order. I wonder if there's any way to get statistics on which environments fail most frequently (preferably where builds where *all* platforms failed and builds that did not finish are discarded).

My instinct is to order this like this:

- python 2.7 x64
- python 3.6 x64
-python 3.3 x64
- python 3.4 x64
- python 3.5 x64
- python 2.7 x86
- python 3.6 x86

But I think that's probably not the right order. Probably the right order (in terms of failing fastest) is this:

- python 2.7 x64
- python 3.3 x64
-python 3.4 x64
- python 3.5 x64
- python 3.6 x64
- python 2.7 x86
- python 3.6 x86

So.. basically exactly what you did, except putting the `x86`  builds at the end instead of the beginning.

Let me know if you feel strongly about this, this is just kinda shooting from the hip.

kirit93:
I'm not sure if there's a way to get statistics, I'll do some research on it. For now I'll go with the order you mentioned second (x86 at the end).

pganssle:
LGTM if @gaborbernat agrees.

appveyor.yml
```diff
@@ -1,14 +1,20 @@
...
   matrix:
-    - PYTHON_VERSION: 27
-    - PYTHON_VERSION: 33
-    - PYTHON_VERSION: 34
-    - PYTHON_VERSION: 35
-    - PYTHON_VERSION: 36
-platform:
-  - x86
-  - x64
+    - PLATFORM : x64
```
gaborbernat: technically correct, but a shorter way of expressing this (and smaller change) is just banning the three excluded envs (https://www.appveyor.com/docs/build-configuration/#exclude-configuration-from-the-matrix)

pganssle: I believe this was @kirit93's original approach. I'm not sure why he abandoned it.

kirit93: I thought it would be simpler to play around with the order of the jobs if I explicitly defined the matrix. But I can revert to the older approach that uses ```exclude``` since the order is just x64 followed by the two x86 jobs.

pganssle:
Hm.. Appveyor does not seem to be clearing the queue, I'm not sure why.

Issue #203, opened 2016-03-04
Version 2.5.0 does not parse string that 2.4.2 does


nixscripter:
I really like dateutils, because it handles some really ugly corner cases for me, like this one:

```
>>> import dateutil
>>> print dateutil.__version__
2.4.2
>>> from dateutil import parser as date_parser
>>> example_date = '15/08/14,10:52:07-20q'  # what my application gives me -- totally non-standard
>>> date_parser.parse(example_date[:17])
datetime.datetime(2014, 8, 15, 10, 52, 7)
```

Or at least, it did until version 2.5.0:

```
>>> import dateutil
>>> print dateutil.__version__
2.5.0
>>> from dateutil import parser as date_parser
>>> example_date = '15/08/14,10:52:07-20q'
>>> date_parser.parse(example_date[:17])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/dateutil/parser.py", line 1161, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/usr/lib/python2.7/site-packages/dateutil/parser.py", line 552, in parse
    raise ValueError("Unknown string format")
ValueError: Unknown string format
```
I hope this is easy to fix.

pganssle:
@nixscripter Just curious, but is 15/08/14 supposed to represent August 14, 2015, or August 15, 2014?

The problem in that string is the `14,10`. I suspect it's related to PR #106.

nixscripter:
@pganssle August 15, 2014 (otherwise, I would use ```yearfirst=True```). The 2.4.2 behavior is what I expect.

And while I didn't do a complete analysis, it does look like the culprit. I imagine it would be pretty difficult to make the state machine recover if it wasn't a number of milliseconds, so I don't know what to suggest.

PR #581, opened 2017-12-10
Make tz.tzstr fail if an invalid GNU tz string is provided

### LINKED ISSUE #259: tzstr not properly detecting invalid strings
I noticed that [this test](https://github.com/dateutil/dateutil/blob/master/dateutil/test/test_tz.py#L903) is not actually parsing the time zone string correctly because that string format is not supported:

``` python
>>> from dateutil import tz
>>> tz.tzstr("EST5EDT") == tz.tzstr("EST5EDT,4,1,0,7200,10,-1,0,7200,3600")
True
>>> tz.tzstr("EST5EDT") == tz.tzstr("EST5EDT,5,4,0,7200,11,-3,0,7200,3600")
True
```

I'm thinking maybe once this is fixed, these tests should be switched to compare tzstr to tzrange, or to some performance-based tests (like checking that the transitions happen at the right time, etc.

pablogsal:
This pr fixes #259. As discussed with @pganssle in the dateutil sprint in London, the fix consist in modifiying `dateutil.parser._parser._tzparser.parse` to check if all tokens generated from the input string have been consumed (except separators that are assumed to be "," and ":" as specified [here](https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html)). To have some control (and help future debugging) on the tokens that are not used, a mark of used_tokens is used internally to keep track of this. To avoid making the parser raise and therefore modifiying the current behaviour, a new slotted field was added to `dateutil.parser._parser._tzparser._result` that indicates if all tokens have been consumed or not. This is used in `dateutil.tz.tz.tzstr` to raise accordingly in the case that this check fails (so not all tokens have been consumed`.

8 test of the test suite for `tz.tz.tzstr` were failing because they were using invalid strings and now the call to  `tz.tz.tzstr` raises. These test were using "default" arguments, that were intented to use the default values that the parser, tzstr and tzrange constructs. To fix these test, the input string has been changed for a GNUtz-compatible equivalent.

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -57,7 +57,7 @@ class _timelex(object):
     # Fractional seconds are sometimes split by a comma
     _split_decimal = re.compile("([.,])")
 
-    def __init__(self, instream):
+    def __init__(self, instream, join_numeric_separated_by_comma=True):
         if six.PY2:
             # In Python 2, we can't duck type properly because unicode has
             # a 'decode' function, and we'd be double-decoding
@@ -77,6 +77,7 @@ class _timelex(object):
         self.charstack = []
         self.tokenstack = []
         self.eof = False
+        self.join_numeric_separated_by_comma = join_numeric_separated_by_comma
 
     def get_token(self):
         """
@@ -144,7 +145,9 @@ class _timelex(object):
                 # numbers until we find something that doesn't fit.
                 if self.isnum(nextchar):
                     token += nextchar
-                elif nextchar == '.':
+                elif nextchar == '.' or \
+                        (self.join_numeric_separated_by_comma
+                         and (nextchar == ',' and len(token) >= 2)):
                     token += nextchar
                     state = '0.'
                 else:
@@ -201,8 +204,8 @@ class _timelex(object):
         return self.__next__()  # Python 2.x support
 
     @classmethod
-    def split(cls, s):
-        return list(cls(s))
+    def split(cls, s, join_numeric_separated_by_comma=True):
+        return list(cls(s,join_numeric_separated_by_comma))
 
     @classmethod
     def isword(cls, nextchar):
@@ -1313,7 +1316,7 @@ class _tzparser(object):
 
     def parse(self, tzstr):
         res = self._result()
-        l = _timelex.split(tzstr)
+        l = _timelex.split(tzstr, join_numeric_separated_by_comma=False)
         used_tokens = [False] * len(l)
         try:
 
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -44,6 +44,8 @@ from six import binary_type, integer_types, text_type
 
 from decimal import Decimal
 
+import re
+
 from .. import relativedelta
 from .. import tz
 
@@ -57,7 +59,7 @@ class _timelex(object):
     # Fractional seconds are sometimes split by a comma
     _split_decimal = re.compile("([.,])")
 
-    def __init__(self, instream, join_numeric_separated_by_comma=True):
+    def __init__(self, instream):
         if six.PY2:
             # In Python 2, we can't duck type properly because unicode has
             # a 'decode' function, and we'd be double-decoding
@@ -77,7 +79,6 @@ class _timelex(object):
         self.charstack = []
         self.tokenstack = []
         self.eof = False
-        self.join_numeric_separated_by_comma = join_numeric_separated_by_comma
 
     def get_token(self):
         """
@@ -145,9 +146,7 @@ class _timelex(object):
                 # numbers until we find something that doesn't fit.
                 if self.isnum(nextchar):
                     token += nextchar
-                elif nextchar == '.' or \
-                        (self.join_numeric_separated_by_comma
-                         and (nextchar == ',' and len(token) >= 2)):
+                elif nextchar == '.' or (nextchar == ',' and len(token) >= 2):
                     token += nextchar
                     state = '0.'
                 else:
@@ -204,8 +203,8 @@ class _timelex(object):
         return self.__next__()  # Python 2.x support
 
     @classmethod
-    def split(cls, s, join_numeric_separated_by_comma=True):
-        return list(cls(s,join_numeric_separated_by_comma))
+    def split(cls, s):
+        return list(cls(s))
 
     @classmethod
     def isword(cls, nextchar):
@@ -625,7 +624,7 @@ class parser(object):
     class _result(_resultbase):
         __slots__ = ["year", "month", "day", "weekday",
                      "hour", "minute", "second", "microsecond",
-                     "tzname", "tzoffset", "ampm","unused_tokens"]
+                     "tzname", "tzoffset", "ampm","any_unused_tokens"]
 
     def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
                fuzzy_with_tokens=False):
@@ -1316,8 +1315,8 @@ class _tzparser(object):
 
     def parse(self, tzstr):
         res = self._result()
-        l = _timelex.split(tzstr, join_numeric_separated_by_comma=False)
-        used_tokens = [False] * len(l)
+        l = [x for x in re.split(r'([,:.]|[a-zA-Z]+|[0-9]+)',tzstr) if x]
+        used_idxs = list()
         try:
 
             len_l = len(l)
@@ -1333,13 +1332,12 @@ class _tzparser(object):
                     if not res.stdabbr:
                         offattr = "stdoffset"
                         res.stdabbr = "".join(l[i:j])
-                        for ii in range(j):
-                            used_tokens[ii] = True
                     else:
                         offattr = "dstoffset"
                         res.dstabbr = "".join(l[i:j])
-                        for ii in range(j):
-                            used_tokens[ii] = True
+
+                    for ii in range(j):
+                        used_idxs.append(ii)
                     i = j
                     if (i < len_l and (l[i] in ('+', '-') or l[i][0] in
                                        "0123456789")):
@@ -1347,7 +1345,7 @@ class _tzparser(object):
                             # Yes, that's right.  See the TZ variable
                             # documentation.
                             signal = (1, -1)[l[i] == '+']
-                            used_tokens[i] = True
+                            used_idxs.append(i)
                             i += 1
                         else:
                             signal = -1
@@ -1361,7 +1359,7 @@ class _tzparser(object):
                             setattr(res, offattr,
                                     (int(l[i]) * 3600 +
                                      int(l[i + 2]) * 60) * signal)
-                            used_tokens[i] = True
+                            used_idxs.append(i)
                             i += 2
                         elif len_li <= 2:
                             # -[0]3
@@ -1369,7 +1367,7 @@ class _tzparser(object):
                                     int(l[i][:2]) * 3600 * signal)
                         else:
                             return None
-                        used_tokens[i] = True
+                        used_idxs.append(i)
                         i += 1
                     if res.dstabbr:
                         break
@@ -1394,34 +1392,34 @@ class _tzparser(object):
                 # GMT0BST,3,0,30,3600,10,0,26,7200[,3600]
                 for x in (res.start, res.end):
                     x.month = int(l[i])
-                    used_tokens[i] = True
+                    used_idxs.append(i)
                     i += 2
                     if l[i] == '-':
                         value = int(l[i + 1]) * -1
-                        used_tokens[i] = True
+                        used_idxs.append(i)
                         i += 1
                     else:
                         value = int(l[i])
-                    used_tokens[i] = True
+                    used_idxs.append(i)
                     i += 2
                     if value:
                         x.week = value
                         x.weekday = (int(l[i]) - 1) % 7
                     else:
                         x.day = int(l[i])
-                    used_tokens[i] = True
+                    used_idxs.append(i)
                     i += 2
                     x.time = int(l[i])
-                    used_tokens[i] = True
+                    used_idxs.append(i)
                     i += 2
                 if i < len_l:
                     if l[i] in ('-', '+'):
                         signal = (-1, 1)[l[i] == "+"]
-                        used_tokens[i] = True
+                        used_idxs.append(i)
                         i += 1
                     else:
                         signal = 1
-                    used_tokens[i] = True
+                    used_idxs.append(i)
                     res.dstoffset = (res.stdoffset + int(l[i])) * signal
             elif (l.count(',') == 2 and l[i:].count('/') <= 2 and
                   not [y for x in l[i:] if x not in (',', '/', 'J', 'M',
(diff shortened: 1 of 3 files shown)


pganssle:
@pablogsal Thanks Pablo!

@jbrockmendel Since you're the resident parser expert, do you want to take a look? I gave it a cursory glance and discussed this in person when it was in development.

pablogsal:
Regarding the tests: They worked because the `tzrange.__eq__` only checks for:

- _std_abbr
- _dst_abbr 
- _std_offset
-  _dst_offset
- _start_delta
- _end_delta

The majority of those are **default** initialized in the parser, so the parser was ignoring the majority of the string and default initializing those. The "worked" but the string used to create the objects was not used in its majority. Therefore this format is not being correctly parsed. As a corollary, no mater what do you write in this format, the result will be the same. Example:

```python

>>> import dateutil.tz
>>> s = "EST5EDT,4,1,0,7200,10,-1,0,7200,3600"
>>> s2 = "EST5EDT,4,4,3,7200,11,-1,0,7200,3600"
>>> tz1 = dateutil.tz.tzstr(s)
>>> tz2 = dateutil.tz.tzstr(s2)
>>> tz1.__dict__
{'_s': 'EST5EDT,4,1,0,7200,10,-1,0,7200,3600', '_std_abbr': 'EST', '_dst_abbr': 'EDT', '_std_offset': datetime.timedelta(-1, 68400), '_dst_offset': datetime.timedelta(-1, 72000), '_start_delta': relativedelta(hours=+2, month=4, day=1, weekday=SU(+1)), '_end_delta': relativedelta(hours=+1, month=10, day=31, weekday=SU(-1)), '_dst_base_offset_': datetime.timedelta(0, 3600), 'hasdst': True}
>>> tz2.__dict__
{'_s': 'EST5EDT,4,4,3,7200,11,-1,0,7200,3600', '_std_abbr': 'EST', '_dst_abbr': 'EDT', '_std_offset': datetime.timedelta(-1, 68400), '_dst_offset': datetime.timedelta(-1, 72000), '_start_delta': relativedelta(hours=+2, month=4, day=1, weekday=SU(+1)), '_end_delta': relativedelta(hours=+1, month=10, day=31, weekday=SU(-1)), '_dst_base_offset_': datetime.timedelta(0, 3600), 'hasdst': True}
```

Notice that both generate `relativedelta(hours=+2, month=4, day=1, weekday=SU(+1))` and `relativedelta(hours=+1, month=10, day=31, weekday=SU(-1))` as `start_delta` and `end_delta`. So if this format was supported at some point, it seems incorrectly parsed,

pganssle:
Interesting. This touches on a discussion @jbrockmendel and I were having in #541.

I'm going to have to go through the history to see when exactly those tests broke, but I don't think that fixing this issue should block the acceptance of this PR, which solves a different goal anyway.

For now can you add back in the old tests and just use @pytest.mark.xfail on them?

I think just restoring the old tests and marking them xfail + adding tests that cover the new code you added should be good, though I'm interested to hear Brock's comments.

pablogsal:
@pganssle In 04d9c2e I have fixed the tokenizer and the parser and I have added test that test though multiple new (and previously uncovered) branches. The old tests are passing with the previous values (so they remain unchanged compared with the base branch). :smile:

pablogsal:
I am currently fixing some minor issues and restoring completely the old tests.

pablogsal:
The only problem now is that ## is failing. This is due to a change in the tokenizer. This change was introduced in bc69c3f3151a3bc9a743d80d29d48701468984ee and it seems is what broke the parser for the default strings because is making the tokenizer tokenize as `['GMT', '0', 'BST', ',', '3', ',', '0', ',', '30.3600', ',', '10.0', ',', '26.7200']` (the bad format). I will adapt the tokenizer to admit both possibilities.

pablogsal:
1b678d4 Allows both behaviours and now everything is passing and the tests remain unchanged.

jbrockmendel:
I'll take a close look at this but won't be able to get to it until tomorrow afternoon.

dateutil/parser/_parser.py
```diff
@@ -57,7 +57,7 @@ class _timelex(object):
     # Fractional seconds are sometimes split by a comma
     _split_decimal = re.compile("([.,])")
 
-    def __init__(self, instream):
+    def __init__(self, instream, join_numeric_separated_by_comma=True):
```
pganssle: Hm.. I don't love this. `_timelex` is semi-public, and this is not something I would ever put in a public API.

pablogsal: I agree. But we need either to select one behavior or implement some flag in this fashion or completely separate de dependencies. What is your preference?

pablogsal: Although I don't like this design I put this so (1) you can identify easily what the problem is and (2) we can discuss better a possible implementation of the fix.

pablogsal: Although I don't like this design I put this so (1) you can identify easily what the problem is and (2) we can discuss better a possible implementation of the fix.

pganssle: Hmmm... I think I see what's going on here.

I find it a bit strange that `_tzparser` re-uses the lexer like this. I see three other possible ways of going here:

1. Split out `_timelexer` and `_tzlexer` into two different lexers with independent rules.
2. Do some sort of token splitting downstream based on the context.
3. Reverse  the commit that keeps the comma token together and dynamically re-join it as necessary.

None of them are particularly satisfying.

Another way to go with your approach is to come up with some more satisfying name for that flag - if there's some unifying theme for the cases where you'd want it to be true, then I'm more OK with adding the flag.

The thing is, 2 and 3 seem wrong because they violate separation of concerns, but we may need to do either 2 or 3 if we want to also solve #203 and other comma-related problems caused by this.

pablogsal: If we want to move meanwhile we implement 1,2 or 3, we can use a new private API to `_timelex` that activates the flag so this is invisible to the end user. Is not good, but is lightweight and it will keep this PR logically separated from bigger fixes. I think 3 underscores will be enough for privacy (`_parser._timelex._privatestuff`). :wink:

jbrockmendel: Having gone through the comments but not yet the code: is there a viable solution here that does not require addressing the `_timelex` design?  If so, I expect that is the best short-term solution.

jbrockmendel: This touches on #541, and as @pganssle says ideally that design question should be addressed separately.

IIRC the comma-as-decimal thing is mostly there for the python logging format.  If that is correct, we can more strictly check for something like `'(?P<h>\d{1,2}):(?P<m>\d{2})(:(?P<s>\d{2}([,\.]\d+)?))` (though likely not using regex)

dateutil/parser/_parser.py
```diff
@@ -201,8 +204,8 @@ def next(self):
         return self.__next__()  # Python 2.x support
 
     @classmethod
-    def split(cls, s):
-        return list(cls(s))
+    def split(cls, s, join_numeric_separated_by_comma=True):
+        return list(cls(s,join_numeric_separated_by_comma))
```
jbrockmendel: flake8 whitespace after comma

dateutil/parser/_parser.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, instream):
         self.charstack = []
         self.tokenstack = []
         self.eof = False
+        self.join_numeric_separated_by_comma = join_numeric_separated_by_comma
```
jbrockmendel: can you add a comment for why this option exists, namely for the python logging format

dateutil/parser/_parser.py
pganssle: @jbrockmendel There are a large number of possible hacky workarounds we could do, but it's basically the same problem as with [this PR](https://github.com/dateutil/dateutil/pull/279#discussion_r71794602) - right now we have some tokens that are combined when they shouldn't be, so the options are:

1. Add an additional tokenizer step in the parser that splits these aggressively-combined tokens (no direct effect on `_timelex`, but does couple to the current behavior)
2. Make it so `_timelex` doesn't combine these anymore and add a re-combination step in the parser portion.
3. Some variation on Pablo's solution, e.g. "tell `_timelex` which behavior we want when we call it" - this can either be a flag like what Pablo is doing or we subclass `_timelex` to make a `_tzlex`, and the two vary depending on this behavior.

I don't see any other obvious solution, since in some contexts you want the tokens combined and in some contexts you want the tokens split.

dateutil/parser/_parser.py
pganssle: Is it just the python logging format? I think that it's *also* for the `ISO-8601` format, which *technically* (though not in practice) prefers comma as the fractional separator.

jbrockmendel:
Is there a spec that can be linked to describing how tzparser is intended to work?

dateutil/parser/_parser.py
```diff
@@ -1329,16 +1333,21 @@ def parse(self, tzstr):
                     if not res.stdabbr:
                         offattr = "stdoffset"
                         res.stdabbr = "".join(l[i:j])
+                        for ii in range(j):
+                            used_tokens[ii] = True
                     else:
                         offattr = "dstoffset"
                         res.dstabbr = "".join(l[i:j])
+                        for ii in range(j):
+                            used_tokens[ii] = True
```
jbrockmendel: This loop is done in both branches of this if/else.  Can you de-indent and just do it once?

pganssle:
@jbrockmendel That's part of the problem. `_tzparser` itself is actually private, but it's used in a few places internally. The closest thing to a "spec" is the examples in the `examples` documentation.

If we were to take `tzstr`'s documentation at its word that it only supports GNU TZ strings, then we could use that as a spec, but the documentation and examples seem to indicate that it actually supports other formats as well.

dateutil/parser/_parser.py
```diff
@@ -1464,6 +1494,7 @@ def parse(self, tzstr):
         except (IndexError, ValueError, AssertionError):
             return None
 
+        res.unused_tokens = not {token for token,is_used in zip(l,used_tokens) if not is_used}.issubset({",",":"})
```
jbrockmendel: This took some staring out to figure out.  If nothing else, please break this into a couple of lines starting with `unused_tokens = {token for token,is_used in zip(l,used_tokens) if not is_used}`.

Related, the name `res.unused_tokens` suggest that it is listlike giving the tokens themselves.  If this is a bool, maybe something like `res.any_unused_tokens`?

More broadly, I'd suggest the following:

Implement `used_tokens` as `used_idxs` mirroring `skipped_idxs` in `_parser._parse`.   Then each occurrence of `used_tokens[i] = True` becomes `used_idxs.append(i)`.  Then `{token for token,is_used in zip(l,used_tokens) if not is_used}` becomes:

```
unused_idxs = set(range(len_l)).difference(used_idxs)
unused_tokens = {l[n] for n in unused_idxs}
```

pganssle:
> Is using timelex.split even necessary here?

This is kinda what I was thinking. It's a little weird that `_timelex` is being reused for this purpose.

dateutil/parser/_parser.py
jbrockmendel: Oh neat, I had no idea.  I guess I've seen that used in conjunction with Euros, e.g. "€ 29.123,45"

Does the ISO usage need to be for any numbers or is it just for fractional seconds?  i.e. will the check-for-colons above work?

pganssle: Technically it's for all fractional time components, but I think basically no one uses anything except fractional seconds. Still, this is technically a valid ISO-8601 datetime:

`2014-01-01T00,50` -> `datetime(2014, 1, 1, 0, 30)`

pablogsal:
@pganssle @jbrockmendel In 4f9e235 I have implemented a simple regexp lexer that makes the trick (and is 35x +- 1.2 times faster) so this leaves `_timelex` and its problems unchanged and therefore is a problem for another day/PR. Notice that this lexer is not as straightforward as splitting on commas+colons+slashes+whitespace because you have to split into numeric/non numeric as well. I have addressed all the style feedback that you indicated. :smile:

dateutil/parser/_parser.py
```diff
@@ -1352,19 +1359,22 @@ def parse(self, tzstr):
...
                         i += 1
                     if res.dstabbr:
                         break
                 else:
                     break
 
+
             if i < len_l:
                 for j in range(i, len_l):
                     if l[j] == ';':
```
pablogsal: Can I eliminate this as a possibility for an alternate separator? Is not being hit by the test and clearly does not work if you consistently use ";" as a separator.

dateutil/parser/_parser.py
```diff
@@ -1313,7 +1315,8 @@ def __init__(self):
 
     def parse(self, tzstr):
         res = self._result()
-        l = _timelex.split(tzstr)
+        l = [x for x in re.split(r'([,:.]|[a-zA-Z]+|[0-9]+)',tzstr) if x]
```
jbrockmendel: If we see something like `"foo,:"bar` with two of `[,:.]` in a row is that indicative of an invalid string?

pablogsal: The parser will fail in that case. I will push for the tokenizer part not doing anything else that tokenize and just let the parser naturaly fail to parse a consecutive `[,:.]`.

jbrockmendel: The reason why I ask is to see if the `if x` clause is necessary.

pablogsal: This is necessary for Python<3.6. This is due to the way `re.split` works. I have made a failing commit to prove this in b15f38a. [Here](https://travis-ci.org/dateutil/dateutil/builds/316445999?utm_source=github_status&utm_medium=notification) are the CI results of that commit:

pganssle: I'm not sure I understand this regular expression completely, but I think the only problem with this approach that I see is that any `split` based function is going to discard the tokens themselves. Are there really no constraints on what the inputs can be for valid TZ strings?

Also, with the `if x` filter, how would you distinguish between `One,Two` and `One,,Two`?

```python
In [6]: [x for x in re.split(r'[.,;]', 'One,Two') if x]
Out[6]: ['One', 'Two']

In [7]: [x for x in re.split(r'[.,;]', 'One,;Two') if x]
Out[7]: ['One', 'Two']
```

dateutil/parser/_parser.py
pganssle: I really wish I knew, honestly. :(

dateutil/parser/_parser.py
pablogsal: You forgot the capturing group:

```python
>>> [x for x in re.split(r'([,:.]|[a-zA-Z]+|[0-9]+)','One,Two') if x]
['One', ',', 'Two']
>>> [x for x in re.split(r'([,:.]|[a-zA-Z]+|[0-9]+)','One,;Two') if x]
['One', ',', ';', 'Two']
```

pganssle:
@jbrockmendel Any other comments on this? I'm pretty happy with this solution so far.

dateutil/parser/_parser.py
```diff
@@ -1464,6 +1492,8 @@ def parse(self, tzstr):
         except (IndexError, ValueError, AssertionError):
             return None
 
+        unused_idxs = set(range(len_l)).difference(used_idxs)
+        res.any_unused_tokens = not {l[n] for n in unused_idxs}.issubset({",",":"})
```
pganssle: It's kinda amazing to me that this seems to be *universally* faster than `any(l[n] not in {',', ':'} for n in unused_idxs)`, even when you pull the `{',', ':'}` creation out of the loop.

Maybe if you have some absurd number of tokens the set creation portion would get faster and this would get slow.

dateutil/parser/_parser.py
pganssle: Oh wow, that's a neat trick. I didn't realized it worked like that. Good to know.

jbrockmendel:
LGTM

pganssle:
OK, I'll clean this up a bit to my exacting standards that feel too nitpicky to mention in an actual code review and then merge.

pganssle:
@pablogsal I have made some changes, mostly to actually test the *behavior* of `tzstr` rather than simply check that it doesn't throw an error. Please let me know if you think I've done it wrong. Once you approve, I think it's ready for merge.

pganssle:
I'm just going to go ahead and merge. If there are issues they can be addressed in a separate PR.

PR #589, opened 2017-12-11
remove min() calls

jbrockmendel:
At least we can get something out of the adventure with asv.  We can be fairly confident this is real because we saw cProfile results.

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -324,19 +324,17 @@ class parserinfo(object):
         return name.lower() in self._jump
 
     def weekday(self, name):
-        if len(name) >= min(len(n) for n in self._weekdays.keys()):
-            try:
-                return self._weekdays[name.lower()]
-            except KeyError:
-                pass
+        try:
+            return self._weekdays[name.lower()]
+        except KeyError:
+            pass
         return None
 
     def month(self, name):
-        if len(name) >= min(len(n) for n in self._months.keys()):
-            try:
-                return self._months[name.lower()] + 1
-            except KeyError:
-                pass
+        try:
+            return self._months[name.lower()] + 1
+        except KeyError:
+            pass
         return None
 
     def hms(self, name):


dateutil/parser/_parser.py
```diff
@@ -324,19 +324,17 @@ def jump(self, name):
         return name.lower() in self._jump
 
     def weekday(self, name):
-        if len(name) >= min(len(n) for n in self._weekdays.keys()):
```
pganssle: This was recently added [here](https://github.com/dateutil/dateutil/commit/505b9b22605d05fa5c6d41a66ea95f88022018fd#diff-ed798069162e7ab7db9a45da6c4a076fL312), replacing a `len(name) >= 3` check that was [part of the original commit in this repo](https://github.com/dateutil/dateutil/commit/68ae2757ae15c84bf947d47a82a314b3b975bc9b).

I think it was just there for performance reasons.

pganssle:
This is an unambigous win over the current code.

To see what the best thing to do is, I made two versions, `parserinfo` does the `min` call in the constructor and stores it as a private attribute, and `parserinfo2` is the code as presented here.  and `parserinfo3` is the code as is now:

```python
>>> i = parserinfo()
>>> i2 = parserinfo2()
>>> i3 = parserinfo3()

>>> %timeit i.weekday('Wed')
1000000 loops, best of 3: 627 ns per loop

>>> %timeit i2.weekday('Wed')
1000000 loops, best of 3: 393 ns per loop

>>> %timeit i3.weekday('Wed')
100000 loops, best of 3: 3.4 µs per loop

>>> %timeit i.weekday('A')
1000000 loops, best of 3: 285 ns per loop

>>> %timeit i2.weekday('A')
1000000 loops, best of 3: 724 ns per loop

>>> %timeit i3.weekday('A')
100000 loops, best of 3: 2.83 µs per loop
```

When the string passed to `weekday` has length >= the shortest string in in the `weekday` keys, this version is best. When it's `<=` the min, the cached version is best. The version we have now performs worst in all cases.

I'm not sure all the code paths for the common case, but I'm guessing that it's actually kinda rare to hit the case where `weekday` or `month` are passed 2-character strings, so this approach seems the best one to me.

jbrockmendel:
Yah, once we get to the point of trying to out-perform a dict lookup, it's time to go read a book.

pganssle:
I suppose we could have done `self._weekdays.get(name.lower(), None)` for `weekdays`.

Issue #396, opened 2017-06-09
Improve and/or automated release procedure


pganssle:
Right now, releases follow a very manual process, following [this checklist](https://github.com/dateutil/dateutil/blob/master/RELEASING). Some parts of that need to be manual (such as updating the NEWS file), but it would be nice to have some tooling that's not quite so prone to mistakes. Low-hanging fruit for things to be scripted:

- [X] Bump version numbers (maybe take user prompt) and checks for the latest version of Python. (Done in #564)
- [ ] Pull the latest zonefile_metadata.json from https://dateutil.github.io/tzdata/zonefile_metadata/zonefile_metadata.json (there's already a mostly-automated process to update the zonefile archive).
- [ ] Build the source and binary distributions (more or less `release.py` does this)
- [ ] Install them in a virtual environment and runs them through tox (possibly this can be done with some CI that would deliver the builds to me for signature).
- [X] Once we have the source and binary distributions, script to automatically generate the hashes and signatures, then uploads them all to the testing version of pypi (Done with `release.py`)
- [X] Same as above, but upload to the real pypi.

Some possible ideas for improvements in general can be found in [Dave Forgac's Pycon 2017 talk](https://www.youtube.com/watch?v=qOH-h-EKKac) (@tylerdave).

CC: @jbrockmendel @jdufresne You guys have been doing some behind-the-scenes improvements lately (for which I am very grateful) - any thoughts on this?

jbrockmendel:
Managing releases like a grown-up is a difficult looking problem that I've (happily) never had to deal with.  Interested in learning about it and being helpful if possible, but don't have any useful input.

wagner-certat:
To sign and upload releases I am using [twine](https://pypi.python.org/pypi/twine/). See also e.g. https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/

pganssle:
Ping @lkollar

lkollar:
I've opened #591 to implement the release process with make. At this point it is more of a proof of concept as I'm not exactly sure what the exact commands should be, but I think most if not all of the checklist can be done this way. Let me know what you think.

mariocj89:
This is already automated, the last step missing: "Pull the latest zonefile_metadata.json" will not be relevant once we move to use python/tzdata.

PR #591, opened 2017-12-13
[WiP] Add Makefile

lkollar:
This implements parts of the release process outlined in #396. I'm not sure of the exact commands used in the release checklist so I've improvised in a few places, but I intended this as a proof of concept and definitely needs mork work.

Let me know what steps are wrong/missing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,47 @@
+
+PYTHON := python
+SETUP_EXTRA_ARGS =
+
+.PHONY: pytest
+pytest:
+	$(PYTHON) -m pytest
+
+.PHONY: test
+test: pytest
+
+.PHONY: tox
+tox:
+	tox
+
+.PHONY: fetch_tzmeta
+fetch_tzmeta:
+	curl -sO https://dateutil.github.io/tzdata/zonefile_metadata/zonefile_metadata.json
+
+.PHONY: sdist
+sdist:
+	$(PYTHON) setup.py sdist
+
+.PHONY: bdist_wheel
+bdist_wheel:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,47 @@
...
+.PHONY: test-upload
+test-upload: SETUP_EXTRA_ARGS = -r https://test.pypi.org/legacy/
+test-upload: upload
+
+.PHONY: release
+release: test fetch_tzmeta sdist bdist_wheel
+
+.PHONY: clean
+clean:
+	-rm -rf *.egg-info
```
pganssle: `__pycache__`?

pganssle:
This doesn't seem like a significant improvement over my `release.py` script, IMO.

lkollar:
I had no idea about `release.py`. The issue linked mentions a checklist only.

pganssle:
I probably wrote `release.py` after I wrote that "automated release procedure" issue, or around the same time.

It's basically the same as this script, except written in Python, which I find a bit easier to maintain. The downsides are:

1. It's non-standard (unlike `make`), so discoverability is probably reasonably low.
2. It does not run the tests against the built libraries.

I *could* also do a sort of automated release where the CI builds the package, runs the tests, and delivers them to PyPI. That's not out of the question, but I don't *love* the trust model there.

lkollar:
I had no idea that parts of the release process was already automated. It doesn't make much sense to have a Makefile and the release script as well so I can just close this PR.

#396 should probably be closed or updated with the missing release steps?

PR #601, opened 2017-12-27
Fixes handling of the GNU C Library default timezone value

lapointexavier:
## Description

I just found an edge case (I believe) that was not handled properly. It seems like the slice on the `TZ` environment variable value was incorrect, using `name[:-1]` instead of `name[1:]`.

I would end up with the following:
```
(Pdb++) name
 ':/etc/localtim'
```

... instead of:

`/etc/localtime`

## Background

The default timezone value can be `TZ=:/etc/localtime`. It may vary, but usually starts with `:`.

Documentation: https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

> If the TZ environment variable does not have a value, the operation chooses a time zone by default. In the GNU C Library, the default time zone is like the specification ‘TZ=:/etc/localtime’ (or ‘TZ=:/usr/local/etc/localtime’, depending on how the GNU C Library was configured; see Installation). Other C libraries use their own rule for choosing the default time zone, so there is little we can say about them.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -69,6 +69,7 @@ switch, and thus all their contributions are dual-licensed.
 - Tomasz Kluczkowski (gh: @Tomasz-Kluczkowski) **D**
 - Tomi Pieviläinen <tomi.pievilainen@iki.fi>
 - Unrud <Unrud@MASKED> (gh: @unrud)
+- Xavier Lapointe <lapointe.xavier@MASKED> (gh: @lapointexavier) **D**
 - X O <xo@MASKED>
 - Yaron de Leeuw <me@jarondl.net> (gh: @jarondl)
 - Yoney <alper_yoney@hotmail.com> **D**
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -910,6 +910,12 @@ def test_tzlocal_utc_unequal(tzvar):
 
 
 @mark_tzlocal_nix
+def test_tzlocal_local_time_trim_colon():
+    with TZEnvContext(':/etc/localtime'):
+        assert tz.gettz() is not None
+
+
+@mark_tzlocal_nix
 @pytest.mark.parametrize('tzvar, tzoff', [
     ('EST5', tz.tzoffset('EST', -18000)),
     ('GMT', tz.tzoffset('GMT', 0)),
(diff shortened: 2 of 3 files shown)


pganssle:
Hmmm... This is an untested branch and *seems* obviously right, but I'm not sure how this has never come up before.

@lapointexavier Did this come up for you "in the wild", or did you figure it out from the docs?

CI failure is fixed in #602, when that is merged you can rebase against master to get a passing CI run.

lapointexavier:
It actually came out since we switched our dev environment to kubernetes. I imagine something's different in the environment as we were using vagrant before.

This block of code started failing and I had to drop in a debugger to see what was going on. 

(code edited for simplicity)
```
LOCAL_TZ = dateutil.tz.gettz()

dt = dateutil.parser.parse('<some datetime>')
dt = dt.astimezone(LOCAL_TZ).replace(tzinfo=None)
```

pganssle:
Interesting, seems like just no one was hitting that branch before, which is interesting in itself.

For your current situation, here's a workaround:

```python
LOCAL_TZ = dateutil.tz.tzlocal()

dt = dateutil.parser.parse('<some datetime>')
dt = dt.astimezone(LOCAL_TZ).replace(tzinfo=None)
```

Since you throw away the time zone immediately anyway, it doesn't really matter if you get a `tzfile` or `tz.tzlocal`.

pganssle:
@lapointexavier I merged #602, so if you want to rebase against master that will fix CI, then I can merge this one.

lapointexavier:
ah great, thanks!

lapointexavier:
@pganssle Should be good to go 🚀

pganssle:
@lapointexavier Oh, one other thing (probably should add this to the PR template when I get a chance), can you add yourself to the [`AUTHORS.md`](https://github.com/dateutil/dateutil/blob/master/AUTHORS.md) file? Goes in alphabetical order by first name. Put a **D** next to your name to indicate that your contributions have always been dual-licensed Apache/BSD. You can use any name you want, all fields are optional.

jbrockmendel:
Nice catch.

pganssle:
@lapointexavier By the way, I masked those e-mail address domains because I hadn't asked permission from those authors to actually publish their e-mail address. If you want your e-mail address included, I can include it next time I make a change to `AUTHORS.MD`. I assume you realize this and didn't want your e-mail included, but I thought I'd make that clear.

lapointexavier:
@pganssle yeah I was a little ambivalent. I guess it wouldn't hurt anyway. Here it is: `lapointe.xavier@protonmail.com`

PR #604, opened 2018-01-02
Fix/rrule str tzinfo

### LINKED ISSUE #603: rrule __str__ strips timezone information
If an `rrule` is created with timezone information in `DTSTART`, it appears to function correctly, but converting it back to a `str` removes the timezone information from `DTSTART`.

```python
>>> from dateutil.rrule import rrulestr
>>> str(rrulestr('DTSTART:20180102T120000Z\nFREQ=DAILY'))
'DTSTART:20180102T120000\nFREQ=DAILY'
```

Expected output:
`DTSTART:20180102T120000Z\nFREQ=DAILY`

OS: Alpine Linux 3.7
dateutil version 2.6.1
Python 3.6.3

I am using RRULE strings to store date/time information in a database. For now, I'm working around this by managing `DTSTART` separately from the RRULE strings and stripping it when serializing/deserializing the RRULE.

jamesmallen:
Fixes #603

### DIFF
--- a/dateutil/test/test_rrule.py
+++ b/dateutil/test/test_rrule.py
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 from ._common import WarningTestMixin
 
-from datetime import datetime, date
+from datetime import datetime, date, tzinfo, timedelta
 import unittest
 from six import PY3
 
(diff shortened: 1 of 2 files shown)


dateutil/test/test_rrule.py
```diff
@@ -4439,6 +4439,26 @@ def testToStrLongIntegers(self):
                                   byweekno=long(2),
                                   dtstart=datetime(1997, 9, 2, 9, 0)))
 
+    def testToStrTimezone(self):
+        ZERO = timedelta(0)
+
+        class UTC(tzinfo):
```
pganssle: There's already a UTC class, `dateutil.tz.tzutc`, which also hhas a singleton, `dateutil.tz.UTC` available for use.

pganssle:
This only solves the problem for the trivial case of fixed offsets, and in fact will give the *wrong* answer in the more common case of non-fixed offsets, since it makes no effort to round-trip the actual `tzinfo` object used (or equivalent).

Consider this:

```python
from datetime import datetime
from dateutil import rrule, tz

ET = tz.gettz('America/New_York')
rr = rrule.rrule(rrule.MONTHLY, dtstart=datetime(2017, 1, 15, 12, tzinfo=ET), count=12)
rr_rt = rrule.rrulestr(str(rr))

for dt1, dt2 in zip(rr, rr_rt):
    assert dt1 == dt2, (str(dt1), str(dt2))
```

With your patch, the result is:

> AssertionError: ('2017-03-15 12:00:00-04:00', '2017-03-15 12:00:00-05:00')

As a general rule, I think that using datetime-aware timezones in `rrule` is not a great idea.

PR #606, opened 2018-01-03
Deprecate tz format

### LINKED ISSUE #595: Deprecate idiosyncratic tzstr format
Per [this question](https://stackoverflow.com/q/47874633/467366) on SO, unless we can find some sort of standard that matches up with `# GMT0BST,3,0,30,3600,10,0,26,7200[,3600]`, I think we can go ahead and deprecate it as a supported `tzstr` format (just in time now that @pablogsal has fixed it our support for it)!

I'm thinking we should

- [x] Add a `DeprecatedFormatWarning` on a successfully parsed string in this format (not a `DeprecationWarning`, I want people to actually see it)
- [x] Change all tests using this format so that they now check to make sure the warning is raised.
- [x] Remove this format from all tests where we're not *explicitly* testing our support for the format.
- [x] Remove any code examples or documentation related to this format, except possibly a `.. caution` note about the deprecation of the format.

I'm fairly confident that this won't cause major problems, since support for this format was inadvertently broken several releases back and we haven't heard anything.

pganssle:
Fixes #595.

This is a release blocker, so we should get this out ASAP.

CC: @pablogsal

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -44,7 +44,7 @@ from six import binary_type, integer_types, text_type
 
 from decimal import Decimal
 
-import re
+from warnings import warn
 
 from .. import relativedelta
 from .. import tz
@@ -1419,6 +1419,14 @@ class _tzparser(object):
                         signal = 1
                     used_idxs.append(i)
                     res.dstoffset = (res.stdoffset + int(l[i]) * signal)
+
+                # This was a made-up format that is not in normal use
+                warn(('Parsed time zone "%s"' % tzstr) +
+                     'is in a non-standard dateutil-specific format, which ' +
+                     'is now deprecated; support for parsing this format ' +
+                     'will be removed in future versions. It is recommended ' +
+                     'that you switch to a standard format like the GNU ' +
+                     'TZ variable format.', tz.DeprecatedTzFormatWarning)
             elif (l.count(',') == 2 and l[i:].count('/') <= 2 and
                   not [y for x in l[i:] if x not in (',', '/', 'J', 'M',
                                                      '.', '-', ':')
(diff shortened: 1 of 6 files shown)


pablogsal:
I will take a look today at this when I get home.

docs/examples.rst
```diff
@@ -1328,7 +1326,7 @@ Check the daylight limit.
     'EDT'
     >>> datetime(2003, 10, 26, 0, 59, tzinfo=tz).tzname()
     'EDT'
-    >>> datetime(2003, 10, 26, 1, 00, tzinfo=tz).tzname()
+    >>> datetime(2003, 10, 26, 2, 00, tzinfo=tz).tzname()
```
pganssle: I think this indicates we're not testing the doctests. Probably we should figure out how to mock out now in doctests so that we can actually get passing doctests.

dateutil/parser/_parser.py
```diff
@@ -1419,6 +1419,14 @@ def parse(self, tzstr):
...
                     used_idxs.append(i)
                     res.dstoffset = (res.stdoffset + int(l[i]) * signal)
+
+                # This was a made-up format that is not in normal use
+                warn(('Parsed time zone "%s"' % tzstr) +
+                     'is in a non-standard dateutil-specific format, which ' +
+                     'is now deprecated; support for parsing this format ' +
+                     'will be removed in future versions. It is recommended ' +
+                     'that you switch to a standard format like the GNU ' +
+                     'TZ variable format.', tz.DeprecatedTzFormatWarning)
```
pablogsal: Suggestion: Include link to the GNU TZ variable format either withing dateutil documentation or to the format specification.

pganssle: That's a good idea, and one I considered. I think it is in the [tzstr documentation](https://dateutil.readthedocs.io/en/latest/tz.html#dateutil.tz.tzstr). I'm a bit hesitant to add it here, since I'm afraid of it going out of date. I'll merge this for now, we can tweak the message later.

PR #593, opened 2017-12-18
Imaginary date test same as #565

Cheukting:
Using a separate branch, clean up format and files, added more test. (continuation of #565)
Also, added doc for function resolve_imaginary
Added myself to the Author page
@pganssle please kindly check and let me know if anything more changes are required :-)

PR #607, opened 2018-01-03
Add resolve_imaginary

### LINKED ISSUE #339: Add resolve_imaginary function to `tz`

pganssle:
Closes #339 in favor of #565 and #593.

@Cheukting I had to make a new branch to clean up your git history, which wouldn't have merged cleanly into the master branch as it was. If you make further PRs, it would probably be advisable to delete your fork and make a new one, because I think the master branch is out of sync with the master branch here.

### DIFF
--- a/dateutil/test/test_imports.py
+++ b/dateutil/test/test_imports.py
@@ -123,9 +123,13 @@ class ImportTZTest(unittest.TestCase):
         from dateutil.tz import tzwin
         from dateutil.tz import tzwinlocal
         from dateutil.tz import UTC
+        from dateutil.tz import datetime_ambiguous
+        from dateutil.tz import datetime_exists
+        from dateutil.tz import resolve_imaginary
 
         tz_all = ["tzutc", "tzoffset", "tzlocal", "tzfile", "tzrange",
-                  "tzstr", "tzical", "gettz", "UTC"]
+                  "tzstr", "tzical", "gettz", "datetime_ambiguous",
+                  "datetime_exists", "resolve_imaginary", "UTC"]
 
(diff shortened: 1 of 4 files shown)


Cheukting:
Yes it was a nightmare as I worked on the master branch. Thanks for tidy it up eventually and I hope my work helps (not causing too much trouble). 

Definitely started a new fork when you have taken over and after make sure that deleting my fork will not delete the edit done.

pganssle:
Only problem is a dead link, merging.

Issue #335, opened 2016-12-20
Parser should raise specific error subclasses


pganssle:
This is a more general problem in the codebase, but ideally, we would use our own error subclasses for various errors, rather than just emitting the base `ValueError` or `TypeError` or whatever. Using specific error subclasses will allow people to more easily catch some errors and not others.

jbrockmendel:
We now raise InvalidDatetimeError in a handful of places.  Does that cover this issue?

pganssle:
I think maybe not. I'll have to look.

pganssle:
@jbrockmendel So this is not 100% solved since there are still cases where it raises `ValueError`, and worse yet, it's not really clear when `ValueError` vs. `InvalidDatetimeError` vs. `InvalidDateError` vs. `InvalidTimeError` would be raised, which makes me think that it will be hard for people to make meaningful use of the different error types, and kinda muddies the waters.

Can you clarify what your thinking was with these error types?

I think there are two very reasonable ways to handle this:

1. Use a single error class derived from `ValueError` like `ParserError` or `UnknownFormatError` or something and always raise that. If we want to distinguish between error types based on the failure mode, we can use `raise from` so that those error types are visible, but since that is not something I think people would be able to (or want to) catch and try to recover from, it's probably sufficient to just use `ValueError` for the internal errors and wrap them for the final (catchable) error.

2. Use multiple error classes depending on the type of failure we got in a way that *could* be recoverable. I think there are approximately four possible failure modes here:
    1. We have something that doesn't look like a date (e.g. `Scramulax`)
    2. We got something that *looks* like a date, but isn't a valid date (e.g. `2017-02-30`)
    3. We got something that looks like a date with a time zone, but we don't know how to map the time zone into something meaningful.
    4. We got something that contains a date, but also contains other non-date stuff and `fuzzy` is not active.


I like the finer-grained controls of the second option *except* that there's a better than average chance that we'll be raising errors, so it won't be reliable enough to actually catch in an automated way. For example, since the parser really tries to give the input the benefit of the doubt, it will likely identify way too many things as option 2-2 when it's actually 2-1. Similarly we're already having a lot of confusion between 2-2, 2-3 and 2-4. Not to mention 2-4 is a situation that can be easily remedied by just parsing *all* dates with `fuzzy_with_tokens=True` on and then doing whatever "recovery" you need by inspecting the returned tokens.

[...]

PR #610, opened 2018-01-07
Only raise ValueError in parser for now

pganssle:
Preparing for the 2.7.0 release, I don't want to have a half-done implementation of #335. For now we'll go back to all `ValueError` and then implement the new subclasses strategy in one fell swoop.


jbrockmendel:
Totally on board with keeping ValueError rather than release something half-baked.  Will review now.

PR #611, opened 2018-01-11
Remove support for end of life Python 3.2 and 3.3

jdufresne:
Python 3.2 and 3.3 are end of life. They are no longer receiving bug fixes including for security issues. They have been EOL since 2016-02-20 and 2017-09-29 respectively. For additional details, see:

https://devguide.python.org/#status-of-python-branches

Reduces testing and maintenance resources.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: python
 cache: pip
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
@@ -30,4 +29,3 @@ script:
 
 after_success:
   - if [[ $TOXENV == "py" ]]; then tox -e coverage,codecov; fi
-
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,6 @@
 UNRELEASED
 ----------
-- **Dropped support for Python 2.6.**
+- **Dropped support for end of life Python 2.6, 3.2, and 3.3.**
 
 Version 2.6.1
 -------------
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,6 @@ build: false
 environment:
   matrix:
     - PYTHON_VERSION: 27
-    - PYTHON_VERSION: 33
     - PYTHON_VERSION: 34
     - PYTHON_VERSION: 35
     - PYTHON_VERSION: 36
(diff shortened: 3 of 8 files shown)


jbrockmendel:
+1.  I think this will also allow replacing a few appearances of `isinstance(whatever, collections.Callable)` with `callable(whatever)`.

pganssle:
I don't really have any problem dropping old versions of Python if we want to use some features of the later versions, but I would prefer it if the supported versions manifest were descriptive rather than prescriptive. I don't really care about sending the "wrong message" by accurately describing that the software works with Python 3.2 and Python 3.3.

That said, if @jbrockmendel is right that there are some improvements that can be made to the codebase by dropping support for these old versions, I'm fine with merging this after the 2.7.0 release so that we can make those improvements.

jdufresne:
>  I think this will also allow replacing a few appearances of isinstance(whatever, collections.Callable) with callable(whatever).

The [Python documentation](https://docs.python.org/3/library/functions.html#callable) says this is available on all Pythons supported by dateutil now. It only doesn't exist in Python 3.0/31. Therefore, I've kept this change separate in #612.

jdufresne:
@pganssle No worries. Just noticed this and thought I'd offer the change. I don't see many projects supporting 3.2 or 3.3 these days.

Python 3.2 is neither listed in `tox.ini`, `.travis.yml`, nor `appveyor.yml`. Do you have confidence that the code still works on 3.2?

To put some numbers out there, here are the pip installs for dateutil from PyPI for the last month. It looks like dateutil has very few 3.3 downloads and virtually no 3.2 downloads.

```
$ pypinfo --percent --pip python-dateutil pyversion
| python_version | percent | download_count | 
| -------------- | ------- | -------------- | 
| 2.7            |   86.1% |     11,811,360 |
| 3.6            |    6.4% |        883,738 |
| 3.5            |    3.9% |        536,262 |
| 3.4            |    2.5% |        339,697 |
| 2.6            |    1.0% |        142,543 |
| 3.3            |    0.1% |          7,014 |
| 3.7            |    0.0% |          3,915 |
| 3.2            |    0.0% |            146 |
| None           |    0.0% |             81 |
| 3.1            |    0.0% |              1 |
```

I've updated the commit message to remove the "wrong message" wording.

pganssle:
@jdufresne Yeah, I think we actually may have dropped Python 3.2 support during the London sprint and failed to update the trove classifiers and `setup_requires`. I can't remember if we actually dropped *support* or if we just dropped the tests because it was too much of a PITA to get it working with tox.

pganssle:
@jdufresne I think this has a bunch of conflicts with the base now that it's been partially implemented. I'm going to close this for now.

PR #612, opened 2018-01-12
Prefer builtin callable() over isinstance(..., collections.Callable)

jdufresne:
Available on all supported Python versions.

https://docs.python.org/3/library/functions.html#callable


pganssle:
LGTM! Thanks for all your improvements to the codebase.

Issue #592, opened 2017-12-15
Add changelog management


pganssle:
Possibly we should use [towncrier](https://pypi.python.org/pypi/towncrier) for this?

Manually updating all the changelogs and such is kind of a pain at the end of each release cycle. This could make things much easier.

wagner-certat:
> Possibly we should use towncrier for this?

As far as I understood the readme correctly, it creates a news-file (for users) from an existing changelog (for developers), where the changelog needs to have a very special format. Not sure, if this really helps you.

pganssle:
> As far as I understood the readme correctly, it creates a news-file (for users) from an existing changelog (for developers), where the changelog needs to have a very special format. Not sure, if this really helps you.

The way it would help is that each PR would require a news entry (which I could detect in the CI), and when I do a release it would be a matter of executing `towncrier` rather than manually going through all the old PRs and issues and seeing who reported them and who fixed them, etc. Just having the work done "as you go" is helpful.

The value of doing it with `towncrier` or some equivalent software is to avoid merge conflicts on the single `NEWS` file.

wagner-certat:
I saw the same approach in https://github.com/xonsh/xonsh though they use their own script to merge the single files in the directory. towncrier also handles different classes of changes, that's for sure a plus.

PR #613, opened 2018-01-18
[WIP] Add new build options for setup.py

### LINKED ISSUE #597: Add no-zoneinfo build option
Currently the `zoneinfo` package is only really necessary if you want to use *our* zoneinfo files, which I suspect are mostly used either out of ignorance or on Windows (which does not ship zoneinfo).

Ideally, we'd have an option to install a version of `dateutil` *without* `zoneinfo` (no files, and without the module). I'm not really sure how to do this, however. There's the [`extras-require`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) option that lets you have some sort of conditional build dependencies, but I don't know if there's a way to have "extras" that essentially **take away** features, rather than add them.

pganssle:
Fixes #597.

This will add the `--no-zoneinfo`, `--tzfiles` and `--tzpath` command line arguments so that downstream distributors can easily build versions of this where they are better able to manage how `zoneinfo` files are distributed.

To do:
- [ ] Add tests
- [ ] Add documentation 
- [ ] Possibly forbid these options in things like `sdist`?
- [ ] Test to see if this can be made to work with `pip`


pganssle:
@wagner-certat The current release procedure doesn't generate the NEWS file until the release is prepared, though I would like to change that with #592.

That said, "add documentation" is one of the bullet points.

Issue #614, opened 2018-01-18
TZID is not supported in rrule DTSTART


ryanpetrello:
```python
from dateutil import rrule
rrule.rrulestr('DTSTART;TZID=America/New_York:20180119T130000Z RRULE:FREQ=DAILY;INTERVAL=1;COUNT=7')

File "~/venvs/tmp-c74b801dfc74a8d7/lib/python2.7/site-packages/dateutil/rrule.py", line 1616, in __call__
  return self._parse_rfc(s, **kwargs)
File "~/venvs/tmp-c74b801dfc74a8d7/lib/python2.7/site-packages/dateutil/rrule.py", line 1569, in _parse_rfc
  raise ValueError("unsupported DTSTART parm: "+parm)
ValueError: unsupported DTSTART parm: TZID=AMERICA/NEW_YORK
```

From https://www.rfc-editor.org/rfc/rfc5545.txt:

>       The following are examples of this property parameter:
> 
>        DTSTART;TZID=America/New_York:19980119T020000
> 
>        DTEND;TZID=America/New_York:19980119T030000
> 
>       The "TZID" property parameter MUST NOT be applied to DATE
>       properties and DATE-TIME or TIME properties whose time values are
>       specified in UTC.

Is there a way to accomplish this in `dateutil` that I'm unaware of?

pganssle:
@ryanpetrello Well, for one thing those `TZID`s are supposed to refer to a a time zone defined elsewhere in the `iCalendar`, they're not supposed to be generic and pulled from a database or something (though I think a lot of people probably *do* use IANA zones like that, and allow the definition to be implicit).

Additionally, your example is actually specifying *two* time zones here, because the datetime is specified as `20180119T130000Z` - where `Z` means "UTC". So I think there would be a conflict either way.

For now, there is a non-ideal way to get what you want, which is to encode the time zone into the `DTSTART` parameter in a different way (unfortunately you have to use the 3-letter acronym because of a well-known limitation of `dateutil.parser.parse`).

```python
from dateutil import rrule
NEW_YORK = tz.gettz('America/New_York')

rrstr = ('DTSTART:20180119T130000EST ' +
         'RRULE:' + 
            'FREQ=DAILY;' +
            'INTERVAL=1;' +
            'COUNT=7')

rr = rrule.rrulestr(rrstr, tzinfos={'EST': NEW_YORK, 'EDT': NEW_YORK)})
```

[...]

PR #615, opened 2018-01-18
add support for DTSTART;TZID= in rrules

ryanpetrello:
see: https://github.com/dateutil/dateutil/issues/614

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -20,6 +20,7 @@ from six.moves import _thread, range
 import heapq
 
 from ._common import weekday as weekdaybase
+from . import tz
 
 # For warning about deprecation of until and count
 from warnings import warn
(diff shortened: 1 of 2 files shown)


pganssle:
Hmm... This is the right *general* idea, but the specific implementation is wrong.

What needs to happen is that `TZID=America/New_York` is split into "TZID" (which is validated) and "America/New_York", and then you use the same logic that `parser.parse` uses when evaluating `tzinfos` to parse `America/New_York`, something like this:

```python
# Set TZID = None somewhere before any conditionals
# Assume that TZID=(somevalue) has been parsed such that tzkey=(somevalue)
_tzinfos = tzinfos if tzinfos is None else tz.gettz
if not callable(_tzinfos):
    _tzinfos = _tzinfos.get
TZID = _tzinfos(tzkey)
```

Then once you've parsed `dtstart` do something like this:

```python
if TZID is not None:
    if dtstart.tzinfo is None:
        dtstart.replace(tzinfo=TZID)
    else:
        dtstart.astimezone(TZID)
```

Alternatively, in the `else` branch there, you can raise an exception (e.g. `'DTSTART specifies time zone twice'`). What we do depends on the RFC.

pganssle:
@ryanpetrello Let me know if you are planning on making some of the required changes and I can re-open this.

ryanpetrello:
@pganssle you mind re-opening this?  I've uploaded some new changes that I think should address your feedback.

pganssle:
Hm.. Apparently you force-pushed so I can't re-open the PR. [See this issue](https://github.com/isaacs/github/issues/361). I don't totally understand why, but I guess the solution is to just make a fresh PR. Sorry, didn't realize this could happen when I closed the PR.

ryanpetrello:
No problem, I'll make a new PR.

PR #619, opened 2018-01-26
add support for DTSTART;TZID= in rrules

ryanpetrello:
see: https://github.com/dateutil/dateutil/issues/614

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -8,6 +8,7 @@ including support for caching of results.
 import itertools
 import datetime
 import calendar
+import re
 import sys
 
 try:
@@ -1502,6 +1503,11 @@ class _rrulestr(object):
         if compatible:
             forceset = True
             unfold = True
+
+        TZID_NAMES = dict(map(
+            lambda x: (x.upper(), x),
+            re.findall('TZID=(?P<name>[^:]+):', s)
+        ))
         s = s.upper()
         if not s.strip():
             raise ValueError("empty string")
@@ -1568,8 +1574,16 @@ class _rrulestr(object):
                     valid_values = {"VALUE=DATE-TIME", "VALUE=DATE"}
                     for parm in parms:
                         if parm.startswith("TZID="):
-                            tzkey = parm.split('TZID=')[-1]
-                            tzlookup = tzinfos.get if tzinfos else tz.gettz
+                            try:
+                                tzkey = TZID_NAMES[parm.split('TZID=')[-1]]
+                            except KeyError:
+                                continue
+                            if callable(tzinfos):
+                                tzlookup = tzinfos
+                            elif isinstance(tzinfos, dict):
+                                tzlookup = tzinfos.get
+                            else:
+                                tzlookup = tz.gettz
                             TZID = tzlookup(tzkey)
                             continue
                         if parm not in valid_values:
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -20,6 +20,7 @@ from six.moves import _thread, range
 import heapq
 
 from ._common import weekday as weekdaybase
+from . import tz
 
 # For warning about deprecation of until and count
 from warnings import warn
@@ -1563,8 +1564,14 @@ class _rrulestr(object):
                     # RFC 5445 3.8.2.4: The VALUE parameter is optional, but
                     # may be found only once.
                     value_found = False
+                    TZID = None
                     valid_values = {"VALUE=DATE-TIME", "VALUE=DATE"}
                     for parm in parms:
+                        if parm.startswith("TZID="):
+                            tzkey = parm.split('TZID=')[-1]
+                            tzlookup = tzinfos.get if tzinfos else tz.gettz
+                            TZID = tzlookup(tzkey)
+                            continue
                         if parm not in valid_values:
                             raise ValueError("unsupported DTSTART parm: "+parm)
                         else:
@@ -1577,6 +1584,11 @@ class _rrulestr(object):
                         from dateutil import parser
                     dtstart = parser.parse(value, ignoretz=ignoretz,
                                            tzinfos=tzinfos)
+                    if TZID is not None:
+                        if dtstart.tzinfo is None:
+                            dtstart = dtstart.replace(tzinfo=TZID)
+                        else:
+                            raise ValueError('DTSTART specifies multiple timezones')
                 else:
                     raise ValueError("unsupported property: "+name)
             if (forceset or len(rrulevals) > 1 or rdatevals
--- a/dateutil/test/test_rrule.py
+++ b/dateutil/test/test_rrule.py
@@ -2679,6 +2679,10 @@ class RRuleTest(WarningTestMixin, unittest.TestCase):
                           datetime(1998, 9, 2, 9, 0, tzinfo=tz.gettz('America/New_York')),
                           datetime(1999, 9, 2, 9, 0, tzinfo=tz.gettz('America/New_York'))])
 
+    def testStrWithConflictingTZID(self):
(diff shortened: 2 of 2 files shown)


ryanpetrello:
@pganssle one oddity I've noticed - and this is why the tests are failing on Travis-CI but they pass for me locally.  When you parse the string this way, `tzkey` gets set to `AMERICA/NEW_YORK`, because of the entire string is uppercased here:  https://github.com/dateutil/dateutil/blame/master/dateutil/rrule.py#L1504

This causes the actual `tz.gettz` call to fail (he test passes on my OS X file system due to case insensitivity).

dateutil/rrule.py
```diff
@@ -1563,8 +1564,14 @@ def _parse_rfc(self, s,
                     # RFC 5445 3.8.2.4: The VALUE parameter is optional, but
                     # may be found only once.
                     value_found = False
+                    TZID = None
                     valid_values = {"VALUE=DATE-TIME", "VALUE=DATE"}
                     for parm in parms:
+                        if parm.startswith("TZID="):
+                            tzkey = parm.split('TZID=')[-1]
+                            tzlookup = tzinfos.get if tzinfos else tz.gettz
```
pganssle: I think it's best to explicitly compare to `None`, since I would think that `tzinfos={}` should not be equivalent to `tzinfos=tz.gettz`.

ryanpetrello: I guess I'm confused about _when_ you would want to use `tz.gettz`.  Wouldn't it be _any_ time `tzinfos` *wasn't* a dictionary with keys?

pganssle:
@ryanpetrello Hmmm.. That might be a tough one to fix if most of the rest of the function is assuming the string is already upper-cased.

dateutil/rrule.py
pganssle: @ryanpetrello I am fine with `tz.gettz` being the default if you don't specify anything, but `tzinfos` can also be a callable, like this:

```python
def my_tzinfos(tzid):
    if tzid in SOME_LOOKUP_TABLE:
        return tzid[SOME_LOOKUP_TABLE]
    else:
        try:
            return dateutil.tz.tzstr(tzid)
        except ValueError:
            return tz.tzutc()
```

I don't really know if that particular function is something someone would actually want to do, but the point is that `tzinfos` is basically a mapping between `tzid` and a `tzinfo` object, so any function that provides `tzinfo` as a function of `tzid` should work. As a convenience, you can also provide a dictionary directly, rather than passing `mytzinfos.get`.

ryanpetrello:
@pganssle I don't _love_ this, but I can't think of another way to do it without changing the parsing algorithm significantly:

```diff
diff --git a/dateutil/rrule.py b/dateutil/rrule.py
index 2f2511c..00edaff 100644
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -1502,6 +1502,7 @@ class _rrulestr(object):
         if compatible:
             forceset = True
             unfold = True
+        orig_s = s
         s = s.upper()
         if not s.strip():
             raise ValueError("empty string")
@@ -1569,6 +1570,13 @@ class _rrulestr(object):
                     for parm in parms:
                         if parm.startswith("TZID="):
                             tzkey = parm.split('TZID=')[-1]
+                            # the string is uppercase, so we have to find the
+                            # TZID=value in its original case
+                            try:
+                                tzpos = orig_s.upper().index(tzkey)
+                                tzkey = orig_s[tzpos:tzpos+len(tzkey)]
+                            except IndexError:
+                                continue
                             tzlookup = tzinfos.get if tzinfos else tz.gettz
                             TZID = tzlookup(tzkey)
                             continue

```

dateutil/rrule.py
ryanpetrello:
```python
if callable(tzinfos):
    tzlookup = tzinfos
elif isinstance(tzinfos, dict):
    tzlookup = tzinfos.get
else:
    tzlookup = tz.gettz
```

More like this?

ryanpetrello:
@pganssle is this current patch looking any better to you?

pganssle:
@ryanpetrello Sorry, just had a baby. It will be a bit before I can review this PR.

ryanpetrello:
👶👶👶 congrats!

Don't let me bother you, then - get to this whenever you find time 😄

dateutil/rrule.py
```diff
@@ -1563,8 +1570,22 @@ def _parse_rfc(self, s,
...
+                        if parm.startswith("TZID="):
+                            try:
+                                tzkey = TZID_NAMES[parm.split('TZID=')[-1]]
+                            except KeyError:
+                                continue
+                            if callable(tzinfos):
+                                tzlookup = tzinfos
+                            elif isinstance(tzinfos, dict):
+                                tzlookup = tzinfos.get
+                            else:
```
pganssle: Once again I don't think this is the right thing to do. `tzlookup` should be set to `tz.gettz` if and only if the user requested the *default* behavior.

I would probably do it this way:

```python
if tzinfos is None:
    tzlookup = tz.gettz
elif callable(tzinfos):
    tzlookup = tzinfos
else:
    tzlookup = getattr(tzinfos, 'get', None)
    if tzlookup is None:
        raise ValueError('tzinfos must be a callable or mapping')
```

dateutil/rrule.py
```diff
@@ -1501,6 +1503,11 @@ def _parse_rfc(self, s,
         if compatible:
             forceset = True
             unfold = True
+
+        TZID_NAMES = dict(map(
```
pganssle: Both `dict` and `map` are very slow compared to a dictionary comprehension in this case:

```python
TZID_NAMES = {x.upper(): x
              for x in re.findall('TZID=(?P<name>[^:]+):', s)}
```

That said, I don't *love* this approach (though it *may* be the best option, unfortunately). I'd fix the comprehension part but don't worry about the regex part - we can fix that with a bigger refactor later I think.

dateutil/rrule.py
```diff
@@ -1577,6 +1598,11 @@ def _parse_rfc(self, s,
                         from dateutil import parser
                     dtstart = parser.parse(value, ignoretz=ignoretz,
                                            tzinfos=tzinfos)
+                    if TZID is not None:
+                        if dtstart.tzinfo is None:
+                            dtstart = dtstart.replace(tzinfo=TZID)
+                        else:
+                            raise ValueError('DTSTART specifies multiple timezones')
```
pganssle: Did you happen to check of the RFC mentions whether or not this is allowed?

This is one of the two reasonable options - the other being that if `DTSTART` is timezone-aware (e.g. `20180204T0800Z`) and the rule has a specific time zone, you convert `DTSTART` into that time zone and generate dates starting from there. If the RFC is not silent on this point, we could choose either one, really, this one being more conservative and so probably preferable for a first release.

dateutil/test/test_rrule.py
```diff
@@ -2669,6 +2670,19 @@ def testStr(self):
                           datetime(1998, 9, 2, 9, 0),
                           datetime(1999, 9, 2, 9, 0)])
 
+    def testStrWithTZID(self):
```
pganssle: I think you should add some tests for actually specifying the `tzinfos` parameter, and especially the case of `tzinfos={}` (which should fail to parse any `TZID` specified).

pganssle:
@ryanpetrello I was going to push directly to your branch, but you made this PR from the master branch of your fork, which is a PITA because I have my own master branch on my local fork, and it means that I can't easily track your upstream branch. In the future it's best to create a branch on your fork before making a PR.

My updated version of this TZID support is in #624, and is slated for inclusion in 2.7.0 (which should go out this weekend).

PR #621, opened 2018-02-21
Update test to not fail after 2027

bmwiedemann:
because then 2076 is closer than 1976

The more general fix would be to add something like
```python
        time.localtime = mock.Mock()
        time.localtime.return_value = time.struct_time(
            [2017, 11, 28, 1, 1, 1, 1, 332, 0]
        )
```


pganssle:
I wonder if it might be better to just use `freeze_gun` on this rather than change the test itself.

bmwiedemann:
it is your choice... I'm fine with having a trivial one-line patch. less likely to break things

pganssle:
OK, I'm going to close this because we have a lot of time to fix this and I'd like to dramatically refactor these tests to use `freeze_gun` anyway.

PR #626, opened 2018-03-11
Drop support for Python 3.2

pganssle:
I *think* Python 3.2 still works, but given it gets almost no downloads and it's nearly impossible to test as we take on more test dependencies that have dropped Python 3.2 support (and AFAIK `python_requires` is not supported by a version of `pip` that also supports Python 3.2).


pganssle:
Travis doesn't currently support Python 3.7, but I ran `tox -e py37` locally, and it worked, so I'm pretty confident adding Python 3.7 to the list.

jbrockmendel:
+1

PR #628, opened 2018-03-11
More caching and singletons in tz

pganssle:
For the upcoming 2.7.0 release, I think it's important for at least `tz.gettz` to return the same zone for the same inputs unless otherwise specified.

This PR makes `tzstr` a singleton, adds caching to `tz.gettz` and adds the `.instance` alternate constructor for `tzoffset` and `tzstr`.

I did not bother with a `.instance` alternate constructor for `tzutc`, because the semantics of `tzutc` are not affected by inter- and intra-zone comparisons (technically true of `tzoffset` as well, too).


pganssle:
This partially satisfies #342

Issue #633, opened 2018-03-12
rrule parsing allows `UNTIL` with non-UTC dates


ryanpetrello:
```shell
~ cat utc_until.py && python utc_until.py
```

```python
from datetime import datetime
import dateutil.rrule

rrule = dateutil.rrule.rrulestr(
    'DTSTART;TZID=America/New_York:20180601T120000 RRULE:FREQ=DAILY;INTERVAL=1;UNTIL=20180602T000000'
)
print rrule.after(datetime.utcnow())
```

```python
Traceback (most recent call last):
  File "utc_until.py", line 7, in <module>
    print rrule.after(datetime.utcnow())
  File "/home/ryan/.local/share/virtualenvs/tmp.jQo0RUu68U-Zv6c7Bw0/lib/python2.7/site-packages/dateutil/rrule.py", line 223, in after
    for i in gen:
  File "/home/ryan/.local/share/virtualenvs/tmp.jQo0RUu68U-Zv6c7Bw0/lib/python2.7/site-packages/dateutil/rrule.py", line 862, in _iter
    if until and res > until:
TypeError: can't compare offset-naive and offset-aware datetimes
```

According to RFC5545 Section 3.3.10:
https://tools.ietf.org/html/rfc5545#section-3.3.10

> If the "DTSTART" property is specified as a date with UTC time or a date with local time and time zone reference, then the UNTIL rule part MUST be specified as a date with UTC time.

ryanpetrello:
(I'm working on a PR to address this)

pganssle:
I believe this is a duplicate of #620

Issue #620, opened 2018-02-06
rrule _until missing timezone info


jeinarsson:
I have a calendar (generated by Google) with a line like so:

> RRULE:FREQ=MONTHLY;UNTIL=20180819;INTERVAL=12;BYDAY=3MO

If I use `rrulestr` to parse this recurrence with a `dtstart` **including timezone info**, the generator fails, because it tries a comparison with the `_until` field which has no timezone info.

Here's a minimal example of what I mean:

```python
from datetime import datetime, timezone
from dateutil.rrule import *

rule_str = 'FREQ=MONTHLY;UNTIL=20180819;INTERVAL=12;BYDAY=3MO'

# this works (no timezone data)
start = datetime(2017,8,21)
rule = rrulestr(rule_str, dtstart=start)
print(list(rule))

# but this does not
start = datetime(2017,8,21,tzinfo=timezone.utc)
rule = rrulestr(rule_str, dtstart=start)
print(list(rule))
```

Output:

```
[datetime.datetime(2017, 8, 21, 0, 0)]
Traceback (most recent call last):
  File "mwe_rrule.py", line 17, in <module>
    print(list(rule))
  File "/usr/local/lib/python3.6/site-packages/dateutil/rrule.py", line 861, in _iter
    if until and res > until:
TypeError: can't compare offset-naive and offset-aware datetimes
```

EDIT: I monkey patched rrule.py after line 446: 
https://github.com/dateutil/dateutil/blob/1c8c90338a696c8c146c7eb6cc1bf52016912852/dateutil/rrule.py#L446
by adding
```
        if until:
            until = until.replace(tzinfo=self._tzinfo)
```

If you're interested I could make a PR, but maybe you have a better way in mind.

jeinarsson:
Thanks for your work and answer.

The RFC actually says that UNTIL must be the same type as DTSTART, specifically (p.40)

> The value of the UNTIL rule part MUST have the same
> value type as the "DTSTART" property.  Furthermore, if the
> "DTSTART" property is specified as a date with local time, then
> the UNTIL rule part MUST also be specified as a date with local
> time.  If the "DTSTART" property is specified as a date with UTC
> time or a date with local time and time zone reference, then the
> UNTIL rule part MUST be specified as a date with UTC time.  In the
> case of the "STANDARD" and "DAYLIGHT" sub-components the UNTIL
> rule part MUST always be specified as a date with UTC time.  If
[...]

PR #634, opened 2018-03-12
require UTC for UNTIL when DTSTART is timezone-aware

ryanpetrello:
related: https://github.com/dateutil/dateutil/issues/620
related: https://github.com/dateutil/dateutil/issues/633

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -21,6 +21,7 @@ from six.moves import _thread, range
 import heapq
 
 from ._common import weekday as weekdaybase
+from .tz import tzutc, tzlocal
 
 # For warning about deprecation of until and count
 from warnings import warn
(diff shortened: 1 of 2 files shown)


ryanpetrello:
@pganssle I think I've addressed your feedback at https://github.com/ryanpetrello/dateutil/commit/a9007c4c68d7bc610d67681229b8a9707bc7c10d#commitcomment-28049100

dateutil/rrule.py
```diff
@@ -447,6 +448,23 @@ def __init__(self, freq, dtstart=None,
             until = datetime.datetime.fromordinal(until.toordinal())
         self._until = until
 
+        if self._dtstart and self._until:
+            if all((
```
pganssle: I think this might be better as:

`if (self._dtstart.tzinfo is not None) != (self._until.tzinfo is not None)`

That's basically the condition under which it would fail because of the comparison anyway, and I don't want to enforce that it's specifically `tzutc` (what if you used `datetime.timezone.utc` or `pytz.UTC`). I think being slightly less strict about the specifics of the RFC is worth it in this case, since I want to tread lightly.

I'll note that there's a very strange edge case that would occur if `self._dtstart.tzinfo is self._until.tzinfo` and the UNTIL value is an ambiguous time, but I'm not terribly worried about that.

pganssle: Oh, and one thing to note is that my formulation also catches the case where *only* UNTIL is aware.

ryanpetrello: Good point - addressed 👍

dateutil/rrule.py
```diff
@@ -447,6 +448,20 @@ def __init__(self, freq, dtstart=None,
             until = datetime.datetime.fromordinal(until.toordinal())
         self._until = until
 
+        if self._dtstart and self._until:
```
pganssle: There's no need for this line. The problem is with the incompatibility between DTSTART and UNTIL. Not sure what the RFC has to say, but if UNTIL is aware and DTSTART has no specified time zone, you still don't know when to end the rule.

ryanpetrello: It seems necessary to me.  If I change this and run the tests:

```python
>       if (self._dtstart.tzinfo is not None) != (self._until.tzinfo is not None):
E       AttributeError: 'NoneType' object has no attribute 'tzinfo'
```

ryanpetrello: ...or at the very least, we need to check truthiness of `self._until`.

pganssle: Ah, you are right, sorry, I read this as `self._dstart.tzinfo`, my bad.

PR #636, opened 2018-03-12
Catch non-ValueError exceptions in decimal

### LINKED ISSUE #632: InvalidOperation ConversionSyntax after replacing float with Decimal (Refs #570)
Hey there,
Thanks for a great package!

We encountered an issue with 2.7.0 release and exactly this change (float to Decimal):
https://github.com/dateutil/dateutil/pull/570/files#diff-4b65e7edb2b73639d2d3360fa8d05343R937

Before release it was float and producing `ValueError`:
```
>>> float('TEST')
ValueError: could not convert string to float: 'TEST'
```
After release, same operation goes with `Decimal`, which produces a different error:
```
>>> Decimal('TEST')
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
```

I am not 100% familiar with `dateutil` code, but I see, that `ValueError` was covered in `dateutil.parser._parser.parser#_parse` here: https://github.com/dateutil/dateutil/blob/master/dateutil/parser/_parser.py#L689-L825

But now that piece is raising `decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]` which is not covered at all and breaking the code, that uses it.

What shall we do in that case? Shall we wrap it in `try/except` like it says in leftovers `TODO` comment?

My environment (just in case):
```
>>> import dateutil; print(dateutil.__version__)
2.7.0

13:16 $ python --version
Python 3.6.4
```

**Upd.** Code example with failing parser:

```
>dateutil.parser.parse('1: test')
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]

>dateutil.parser.parse('9999: 12, test')
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
```

pganssle:
Fixes #632.

This adds both function call overhead and exception catching overhead to the function. Based on running `%timeit` on a few strings, the cost is on the order of a few microseconds out of 60-100 microseconds. If we can come up with a version that doesn't have to re-raise, that might be preferable.

Also, we could theoretically add whatever errors `Decimal` is raising into the huge try-except loop and handle it that way, but the documentation doesn't seem to clearly enumerate an exception hierarchy for `decimal.Decimal`, so I'm not confident that we could catch the "root" exception easily.

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -836,7 +836,11 @@ class parser(object):
     def _parse_numeric_token(self, tokens, idx, info, ymd, res, fuzzy):
         # Token is a number
         value_repr = tokens[idx]
-        value = Decimal(value_repr)
+        try:
+            value = self._to_decimal(value_repr)
+        except Exception as e:
+            six.raise_from(ValueError('Unknown numeric token'), e)
+
         len_li = len(value_repr)
 
         len_l = len(tokens)
@@ -895,7 +899,7 @@ class parser(object):
         elif idx + 2 < len_l and tokens[idx + 1] == ':':
             # HH:MM[:SS[.ss]]
             res.hour = int(value)
-            value = Decimal(tokens[idx + 2])  # TODO: try/except for this?
+            value = self._to_decimal(tokens[idx + 2])  # TODO: try/except for this?
             (res.minute, res.second) = self._parse_min_sec(value)
 
             if idx + 4 < len_l and tokens[idx + 3] == ':':
@@ -996,7 +1000,7 @@ class parser(object):
 
     def _assign_hms(self, res, value_repr, hms):
         # See GH issue #427, fixing float rounding
-        value = Decimal(value_repr)
+        value = self._to_decimal(value_repr)
 
         if hms == 0:
             # Hour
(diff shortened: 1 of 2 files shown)


pganssle:
@amureki Looks good?

dateutil/parser/_parser.py
```diff
@@ -895,7 +899,7 @@ def _parse_numeric_token(self, tokens, idx, info, ymd, res, fuzzy):
         elif idx + 2 < len_l and tokens[idx + 1] == ':':
             # HH:MM[:SS[.ss]]
             res.hour = int(value)
-            value = Decimal(tokens[idx + 2])  # TODO: try/except for this?
+            value = self._to_decimal(tokens[idx + 2])  # TODO: try/except for this?
```
amureki: Shall we remove `TODO` comment from here? Since we are covering this in `_to_decimal`

pganssle: Well, `_to_decimal` can still raise a `ValueError`, and that `TODO` predates the switch to `Decimal`:

https://github.com/dateutil/dateutil/commit/3b43612c35c7d0918ae96d132a698f7e19c7cdff#diff-4b65e7edb2b73639d2d3360fa8d05343L935

That said, I don't really know what it refers to. I'll leave it in for now.

PR #643, opened 2018-03-24
Update tzdata to 2018d

pganssle:
Bumps zoneinfo data.


pganssle:
Failure on Windows is because in 2018d there was a correction to the Kiritimati data about which day was imaginary.

Annoyingly, the Olson database provides no versioning (even though `zonefile` does), which means we can't just dispatch the date based on the version.

PR #645, opened 2018-03-24
Add towncrier for changelog management

### LINKED ISSUE #592: Add changelog management

pganssle:
Fixes #592.

At the moment town crier seems like the best option out there for changelog management. It does pretty much what I want, except I don't think the ticket-based workflow definitely works for this project.

Ideally we'd have something like town crier that allows additional metadata to be specified in the `.rst` so that I can add "reporter", "pr", "issue" and "patch submitter". Also, it would be nice to be able to specify an "issue template" so users could do something like `towncrier entry` and it would maybe ask a few questions then generate a basic template.

Still, this is waaaay better than what I've been doing, and seem better than anything else out there.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,32 @@ If you would like to fix something in `dateutil` -  improvements to documentatio
 
 The most important thing to include in your pull request are *tests* - please write one or more tests to cover the behavior you intend your patch to improve. Ideally, tests would use only the public interface - try to get 100% difference coverage using only supported behavior of the API.
 
+#### Changelog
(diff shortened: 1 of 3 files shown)


jbrockmendel:
@pganssle Looks like you're on a roll today.  I can't keep up with everything in real-time, but let me know if there's anything in particular you'd like an extra pair of eyes on.

pganssle:
@jbrockmendel I'm mostly doing metadata housekeeping stuff in preparation for a 2.7.1 release.

I'd like to move towards more frequent releases, so I'm trying to make that as painless as possible. Now that changelog management is done, I just need to add a good workflow for maintaining "patch", "minor" and "major" branches in such a way that it's simple to backport various functionality.

PR #655, opened 2018-03-26
Update readme links to pypi.org

pganssle:
Warehouse seems definitely mature enough that we should be linking to it rather than the old PyPI.

Correct me if I'm wrong @ewdurbin, but I think it's time for this.


pganssle:
@ewdurbin Thanks for the status update. I'm not terribly worried since I doubt that the dateutil docs/github page are a major driver of traffic to the PyPI page, and people can always still google for it.

Issue #443, opened 2017-08-11
Examples in documentation should be broken up by module


pganssle:
Currently the documentation has a huge monolithic ["examples" page](https://dateutil.readthedocs.io/en/latest/examples.html) that covers how a bunch of the classes and functions are used. I think it would be much easier to stumble on if these examples were moved into the base page for their respective modules (e.g. the module docstring, for the most part - though maybe we want to duplicate *some* of the examples in the class/function docstrings).

If this is done, to avoid breaking old links, the "examples" page should be replace with links to the examples anchor on each of the respective module pages.

colfax23:
I'll take this one

pganssle:
After seeing the results of a lot of the moves here, I am thinking this is not as good an idea as I had originally thought.

PR #671, opened 2018-04-12
Added examples to tzfile docstring

weatherpattern:
Added examples to the tzfile to the docstring.
Note sure if there was a format you wanted.

Relates to #443

### DIFF
--- a/changelog.d/671.doc.rst
+++ b/changelog.d/671.doc.rst
@@ -1,0 +1,1 @@
+Add doctest examples to tzfile documentation. Done by @weatherpattern and @pganssle (gh pr #671)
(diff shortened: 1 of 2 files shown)


pganssle:
I think these need to be in `.. code-block` or `.. doctest` blocks.

That said, I don't think people are supposed to directly create `tzfile` objects, so I think these examples might need a bit of context. I can handle that part, though.

weatherpattern:
Thanks for taking a look. 
I am not familiar this doctest blocks but based on the [sphinx doctest documentation](http://www.sphinx-doc.org/en/stable/ext/doctest.html), I tried to match the example in tzrange.

PR #672, opened 2018-04-13
Updates Tz Singletons and Caches with `WeakValueDictionary`s

### LINKED ISSUE #635: Switch timezone singletons and caches with weak references
Right now `tzoffset`, `tzstr` and `tz.gettz` each retain strong references to every unique time zone ever constructed by that method. I think in the end this will be a net savings in memory in *most* use cases since it prevents duplication of time zones, but it could also cause a huge proliferation of memory in applications that make short-lived references to a large number of time zones (since those time zones, when they go out of scope, would normally get garbage collected).

In Python 3 there is the [weakref](https://docs.python.org/3/library/weakref.html) module, which allows the creation of weak references, such that an object will be garbage collected if only weak references to it exist.

Rather than try and use `backports.weakref` in Python 2, I think we can just write a "pass through" compatibility shim that has the same external form as `weakref` but simply holds a strong reference to the object in question.

cs-cordero:
Closes #635.

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -14,6 +14,7 @@ import unittest
 import sys
 import base64
 import copy
+import gc
 
 from functools import partial
 
@@ -731,6 +732,20 @@ class TzOffsetTest(unittest.TestCase):
 
         assert tz1 is tz2
 
+    def testTzOffsetWeakRef(self):
+        UTC1 = tz.tzoffset('UTC', 0)
+        UTC2 = tz.tzoffset('UTC', 0)
+        assert id(UTC1) == id(UTC2)
+
+        weak_ref_id = id(UTC1)
+        del UTC1
+        del UTC2
+        gc.collect()
+
+        UTC3 = tz.tzoffset('UTC', 0)
+        assert not weak_ref_id == id(UTC3)
+
+
 @pytest.mark.tzoffset
 @pytest.mark.parametrize('args', [
     ('UTC', 0),
@@ -1036,6 +1051,21 @@ class GettzTest(unittest.TestCase, TzFoldMixin):
         local2 = tz.gettz()
 
         assert local1 is not local2
+
+    def testGettzWeakRef(self):
+        tz.gettz.cache_clear()
+        NYC1 = tz.gettz('America/New_York')
+        NYC2 = tz.gettz('America/New_York')
+        assert id(NYC1) == id(NYC2)
+
+        weak_ref_id = id(NYC1)
+        del NYC1
+        del NYC2
+        gc.collect()
+
+        NYC3 = tz.gettz('America/New_York')
+        assert not weak_ref_id == id(NYC3)
+
 
 @pytest.mark.gettz
 @pytest.mark.xfail(IS_WIN, reason='zoneinfo separately cached')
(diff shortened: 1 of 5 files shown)


dateutil/utils.py
```diff
@@ -4,6 +4,10 @@
 from datetime import datetime, time
 
 
+class FakeWeakRef:
```
pganssle: `.utils` is a public module and very much *not* the right place for this. It should probably be in `tz._common.py`

dateutil/tz/_factories.py
```diff
@@ -1,5 +1,10 @@
 from datetime import timedelta
 
+try:
+    import weakref
```
pganssle: I would change this to:

```python
try:
    from weakref import WeakRefDictionary
except ImportError:
    WeakRefDictionary = dict
```

dateutil/test/test_tz.py
```diff
@@ -14,10 +14,12 @@
 import sys
 import base64
 import copy
+import gc
 
 from functools import partial
 
 IS_WIN = sys.platform.startswith('win')
+NOT_PY3 = not sys.version_info[0] == 3
```
pganssle: Normally you would not have to define this. There's already a `six.PY3` and `six.PY2` that will give you that.

That said, I may have been wrong about weak references being a Python 3 thing. [Seems like they are available in Python 2.7](https://docs.python.org/2.7/library/weakref.html), so we should drop this skip test.

dateutil/tz/_factories.py
pganssle: Actually it turns out that Python 2 has weak references, so we can just drop the whole compatibility shim. My bad.

pganssle:
Hard to tell if this is just holding a strong reference somewhere we don't realize or if the assumptions about one of the inputs is wrong. Seems to only happen on Windows on old versions of Python, though.

cs-cordero:
Sorry for the flurry of seemingly random commits.  I can't seem to replicate AppVeyor's failure on my local machine even after matching the Python version, so I'm trying to troubleshoot with small changes.

cs-cordero:
Just reverted all the troubleshooting commits, but I'm kind of at a loss here for how this is failing on the AppVeyor CI.

The stdout from the 46bdab7 commit is really what's really tripping me up:
https://ci.appveyor.com/project/dateutil/dateutil/build/2.3.1234/job/4ru7kdg7wipal15w

In this commit, I print the keys inside the `WeakValueDictionary` before and after deleting the strong `ATK1` and `ATK2` references, expecting the WeakValueDictionary to clear itself, but it does not.  I also used `pytest.raises(NameError)` to assert that `ATK1` and `ATK2` are indeed deleted, and they are.  There shouldn't be any remaining strong references to the value in the dictionary, but somehow it is persisting.  Running the equivalent code locally off of CI yields the expected behavior.

I'll keep thinking about how to get around this.  One idea I had was to have some kind of `_cache_hit` attribute on the factory that keeps a count of every time it is returning a value from its `WeakValueDictionary`.  We can assert that this counter does/does not change depending on the test.

pganssle:
I'm guessing that this is a real issue with Windows and not a problem with the tests. I'll grab this branch and try debugging it on a Windows VM.

pganssle:
Cleaned up the commit history. Will merge when tests pass.

Thanks so much @cs-cordero!

cs-cordero:
Happy to help. Thanks for taking it home with that clutch Windows `ZoneInfoFile` issue.

pganssle:
Hm.. The Python 3.5 test failed but succeeded on reboot. This "id won't get reused" gambit may be less stable than I hoped.

I just had a new idea for a more robust test, though. Create a new weak reference to the object before deleting it and make sure its referrent is gone after the gc step.

pganssle:
@cs-cordero One last thing, can you add yourself to `AUTHORS.md` - names are in alphabetical order. You are welcome to put your real e-mail and encouraged to put github name. Make sure to put a **D** after your name line, which indicates your contributions are under the dual license.

cs-cordero:
Will do re: `authors.md` once I get access to a computer and little later in the day. Thanks!

pganssle:
@cs-cordero No this test assumes that the `weakref` module works as expected and tests that we didn't screw up and accidentally start holding a strong reference to the `tzinfo` objects in the cache.

If you do:

```python
git checkout master
git checkout c610fc509674a -- dateutil/test/test_tz.py
pytest -k test_tz.py
```

You'll see that these three tests fail, which means the problem - that the cache itself holds a strong reference to the cached time zones - is solved by this patch.

cs-cordero:
Ah yes that makes perfect sense! Thank you for explaining.

pganssle:
@cs-cordero You'll probably need to do a force pull when you edit this branch again, because I've rebased your branch and rewritten the history. If you don't know how to do that it's `git pull --force`

cs-cordero:
🎉

PR #674, opened 2018-04-13
Hypothesis tests on convertyear

coreygirard:
Added property-based tests to convertyear in dateutil/parser/_parser.py

Refactored convertyear


dateutil/parser/_parser.py
```diff
@@ -46,8 +46,8 @@
 
 from warnings import warn
 
-from .. import relativedelta
-from .. import tz
+#from .. import relativedelta
+#from .. import tz
```
jbrockmendel: Please don't leave commented-out code

dateutil/parser/_parser.py
```diff
@@ -364,15 +364,30 @@ def tzoffset(self, name):
...
-            year += self._century
-            if abs(year - self._year) >= 50:
-                if year < self._year:
-                    year += 100
-                else:
-                    year -= 100
+        '''
+        Converts two-digit years to year within [-50, 49]
+        range of self._year (current local time)
+        '''
```
jbrockmendel: """ instead of '''

dateutil/parser/_parser.py
```diff
@@ -364,15 +364,30 @@ def tzoffset(self, name):
...
+        if year - self._year >= 50: # if too far in future
+            year -= 100
+        elif year - self._year < -50: # if too far in past
+            year += 100
+
         return year
 
+
+
+
```
jbrockmendel: Don't add these new whitespace lines.  pep8 convention is 1 blank line between methods.

dateutil/parser/_parser.py
```diff
@@ -364,15 +364,30 @@ def tzoffset(self, name):
...
+        if century_specified or year >= 100:
+            return year
+
+        # assume current century to start
+        year += self._century
+
+        if year - self._year >= 50: # if too far in future
+            year -= 100
+        elif year - self._year < -50: # if too far in past
+            year += 100
```
jbrockmendel: It looks like the main difference from the old version is in the case with `year - self.year == -50`.  Is that intentional?  (possibly clarified in the tests, which I haven't read yet)

dateutil/parser/_parser.py
```diff
@@ -364,15 +364,30 @@ def tzoffset(self, name):
...
+        # only called internally. Should never be negative
+        assert year >= 0
+
+        if century_specified or year >= 100:
+            return year
+
+        # assume current century to start
+        year += self._century
+
+        if year - self._year >= 50: # if too far in future
```
jbrockmendel: pep8 two spaces before an inline comment

dateutil/test/test_parser.py
```diff
@@ -27,6 +30,24 @@
...
+    def test_convertyear(self, n):
+        self.assertEqual(n, _parser.parserinfo().convertyear(n))
+
+    @given(integers(min_value=-50,
+                    max_value=49))
+    def test_no_specified_century(self, n):
+        p = _parser.parserinfo()
+        new_year = p._year + n
+        result = p.convertyear(new_year % 100, century_specified=False)
+        self.assertEqual(result, new_year)
```
jbrockmendel: @pganssle is there a preferred convention for how to do these assertions?  I tend to prefer just `assert result == new_year`

pganssle: @jbrockmendel Yeah, these should be refactored as pytest tests. This was part of a sprint event and I told @coreygirard that it was fine to write them as UnitTest tests and I'll refactor them into pytest tests before merging.

dateutil/parser/_parser.py
pganssle: @jbrockmendel Yeah, so that someone can see my reasoning here, for the boundary of years exactly 50 years ago I think the "tie" should go to the past, not the future, since it's much more likely that you'll be parsing a specific date 50 years in the past than 50 years in the future.

dateutil/test/test_parser.py
```diff
@@ -27,6 +30,24 @@
...
+
+    @given(integers(max_value=-1))
+    def test_assert_non_negative(self, n):
+        with self.assertRaises(AssertionError):
+            _parser.parserinfo().convertyear(n)
```
pganssle: `_parser` is a private module and should not be directly tested. Only test the public interface, so `dateutil.parser.parserinfo`.

PR #681, opened 2018-04-14
661 tzinfos could be None in call to parse

### LINKED ISSUE #661: Allow tzinfos to return None
Currently, if you the result of a `tzinfos` call *explicitly* returns `None`, that will throw an error:

```python
>>> parse('2017-02-03 12:40 MGX', tzinfos={'MGX': None})
...
ValueError: Offset must be tzinfo subclass, tz string, or int offset.
>>> parse('2017-02-03 12:40 RDT', tzinfos={'XXX': None})
[...] UnknownTimezoneWarning: tzname RDT identified but not understood.  Pass `tzinfos`
argument in order to correctly return a timezone-aware datetime.  In a future version,
this raise an exception.
  category=UnknownTimezoneWarning)
datetime.datetime(2014, 2, 3, 2, 3)
```

The whole point of #540  was that implicitly silently failing was a bad behavior, but you should certainly be able to opt in to not caring about the time zone by returning `None`.

Note: As an aside, there's a typo I didn't notice there, it says "In a future version, this raise an exception", should be "this *will* raise an exception."

ParseThis:
## Summary of changes

Return `None` if the user would like to ignore the timezone in the date parse.
Fix typo in warning in `_build_tzaware` 

This is a WIP on #661.
 
Closes #661

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -1125,16 +1125,14 @@ class parser(object):
             tzdata = tzinfos(tzname, tzoffset)
         else:
             tzdata = tzinfos.get(tzname)
-
-        if isinstance(tzdata, datetime.tzinfo):
+        # handle case where tzinfo is paased an options that returns None
+        # eg tzinfos = {'BRST' : None}
+        if isinstance(tzdata, datetime.tzinfo) or tzdata is None:
             tzinfo = tzdata
         elif isinstance(tzdata, text_type):
             tzinfo = tz.tzstr(tzdata)
         elif isinstance(tzdata, integer_types):
             tzinfo = tz.tzoffset(tzname, tzdata)
-        else:
-            raise ValueError("Offset must be tzinfo subclass, "
-                             "tz string, or int offset.")
         return tzinfo
 
     def _build_tzaware(self, naive, res, tzinfos):
@@ -1170,7 +1168,7 @@ class parser(object):
             warnings.warn("tzname {tzname} identified but not understood.  "
                           "Pass `tzinfos` argument in order to correctly "
                           "return a timezone-aware datetime.  In a future "
-                          "version, this raise an "
+                          "version, this will raise an "
                           "exception.".format(tzname=res.tzname),
                           category=UnknownTimezoneWarning)
             aware = naive
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -170,7 +170,6 @@ class ParserTest(unittest.TestCase):
                                    tzinfos={"BRST": long(-10800)}),
                              datetime(2003, 9, 25, 10, 36, 28,
                                       tzinfo=self.brsttz))
-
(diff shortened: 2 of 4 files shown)


pganssle:
@ParseThis This PR came in just before we added a PR template with further instructions. I edited your comment into the new PR template.

There are a few housekeeping items left. You mention that this is a WIP - is there more to do other than the changelog and AUTHORS.md?

dateutil/parser/_parser.py
```diff
@@ -1115,16 +1115,18 @@ def _build_tzinfo(self, tzinfos, tzname, tzoffset):
             tzdata = tzinfos(tzname, tzoffset)
         else:
             tzdata = tzinfos.get(tzname)
+            # user opted to not pass in tzinfos
```
pganssle: I don't think "user opted not to pass in tzinfos" is correct here. I believe this line is only reached if you explicitly pass something that returns `None`.

dateutil/parser/_parser.py
```diff
@@ -1115,16 +1115,18 @@ def _build_tzinfo(self, tzinfos, tzname, tzoffset):
...
+            # or tzinfos is set to None  e.g { 'BRST' : None}
+            # don't punish them. 
+            if tzdata is None:
+                return
```
pganssle: I prefer to avoid multiple returns if possible. I would probably refactor this by changing line 1124 below to:

`if isinstance(tzdata, datetime.tzinfo) or tzdata is None:`

dateutil/test/test_parser.py
```diff
@@ -748,6 +747,11 @@ def testUnspecifiedDayFallbackFebLeapYear(self):
         self.assertEqual(parse("Feb 2008", default=datetime(2010, 1, 31)),
                          datetime(2008, 2, 29))
 
+    def testTzinfoCouldbeNone(self):
```
pganssle: Can you also add a separate test that checks that `lambda *args: None` also works without raising an error?

dateutil/parser/_parser.py
jbrockmendel: +1

dateutil/parser/_parser.py
```diff
@@ -1160,7 +1162,7 @@ def _build_tzaware(self, naive, res, tzinfos):
             warnings.warn("tzname {tzname} identified but not understood.  "
                           "Pass `tzinfos` argument in order to correctly "
                           "return a timezone-aware datetime.  In a future "
-                          "version, this raise an "
+                          "version, this will raise an "
```
jbrockmendel: Nice catch.

ParseThis:
@pganssle Nothing else, since its my first PR I wanted to sure of the standard here. Thanks for editing the comments! Everything else you mentioned makes sense and will be added.

pganssle:
This looks good to me. I cleaned up the history a bit, but once the tests pass I'll merge this.

Thanks for your work on this!

PR #693, opened 2018-04-18
Issue 652

absreim:
## Summary of changes

Adds a feature when initializing an rrule with a specified UNTIL but without an explicitly specified DTSTART so that the generated DTSTART takes on the time zone of the UNTIL. Without this feature in that situation, rrule would raise a ValueError because the generated DTSTART is not time zone aware.

Closes 652

### Pull Request Checklist
- [X] Changes have tests


dateutil/test/test_rrule.py
```diff
@@ -14,6 +14,9 @@
     MO, TU, WE, TH, FR, SA, SU
 )
 
+from freezegun import freeze_time
+
+import pytest
```
pganssle: This isn't used.

PR #695, opened 2018-04-20
change the order of cache and byeaster in rrule docstring

### LINKED ISSUE #686: Ordering in the rrule documentation keywords explanation
There is a difference in the order in which the arguments are listed in the function definition(at the beginning of the documentation), and the order in which they are explained. Specifically, cache is the last argument given when we are defining the function in the documentation, but is the first one to be explained.  (http://dateutil.readthedocs.io/en/stable/rrule.html). Another out-of-order keyword is byeaster.

Should we change the ordering in http://dateutil.readthedocs.io/en/stable/_modules/dateutil/rrule.html#rrule to make it consistent?

rmahajan14:
Changed order in the rrule docstring to match with order of the keywords.
Closes #686

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -67,6 +67,7 @@ switch, and thus all their contributions are dual-licensed.
 - Peter Bieringer <pb@MASKED>
 - Pierre Gergondet <pierre.gergondet@MASKED> (gh: @gergondet)
 - Quentin Pradet <quentin@MASKED>
+- Ridhi Mahajan <ridhikmahajan@MASKED> **D**
 - Roy Williams <rwilliams@MASKED>
 - Rustem Saiargaliev (gh: @amureki) **D**
 - Savraj <savraj@MASKED>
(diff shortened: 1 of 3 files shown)


pganssle:
@rmahajan14 You need to add yourself to the AUTHORS.md file and add a changelog.d.

AUTHORS.md
```diff
@@ -1,4 +1,4 @@
-This is a (possibly incomplete) list of all the contributors to python-dateutil,
+﻿This is a (possibly incomplete) list of all the contributors to python-dateutil,
```
pganssle: There are a bunch of weird unprintable characters at the beginning of this line. Is this something your editor is doing?

```python
with open('AUTHORS.md', 'rb') as f:
    print(f.read(60))
# Result:
# b'\xef\xbb\xbfThis is a (possibly incomplete) list of all the contribut'
```

Since the sequence is `0xEFBBBF`, I'm guessing that it's a Windows [Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8) indicating UTF-8. I recommend that you fix your editor so that it doesn't do this anymore.

PR #698, opened 2018-04-21
Fix for B.Y.d format corner case #687

jbrockmendel:
## Summary of changes

At the moment `_ymd.resolve_ymd` can ignore `ystridx` and `dstridx` in some corner cases.  This PR has resolve_ymd check if enough of [ystridx, mstridx, dstridx] have been specified to uniquely pin down (year, month, day), and if so, use those instead of other heuristics.

Closes #687

### Pull Request Checklist
- [x] Changes have tests
- [x] Authors have been added to [AUTHORS.md](https://github.com/dateutil/dateutil/blob/master/AUTHORS.md)
- [x] News fragment added in changelog.d. See [CONTRIBUTING.md](https://github.com/dateutil/dateutil/blob/master/CONTRIBUTING.md#changelog) for details

PR #700, opened 2018-04-21
Fix for B.Y.d format corner case #687

### LINKED ISSUE #687: Parser error with unambiguous date in year 31
I started converting [this test](https://github.com/dateutil/dateutil/blob/master/dateutil/test/test_parser.py#L33) over to a hypothesis-based property test and found the following failure:

```python
from datetime import datetime

dt = datetime(31, 12, 30)
dstr = dt.strftime('%B.%04Y.%d')    # December.0031.30
parse(dt.strftime('%B.%04Y.%d'))     # datetime(30, 12, 31, 0, 0)
```

Pretty low priority, but related to #293. Here's the full hypothesis test that generates these examples:

```python
from datetime import datetime, time
import itertools

from hypothesis.strategies import integers, dates
from hypothesis import given

import pytest

from dateutil.parser import parserinfo, parse


def _get_ybd_formats():
    # If we have a 4-digit year, a non-numeric month (abbreviated or not),
    # and a day (1 or 2 digits), then there is no ambiguity as to which
    # token is a year/month/day.  This holds regardless of what order the
    # terms are in and for each of the separators below.
    try:
        datetime.now().strftime('%-d')
        PLATFORM_HAS_DASH_D = True
    except ValueError:
        PLATFORM_HAS_DASH_D = False

    seps = ['-', ' ', '/', '.']

    year_tokens = ['%04Y']
    month_tokens = ['%b', '%B']
    day_tokens = ['%d']

    if PLATFORM_HAS_DASH_D:
        day_tokens.append('%-d')

    prods = itertools.product(year_tokens, month_tokens, day_tokens)
    perms = [y for x in prods for y in itertools.permutations(x)]
    unambig_fmts = [sep.join(perm) for sep in seps for perm in perms]

    return unambig_fmts


@pytest.mark.parser
@pytest.mark.parametrize('fmt', _get_ybd_formats())
@given(dt=dates())
def test_ybd(dt, fmt):
    dt = datetime.combine(dt, time(0))  # Make this a datetime
    dstr = dt.strftime(fmt)
    res = parse(dstr)
    assert dt == res
```

In the end we may also want to change `_get_ybd_formats` over to a strategy, because I think repeating dozens of hypothesis tests separately is expensive.

jbrockmendel:
Replaces #698.

At the moment `_ymd.resolve_ymd` can ignore `ystridx` and `dstridx` in some corner cases. This PR has `resolve_ymd` check if enough of `[ystridx, mstridx, dstridx]` have been specified to uniquely pin down (year, month, day), and if so, use those instead of other heuristics.

Edits resolve_ymd  to prevent deleting an element in-place, which makes step-through debugging more difficult.

Closes #687

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/687.bugfix.rst
+++ b/changelog.d/687.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed incorrect parsing of certain dates earlier than 100 AD when repesented in the form "%B.%Y.%d", e.g. "December.0031.30". (gh issue #687, pr #700)
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -458,14 +458,11 @@ class _ymd(list):
                 raise ValueError('Year is already set')
             self.ystridx = len(self) - 1
 
-    def _resolve_from_stridxs(self):
+    def _resolve_from_stridxs(self, strids):
         """
         Try to resolve the identities of year/month/day elements using
         ystridx, mstridx, and dstridx, if enough of these are specified.
         """
-        strids = {'y': self.ystridx, 'm': self.mstridx, 'd': self.dstridx}
-        strids = {key: strids[key] for key in strids if strids[key] is not None}
-
         if len(self) == 3 and len(strids) == 2:
             # we can back out the remaining stridx value
             missing = [x for x in range(3) if x not in strids.values()]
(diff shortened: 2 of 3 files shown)


pganssle:
This more or less looks good to me, though it does seem like a concerning amount of overhead to fix this *very* specific corner case. Not a big deal though.

jbrockmendel:
> though it does seem like a concerning amount of overhead to fix this very specific corner case

Yah, the more specific fix would have been at L498 to check `self.ystridx` in addition to `self.mstridx`, but I didn't want to get to a situation where we're checking for `ystridx` in only one of the many if/else branches, not checking `dstridx` at all, etc.

The change from 501-511 could be considered orthogonal.

jbrockmendel:
Looks fine.  Is the perf really affected?  If so then constructing `strids` once and passing it as an arg (instead of constructing it twice like in the previous version) makes sense.

Aesthetically I like the `{"y": self.stridx, ...}` better than the `(("y", self.ystridx), ...)` but that's not a big deal.

pganssle:
I doubt it will really make a major difference in the long run given that the parser is not amazingly performant to start with, but I'd like to minimize the performance impact since this particular bug is for such an incredibly niche problem.

From micro-optimizations I would estimate I saved a few microseconds by organizing it this way.

pganssle:
I doubt it will really make a major difference in the long run given that the parser is not amazingly performant to start with, but I'd like to minimize the performance impact since this particular bug is for such an incredibly niche problem.

From micro-benchmarks I would estimate I saved a few microseconds by organizing it this way.

PR #704, opened 2018-04-24
gettz documentation

### LINKED ISSUE #647: No documentation for gettz
Currently there is no documentation for `gettz`, which is the primary interface through which one gets Olson database zones.

Some prior art that should be cobbled together to make real documentation:

- [Here are some slides about gettz](https://pganssle.github.io/pygotham-2016-dateutil-talk/#/4/3) - [video here](https://youtu.be/X_bCto95Imc?t=20m25s) (audio is messed up)
- [Here is the actual implementation](https://github.com/dateutil/dateutil/blob/master/dateutil/tz/tz.py#L1431)

Unfortunately not a whole lot to go on.

weatherpattern:
## Summary of changes
Added parameters, return and examples to gettz.

Closes #647 

### Pull Request Checklist

### DIFF
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -1458,49 +1458,74 @@ def __get_gettz():
         tzlocal_classes += (tzwinlocal,)
 
     class GettzFunc(object):
-        """gettz(name=None)
+        """
+        Retrieve a time zone object from a string representation
 
-        Given a variety of inputs, like time zone name and ``TZ`` varible,
-        return the tz (aka Olson) database time zone.
+        This function is intended to retrieve the :py:class:`tzinfo` subclass
+        that best represents the time zone that would be used if a POSIX
+        `TZ variable`_ were set to the same value.
 
-        :param name:
-            A time zone name, a ``TZ`` variable, or ``None``.
+        If no argument or an empty string is passed to ``gettz``, local time
+        is returned:
 
-        :return:
-            A :class:`dateutil.tz.tzfile`. If name is ``None``,
-            the ``tzfile`` localtime is returned. 
-
-        **Examples:**
-
-        You can get the local time.
-
-        .. code-block:: python
+        .. code-block:: python3
 
             >>> gettz()
             tzfile('/etc/localtime')
 
-        You can parse time zone names.
+        This function is also the preferred way to map IANA tz database keys
+        to :class:`tzfile` objects:
 
-        .. code-block:: python
+        .. code-block:: python3
 
             >>> gettz('Pacific/Kiritimati')
             tzfile('/usr/share/zoneinfo/Pacific/Kiritimati')
 
-        You can parse a ``TZ`` variable.
+        On Windows, the standard is extended to include the Windows-specific
+        zone names provided by the operating system:
 
-        .. code-block:: python
+        .. code-block:: python3
+
+            >>> gettz('Egypt Standard Time')
+            tzwin('Egypt Standard Time')
+
+        Passing a GNU ``TZ`` style string time zone specification returns a
+        :class:`tzstr` object:
+
+        .. code-block:: python3
 
             >>> gettz('AEST-10AEDT-11,M10.1.0/2,M4.1.0/3')
             tzstr('AEST-10AEDT-11,M10.1.0/2,M4.1.0/3')
-        
-        Calling with the same name will return the same object.
 
-        .. code-block:: python
+        :param name:
+            A time zone name (IANA, or, on Windows, Windows keys), location of
+            a ``tzfile(5)`` zoneinfo file or ``TZ`` variable style time zone
+            specifier. An empty string, no argument or ``None`` is interpreted
+            as local time.
 
-            >>> tz.gettz('Pacific/Kiritimati') is tz.gettz('Pacific/Kiritimati')
-            True
+        :return:
+            Returns an instance of one of ``dateutil``'s :py:class:`tzinfo`
+            subclasses.
+
+        .. versionchanged:: 2.7.0
+
+            After version 2.7.0, any two calls to ``gettz`` using the same
+            input strings will return the same object:
+
+            .. code-block:: python3
+
+                >>> tz.gettz('America/Chicago') is tz.gettz('America/Chicago')
+                True
+
+            In addition to improving performance, this ensures that
+            `"same zone" semantics`_ are used for datetimes in the same zone.
 
 
+        .. _`TZ variable`:
+            https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html
+
+        .. _`"same zone" semantics`:
+            https://blog.ganssle.io/articles/2018/02/aware-datetime-arithmetic.html
         """
         def __init__(self):
 
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -190,8 +190,6 @@ class tzoffset(datetime.tzinfo):
 class tzlocal(_tzinfo):
     """
     A :class:`tzinfo` subclass built around the ``time`` timezone functions.
-
-
     """
     def __init__(self):
         super(tzlocal, self).__init__()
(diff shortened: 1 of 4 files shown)


dateutil/tz/tz.py
```diff
@@ -1404,6 +1432,43 @@ def __get_gettz():
         tzlocal_classes += (tzwinlocal,)
 
     class GettzFunc(object):
+        """
+        gettz is designed not to be a public function. It is defined in a closure as a private function and called once.
```
pganssle: This is not true. `GettzFunc` is not a public function. `gettz` is very much a public function, otherwise it wouldn't be in `__all__` and wouldn't be getting any documentation.

pganssle:
You seem to have checked in a `.DS_STORE` file.  You should add that to your global gitignore patterns.

I think this also needs to be rebased against master.

weatherpattern:
Thanks for the quick responses. I'll do another pass as soon as I'm able, but it might take a couple of days.

dateutil/tz/tz.py
weatherpattern: Ah sorry about that. I misunderstood your other comment in the original issue.

pganssle:
Something seems to have gone wrong with the rebase because this PR now includes a bunch of commits unrelated to it. I'll go in there and clean up the history.

weatherpattern:
Yeah, sorry about that, it was a bit of a mess trying to get all the merge conflicts sorted out. Thanks in advance for cleaning up the history.

pganssle:
OK, I've significantly cleaned up the history, including squashing all the fixup stuff. You'll need to do `git pull --force` before making any additional changes or pushing to your branch.

Also, I noticed that your `user.email` is set to `raymondcha@MacBook-Air-2.local`. If that is not the e-mail you use on Github (and I assume it is not), Github won't know to associate those commits with you.

If you don't want your e-mail address in the git history, you can set it to `weatherpattern@users.noreply.github.com` and your commits will be associated with your Github username, if you care about that.

I can rewrite the git history to use that e-mail address if you'd prefer.

dateutil/tz/tz.py
```diff
@@ -1460,6 +1458,43 @@ def __get_gettz():
...
+
+            >>> gettz('Pacific/Kiritimati')
+            tzfile('/usr/share/zoneinfo/Pacific/Kiritimati')
+
+        You can parse a TZ variable.
+
+        ..code-block::
+
+            >>> from dateutil.tz import tzutc, UTC
+            >>> tgettz('AEST-10AEDT-11,M10.1.0/2,M4.1.0/3')
```
pganssle: The UTC and tzutc imports are unnecessary, and `tgettz` is definitely not right.

weatherpattern:
Hi for the clean up. I did a `git pull --force` and my commit matches upstream, which is a good sign.
Also I changed my global email address, but when you have some time,  if you could rewrite the git history to use that no-reply email that would be great. Thanks.

pganssle:
OK done.

dateutil/tz/tz.py
```diff
@@ -1459,7 +1459,8 @@ def __get_gettz():
 
     class GettzFunc(object):
         """
-        gettz takes a variety of inputs and returns a time zone.
+        Given a variety of inputs, like time zone name and ``TZ`` varible,
+        return a time zone.
 
         :param object:
```
pganssle: This is not actually the interface. `GettzFunc` is a class representing a callable function. `object` is the name of the base type of all objects in Python 3 ("new style classes"), and so `GettzFunc` is a subclass of `object` for backwards compatibility with Python 2.

I realize it's confusing, but what's being written here is *actually* the documentation for `tz.gettz`, which is an instance of `GettzFunc`, but for practical purposes is equivalent to a `function`.

When you call `tz.gettz()`, it actually invokes [`GettzFunc.__call__`](https://github.com/weatherpattern/dateutil/blob/503b5e9331a4d800bb5875d02612b097c57d12de/dateutil/tz/tz.py#L1501), which itself is a thin wrapper around [`GettzFunc.nocache`](https://github.com/weatherpattern/dateutil/blob/503b5e9331a4d800bb5875d02612b097c57d12de/dateutil/tz/tz.py#L1520).

The reason it's built this way is partially for performance reasons, but mostly because I want `tz.gettz` to satisfy the following invariant:

```python
assert tz.gettz(name) is tz.gettz(name)
```

So I need to store a reference to the arguments that have been passed to `tz.gettz` and all the results, so that I can return the same object if I get the same arguments. Those get stored on the `GettzFunc` instance in the `_cache`. Because you might want the old semantics, I've exposed a "nocache" [variant](https://github.com/python-variants/variants) that bypasses the cache: `tz.gettz.nocache(name)`, and since cache misses are identical to calls to `nocache`, `__call__` is implemented in terms of `nocache`.

There are also functions on `gettz` for manipulating the cache, like `clear_cache`.

weatherpattern: Thanks for the extended comments. I'm going to a little more time digest and a bit more research and make another pass. More coming.

dateutil/tz/tz.py
```diff
@@ -1460,6 +1458,50 @@ def __get_gettz():
...
+        """gettz(name=None)
+
+        Given a variety of inputs, like time zone name and ``TZ`` varible,
+        return the tz (aka Olson) database time zone.
+
+        :param name:
+            A time zone name, a ``TZ`` variable, or ``None``.
+
+        :return:
+            A :class:`dateutil.tz.tzfile`. If name is ``None``,
```
pganssle: This is also not accurate. If you look at the code, it can return `tzfile`, `tzlocal`, `tzstr`, possibly others.

This function is that it takes a string representing a time zone and returns the best-matching `tzinfo` subclass instance. The specific subclass returned is not part of the function's contract.

dateutil/tz/tz.py
```diff
@@ -1460,6 +1458,50 @@ def __get_gettz():
         tzlocal_classes += (tzwinlocal,)
 
     class GettzFunc(object):
+        """gettz(name=None)
```
pganssle: The function prototype will be auto-generated by sphinx and should not be included.

dateutil/tz/tz.py
weatherpattern: OK I took another pass and hopefully more accurate now. I also added the assertion as an example.

A couple questions:
Your explanation above was really helpful. Do you want me to make a new issue to put that information into the individual methods and functions as comments?

Also, I wanted to double check:
```
>>tz.gettz() is tz.gettz()
False
```
This is as designed, right? Because if ```name=None```, then the time zone is not cached.

pganssle: > Do you have an example (could be anywhere) of how much detail you prefer in the docstring?

I'm not really satisfied with the `dateutil` documentation in general. I personally originated the functions (and hence the docstrings) for [isoparse](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse) and [default_tzinfo](https://dateutil.readthedocs.io/en/stable/utils.html#dateutil.utils.default_tzinfo), so something like that is likely what I would write, but I actually need to do a full re-organization of the documentation, and I'm not sure where information should go for each function / class.

I'm thinking in the end that however it's organized, *somewhere* in the documentation, the functions and classes would be documented roughly with a `man` page level of detail.

> Your explanation above was really helpful. Do you want me to make a new issue to put that information into the individual methods and functions as comments?

Sure.

> Also, I wanted to double check:

```python
>>> tz.gettz() is tz.gettz
False
```
> This is as designed (I think.) Because if name=None, then the time zone is not cached. Is that correct?

Yes and no. Ideally that would be `True` even for `None`, but since `tz.gettz()` represents local time it's made a bit more complicated by the fact that local time might change over the runtime of a program (which leads to a bunch of bugs, but we don't want to add to the bugginess by adding to the problem). As such, I'd rather err on the side of the previous non-cached behavior until we can work out how best to figure out if the "local zone" cache has been invalidated by changes in the execution environment.

weatherpattern:
I agree that it would be good pause and regroup on this. Thanks your time and all your responses.

pganssle:
@weatherpattern Thanks for your hard work on this!

Issue #521, opened 2017-11-13
Get coverage to 100%


pganssle:
This is an issue for tracking progress towards 100% test coverage. I believe that, as measured, #520 will be a prerequisite for this.

A lot of the uncovered lines according to [Codecpv](https://codecov.io/github/dateutil/dateutil) that *aren't* related to the "no builds on Windows" stuff seem to be error conditions and `repr` values for various private and/or public classes.

For private class `repr`s, we can "smoke test" them in `test_internals.py`, which is where we put tests that *aren't* intended to document the public interface, but just to exercise the various lines that otherwise would go under a `nocover` pragma.

PR #712, opened 2018-05-08
Added tests DST only timezone file

Bergvca:
## Summary of changes

Added two unittests that check that in case a timezone file is loaded where all ttinfo entries have isdst set to 1: 
- The tzfile's _ttinfo_std becomes the same as it's _ttinfo_dst
- The tzfile's _ttinfo_before becomes the same as it's first dst

Closes 
N/a - related to *Get coverage to 100%* (#521)
### Pull Request Checklist
- [x] Changes have tests


pganssle:
Tests should not test the private interface. If you want to test this, make use of the public methods only. Anything with a leading underscore is private and as such is an implementation detail.

The proper tests for this would presumably involve checking the `dst()` values for this invented zone.

Bergvca:
Changed the test to only hit public interface. These test are testing lines 643 and 650 of tz.py and adding it to the code coverage.

dateutil/test/test_tz.py
```diff
@@ -1947,6 +1954,30 @@ def testIsStd(self):
...
+    def testBeforeTtinfoFirstDst(self):
+        # If fileobj has only DST ttinfo entries, its "before" ttinfo entry
+        # (the ttinfo before all transitions) will be its first DST entry.
+        # tzfile.dst() will return the dst offset of this first DST entry.
+        # TZFILE_DSTONLY is an artificial timezone file where each dst byte is
+        # set to '1'
+        fileobj = BytesIO(base64.b64decode(TZFILE_DSTONLY))
+        tzc = tz.tzfile(fileobj)
+        dt = datetime(2002, 1, 1, 10, 0, 0)
+        tzc._ttinfo_list[0].dstoffset = timedelta(15)
```
pganssle: This is not part of the public interface. How does setting this value help test this?

Bergvca: I've added that line for readability - as the input is a binary file it is not clear for the reader what the dstoffset of the first ttinfo entry is. 

Another option is to use the dstoffset in the file and mention that value with a comment. The reader will have to "trust" this comment is correct though. 

Something else: do you have an idea why the Travis CI build failed? It worked the first time I pushed this branch...

pganssle: It appears the test actually fails if this line is removed, so it's apparently *not* just making it clear but actually playing a vital role in the test.

pganssle: Yes it's best to just construct a file object that has the properties you want and then test the public interface of those properties.

Bergvca: Fixed it to test for an dst offset of 0. A ttinfo dst offset is dependent on its last std offset. This file doesn't have any std offsets, so it can only be 0. 

Btw, do you think the ```dst``` and ```utcoffset``` need documentation for ```tzfile```? I can add it if you want.

pganssle: Hm... This is an interesting situation, because I *think* it could be actually *wrong*, in the sense that these are in fact DST offsets, but their `dst()` will return `ZERO` and thus be falsy.  But because of the way the file format works, we won't be able to figure out what the actual DST offset should be, because we have no examples of a standard offset for contrast.

pganssle: Well, AFAIK there's nothing in the `tzfile(5)` spec that says you couldn't have an all-DST time zone. The problem is that the Python time zone API made the somewhat strange choice of trying to expose the specific offset that is due to DST. This information is not contained in the `tzfile` and has to be inferred from the difference in offset between adjacent DST and STD offsets.

The problem here is that there is no "right" behavior because we're looking at a made-up zone (so we can't try to fall back on the *actual* data), and the concept of having a separate DST offset (as opposed to just marking that an offset is in DST) is not present in the original files. I am tempted to say that the correct behavior (the one we want to assert on) would be that `dst()` should be truthy whenever `isdst` is 1, but that would be a pretty weird code change. I'll take a look at this and see if I can come up with an assertion I'm happy with.

Thanks for this PR, by the way, it sheds light on a pretty weird edge case.

PR #725, opened 2018-05-22
Parametrize pytest

jpurviance:
## Summary of changes

This is a possible partial conversion.
I've tried to keep the grouping and meaning of test cases the same. 

For the test cases in `test_parser.py` I've converted the old test function signatures to assertion messages. Where appropriate, I've integrated the comments from the test cases into the error messages as well. I've noticed that a large number of the test cases were` testSomeThing1, testSomeThing2, testSomeThing3,...` where the trailing number has appeard to be indexing the `i'th` test case I have dropped it. 

For test cases in `test_easter.py` I've dropped the `for easter_date in` to loop over cases. The problem with that is once an assertion fails, none of the following tests in the loop will be ran. Now, even if the first test case for each of the easter tests cases fails the following will still be ran. Also, since the refactor was for all of the test cases in the file I applied a `PEP8` format to the file. 

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/dateutil/test/test_easter.py
+++ b/dateutil/test/test_easter.py
@@ -2,94 +2,92 @@ from dateutil.easter import easter
 from dateutil.easter import EASTER_WESTERN, EASTER_ORTHODOX, EASTER_JULIAN
 
 from datetime import date
-import unittest
+import pytest
 
 # List of easters between 1990 and 2050
 western_easter_dates = [
     date(1990, 4, 15), date(1991, 3, 31), date(1992, 4, 19), date(1993, 4, 11),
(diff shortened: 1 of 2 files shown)


pganssle:
@jpurviance Can you split this up into separate PRs? The Easter stuff looks good so far, though I'm not really happy with the changes to the data formatting.

I'd like to review the parser stuff separately, though, because that requires a much more careful eye to make sure we're not changing anything (due to the implicit filling-in of missing elements).

pganssle:
Also the split PRs need to be rebased against master to get the Python 3.3 tests fixed.

PR #726, opened 2018-05-23
conversion of test_easter.py to use pytest

jpurviance:
## Summary of changes

I've dropped the for easter_date in to loop over cases. The problem with that is once an assertion fails, none of the following tests in the loop will be ran. Now, even if the first test case for each of the easter tests cases fails the following will still be ran. Also, since the refactor was for all of the test cases in the file I applied a PEP8 format to the file.

Closes 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -45,6 +45,7 @@ switch, and thus all their contributions are dual-licensed.
 - Jake Chorley (gh: @jakec-github) **D**
 - Jan Studený <jendas1@MASKED>
 - Jitesh <jitesh@MASKED>
+- John Purviance <jpurviance@MASKED> (gh @jpurviance) **D**
 - Jon Dufresne <jon.dufresne@MASKED> (gh: @jdufresne) **R**
 - Jonas Neubert <jonas@MASKED>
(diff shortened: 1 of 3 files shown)


pganssle:
Needs a changelog and you can add yourself to the changes file, otherwise I think this one is ready to go.

jpurviance:
for the first pr, why was the pep8 for the lists not desired?

pganssle:
@jpurviance I am not sure I 100% agree with PEP8 on everything, but in particular when large arrays of "data literals" like this, I find it *much* easier to read if it is neatly aligned instead of having these sort of ragged edges.

This is different from application code where you aren't likely to have some huge data array just sitting in the middle of your code.

At some point I'll set up automatic linting for the repo and we'll have to set out the rules in a way that linters can understand, but I don't think that starts with making `test_easter` uglier and harder to read.

jpurviance:
Ok, I understand. Sounds reasoanble. What other style guildlines does this project follow that pep8 disagrees with?

jpurviance:
added changelog entry and updated authors.

PR #727, opened 2018-05-23
parametrize part of parser test cases with pytest

jpurviance:
## Summary of changes

This is a possible partial conversion.
I've tried to keep the grouping and meaning of test cases the same.

 I've converted the old test function signatures to assertion messages. Where appropriate, I've integrated the comments from the test cases into the error messages as well. I've noticed that a large number of the test cases weretestSomeThing1, testSomeThing2, testSomeThing3,... where the trailing number has appeared to be indexing the i'th test case I have dropped it.

Closes 

### Pull Request Checklist

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -27,6 +27,130 @@ try:
 except ValueError:
     PLATFORM_HAS_DASH_D = False
 
+# Parser test cases using no keyword arguments. Format: (parsable_text, expected_datetime, assertion_message)
+PARSER_TEST_CASES = [
+    ("Thu Sep 25 10:36:28 2003", datetime(2003, 9, 25, 10, 36, 28), "date command format strip"),
+    ("Thu Sep 25 2003", datetime(2003, 9, 25), "date command format strip"),
+    ("2003-09-25T10:49:41", datetime(2003, 9, 25, 10, 49, 41), "iso format strip"),
+    ("2003-09-25T10:49", datetime(2003, 9, 25, 10, 49), "iso format strip"),
+    ("2003-09-25T10", datetime(2003, 9, 25, 10), "iso format strip"),
+    ("2003-09-25", datetime(2003, 9, 25), "iso format strip"),
+    ("20030925T104941", datetime(2003, 9, 25, 10, 49, 41), "iso stripped format strip"),
+    ("20030925T1049", datetime(2003, 9, 25, 10, 49, 0), "iso stripped format strip"),
+    ("20030925T10", datetime(2003, 9, 25, 10), "iso stripped format strip"),
+    ("20030925", datetime(2003, 9, 25), "iso stripped format strip"),
+    ("2003-09-25 10:49:41,502", datetime(2003, 9, 25, 10, 49, 41, 502000), "python logger format"),
+    ("199709020908", datetime(1997, 9, 2, 9, 8), "no separator"),
+    ("19970902090807", datetime(1997, 9, 2, 9, 8, 7), "no separator"),
+    ("2003-09-25", datetime(2003, 9, 25), "date with dash"),
+    ("09-25-2003", datetime(2003, 9, 25), "date with dash"),
+    ("25-09-2003", datetime(2003, 9, 25), "date with dash"),
+    ("10-09-2003", datetime(2003, 10, 9), "date with dash"),
+    ("10-09-03", datetime(2003, 10, 9), "date with dash"),
+    ("2003.09.25", datetime(2003, 9, 25), "date with dot"),
+    ("09.25.2003", datetime(2003, 9, 25), "date with dot"),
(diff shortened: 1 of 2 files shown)


dateutil/test/test_parser.py
```diff
@@ -27,6 +27,132 @@
 except ValueError:
     PLATFORM_HAS_DASH_D = False
 
+parser_test_cases = [
```
pganssle: For these sorts of globals I have tended to use ALL_CAPS

jpurviance: ok, i'm not sure how i fee about the variable name anyway. All of these tests are "parser test cases"

dateutil/test/test_parser.py
```diff
@@ -27,6 +27,132 @@
...
+    ("01h02", datetime(2003, 9, 25, 1, 2), "random format"),
+    ("01h02s", datetime(2003, 9, 25, 1, 0, 2), "random format"),
+    ("01m02", datetime(2003, 9, 25, 0, 1, 2), "random format"),
+    ("01m02h", datetime(2003, 9, 25, 2, 1), "random format"),
+    ("2004 10 Apr 11h30m", datetime(2004, 4, 10, 11, 30), "random format")
+]
+
+
+@pytest.fixture
+def default():
```
pganssle: What is the advantage of using a fixture for this? I will admit I don't love fixtures in general because they feel too magical (being being injected by name into the arguments is... very weird).

Since all it does is provide a `datetime` literal, maybe just pass that to `default` instead.

dateutil/test/test_parser.py
```diff
@@ -536,9 +536,6 @@ def testMicrosecondPrecisionErrorReturns(self):
             dt = datetime(2008, 2, 27, 21, 26, 1, ms)
             self.assertEqual(parse(dt.isoformat()), dt)
 
-    # def testHighPrecisionSeconds(self):
```
pganssle: Hm... Not sure why this is commented out. If it fails it should be converted to an xfail.

dateutil/test/test_parser.py
pganssle: They don't have to go into a variable at all if you don't want.

Most of the time I just pass them directly into the `parametrize` decorator, like so:

```python
@pytest.mark.parametrize('arg1, arg2', [
    (1, 2),
    (3, 4),
    (5, 6),
])
def test_func(arg1, arg2):
    assert func(arg1, arg2)

dateutil/test/test_parser.py
jpurviance: pytest likes dependency injection. `test_parser_default` is not testing a specific default value, it is instead testing that given some text and a default that you create an expected datetime object. The test is currently written to assert that given some text an a default a particular result is created. The alternative would given some text with an exact default a particular result is created.

dateutil/test/test_parser.py
jpurviance: It should be deleted in the last commit. I missed it when I removed all the commented out tests. its here as a specific commit because I cherry picked two of the commits from the previously closed pr. see: https://github.com/dateutil/dateutil/pull/725/commits/42f62c2e659f617cc8153d38ecc8710cb9258bbd

dateutil/test/test_parser.py
pganssle: But if we ever change `default` from returning this specific `datetime` literal, all the parametrized tests would immediately break, right? So really the test is whether that specific result is created *given this specific default value*. What specific things could we do with this fixture that a datetime literal don't accomplish?

dateutil/test/test_parser.py
pganssle: Oh you commented them out as you moved them over, I see.

dateutil/test/test_parser.py
jpurviance: I saw this style of parameterization in some of the other test files. I chose against it because a lot of the important code, like: `@pytest.mark.parametrize('arg1, arg2', [`, is not close to the function being decorated. 

I didn't like that it did not fit on one screen for both how the args are being used and what the args are. With them separated, you have to the data to be injected separate form how it is injected. I liked this because I felt the "how" should live near the function being decorated.

pganssle: Yeah, it's a reasonable choice. I would probably go with ALL_CAPS if you are going with global scope, though. See `dateutil/tests/test_isoparser.py` for how I did it.

dateutil/test/test_parser.py
jpurviance: yeah, very scientific.

dateutil/test/test_parser.py
jpurviance: For this test case and your example, not much. 

Things like this would be possible: https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures

However, since that datetime object needs to be in a particular state, I don't see parametrizing fixtures useful.

PR #728, opened 2018-05-27
Separate out _parse_token; small whitespace (flake8) cleanups

jbrockmendel:
Separates _parse_token out from the big while loop in _parse.  The main motivation here is that when debugging we generally want to step through the parse process token-by-token, which is difficult in the status quo.

Zero logic is affected.

Small miscellaneous flake8 whitespace fixups.

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -663,7 +663,7 @@ class parser(object):
     class _result(_resultbase):
         __slots__ = ["year", "month", "day", "weekday",
                      "hour", "minute", "second", "microsecond",
-                     "tzname", "tzoffset", "ampm","any_unused_tokens"]
+                     "tzname", "tzoffset", "ampm", "any_unused_tokens"]
 
     def _parse(self, timestr, dayfirst=None, yearfirst=None, fuzzy=False,
                fuzzy_with_tokens=False):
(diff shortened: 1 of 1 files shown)


pganssle:
I'm kinda surprised that there were no `break` or `continue` statements in that entire `while` loop.

That said, I'm not sure I understand how this helps anything. `_parse_token` still parses multiple tokens on each pass, so how does this help you advance token-by-token? Seems like you could just set a breakpoint at the beginning of the `while` loop and it would do the same thing. I don't think it really makes things much cleaner since the contents of the loop were already basically the only thing in the function, and now you incur function call overhead on every iteration of the loop.

jbrockmendel:
> I'm kinda surprised that there were no break or continue statements in that entire while loop.

There were a bunch of `continue` statements until we cleaned them out in #419.

> That said, I'm not sure I understand how this helps anything. [...]

I thought about most of these.  Eventually I concluded that the function-call-overhead is the only one that really bothers me (and in fairness, given that it is the only one that _actually_ changes anything in this PR, it's not nothing).

> Seems like you could just set a breakpoint at the beginning of the while loop and it would do the same thing

In practice this is a hassle.

> `_parse_token` still parses multiple tokens on each pass, so how does this help you advance token-by-token

This is a an important long-term issue.  At the moment the parser is forward-looking, so from token`i` it will, if necessary, look at tokens `i+1`, `i+2`, ...  To change the multiple-tokens-on-each-pass behavior we would need to make it backward-looking and track more state information (e.g. to know that token `i-2` was assigned to `res.hour`, see that token `i-1` is `:`, so infer that token `i` is likely a minute).

This more-state-information is likely to be necessary/helpful for two other goals that have been discussed:
a) finding a format string so future calls can be sped up with `strptime`
b) some of the difficult test [cases](https://github.com/dateutil/dateutil/blob/master/dateutil/test/test_parser.py#L996)

Now that I think about it, you're right and this PR is putting the Cart Before The Horse in moving towards those goals.  The state-tracking discussion should come first.  Closing.

PR #731, opened 2018-05-29
restrict timedelta for tz.tzrange

### LINKED ISSUE #594: tzrange offsets can be outside of -24:00/24:00 range
Noticed this when reviewing #581:

```python
from dateutil import tz
from datetime import *

tz1 = tz.tzstr('WGT30')
tz2 = tz.tzrange('WGT', timedelta(hours=-30))

dt1 = datetime(2011, 1, 1, tzinfo=tz1)
dt2 = datetime(2011, 1, 1, tzinfo=tz2)

for dt in [dt1, dt2]:
    try:
        dt.utcoffset()
    except ValueError as e:
        print(e)

# Result:
# offset must be a timedelta strictly between -timedelta(hours=24) and timedelta(hours=24).
# offset must be a timedelta strictly between -timedelta(hours=24) and timedelta(hours=24).
```

Simple matter of checking the range in the constructor. Not sure if `tzwin` is also affected.

jpurviance:
## Summary of changes

fixes #594

`tz.tzrange("...", .. )` is not checking bounds for offsets. Values that are outside [-24,24] are invalid tzoffsets. With this fix, a value error is thrown when invalid values are provided. 

Closes 

### Pull Request Checklist
- [ ] Changes have tests


dateutil/tz/tz.py
```diff
@@ -941,6 +941,13 @@ def __init__(self, stdabbr, stdoffset=None,
         self._std_abbr = stdabbr
         self._dst_abbr = dstabbr
 
+        if isinstance(stdoffset, datetime.timedelta) and \
+                not (datetime.timedelta(hours=-24) <= stdoffset <= datetime.timedelta(hours=24)):
+            tzrange_error_message = "Invalid timezone offset, " + \
+                                    "offset must be a timedelta strictly between: " +\
+                                    "-timedelta(hours=24) and timedelta(hours=24)"
+            six.raise_from(ValueError(tzrange_error_message), None)
```
pganssle: Why are you using `six.raise_from` in this context?

dateutil/tz/tz.py
```diff
@@ -941,6 +941,13 @@ def __init__(self, stdabbr, stdoffset=None,
         self._std_abbr = stdabbr
         self._dst_abbr = dstabbr
 
+        if isinstance(stdoffset, datetime.timedelta) and \
```
pganssle: Won't this still allow things like `tz.tzrange(None, 90000)`?

dateutil/test/test_tz.py
```diff
@@ -2650,3 +2650,19 @@ def test_resolve_imaginary_monrovia():
...
+        tz.tzrange("", delta)
+
+
+@pytest.mark.parametrize("delta", [timedelta(hours=0), timedelta(hours=-24), timedelta(hours=-6),
+                                   timedelta(), timedelta(hours=6), timedelta(hours=24)])
+def test_valid_tzrange_offset(delta):
+    try:
+        tz.tzrange("", delta)
+    except ValueError:
+        pytest.fail()
```
pganssle: Is there a reason to fail this way rather than just allowing the `ValueError` to *be* the failure condition?

pganssle:
A few comments about this:

1. I'm doing a thing now where as I go I start adding `pytest.mark` marks where previously I would have grouped everything under the same `TestCase`, so for the `tzrange` tests, please add `@pytest.mark.tzrange`
2. I'm torn about the "valid `tzrange`" test. On the one hand, it's nice that there's an obvious dedicated test that these edge cases are valid. On the other hand, I think that these tests can actually just be subsumed into the normal test cases for `tzrange` that actually assert something (e.g. they can be used).
3. The tests (and possibly the implementation) should cover both the `timedelta` way of creating `tzrange` and the "integer seconds" way fo creating `tzrange`.
4. The tests (and possibly the implementation) should check that this is the case for *both* the STD and DST offsets, not just STD.

dateutil/tz/tz.py
jpurviance: when looking at the doc string, i see the spec says:
```python
"""
...
:param stdoffset:
        An integer or :class:`datetime.timedelta` object or equivalent
        specifying the base offset from UTC.
...
"""
```

What would be an equivalent? Also, while its not explicity said, im assuming that if its an int that is provided, its the number of seconds as the offset?

jpurviance:
I have a thought about `2.`, I'm in favor of having some explicit test cases for the valid range because it pushes the documentation into the codebase. I like this because as a user, if I'm feeling curious, I can see that there are test cases that help clarify the valid input domain. While this can still be achieved with the information being encoded into more broad TZ test cases, I feel that it is easier to pick up on these if they live in their own test cases. That being said, I do think that this information should also sit in the normal tztest cases (in some manner, it might not make sense to move all the test cases that are needed to test the entire function domain into broad test cases)

dateutil/tz/tz.py
pganssle: Yes, `int` would be seconds. An equivalent would be anything that implements `timedelta`'s interface, I guess.

PR #732, opened 2018-05-29
simplify+optimize timelex

jbrockmendel:
If we're not going to make timelex subclassable, then we might as well make it performant.  This replaces an in-python loop-over-every-character state machine (with lots of unnecessary function calls) with a single regexp (admittedly not an entirely trivial regexp, but the "4.5.6" case takes a little effort).

Also in some cases removes an unnecessary StringIO instantiation.

Speedup ~22%

Before:
```
In [8]: %timeit dateutil.parser.parse('2018-05-28 08:18 PM')
1000 loops, best of 3: 229 µs per loop
```

After:
```
In [3]: %timeit dateutil.parser.parse('2018-05-28 08:18 PM')
10000 loops, best of 3: 178 µs per loop
```

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -56,8 +56,6 @@ __all__ = ["parse", "parserinfo"]
 # making public and/or figuring out if there is something we can
 # take off their plate.
 class _timelex(object):
-    # Fractional seconds are sometimes split by a comma
-    _split_decimal = re.compile("([.,])")
 
     def __init__(self, instream):
         if six.PY2:
@@ -70,17 +68,14 @@ class _timelex(object):
                 instream = instream.decode()
 
         if isinstance(instream, text_type):
-            instream = StringIO(instream)
+            self.stream = instream
         elif getattr(instream, 'read', None) is None:
             raise TypeError('Parser must be a string or character stream, not '
                             '{itype}'.format(itype=instream.__class__.__name__))
+        else:
+            self.stream = instream.read()
 
-        self.instream = instream
-        self.charstack = []
-        self.tokenstack = []
-        self.eof = False
-
-    def get_token(self):
+    def get_tokens(self):
         """
         This function breaks the time string into lexical units (tokens), which
         can be parsed by the parser. Lexical units are demarcated by changes in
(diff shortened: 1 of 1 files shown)


pganssle:
I'm not sure what gave you the impression that I'm not intending for `_timelex` to be made subclassable. I have an open issue specifically to address this and in #729 I said many times that the plan is for `timelex` to be made public.

That said, this is a pretty tempting speed increase. It might be worth merging for now even if it has to be reverted later. I don't love that it involves two passes, or that it goes from lazy to eager, but it's not like the parser is actually *using* the fact that it's lazy.

dateutil/parser/_parser.py
```diff
@@ -94,132 +89,38 @@ def get_token(self):
...
-            raise StopIteration
-
-        return token
-
-    def next(self):
-        return self.__next__()  # Python 2.x support
+        stream = self.stream.replace('\x00', '')
+
+        # TODO: Change \s --> \s+ (this doesn't match existing behavior)
+        # TODO: change the punctuation block to punc+ (doesnt match existing)
```
pganssle: Why `\s+` and `punc+`? What formats use multiple punctuation separators like that?

jbrockmendel: For the first pass at this I wanted to keep the tokenization behavior as close to the existing as possible.  In all of the test cases the tokenization is identical ATM.  There's no particular reason for the punctuation entry to be retained.

That said, the `\s` cannot be changed to `\s+` without making other edits to the parser (which I think would be a good change, but not necessarily in this PR).  This is because 'info.jump` recognizes single spaces but not multiple spaces (related: #732)

jbrockmendel:
> I don't love that it involves two passes

Yah this is my least-favorite part too.  It _could_ be accomplished with a much more complicated regex, but I don't think its worth it.  My preference would be to do a follow-up to remove this second pass and instead check for that comma behavior in the `parse` step.

jbrockmendel:
Side note: looking at one of the xfail cases `"1,700"` it seems like a couple of not-great things are happening, one of which is its getting converted to `"1.700"`.  Then the `1` gets pegged as the day and... not sure why the `".700"` gets ignored.  Will look into it.

jbrockmendel:
@pganssle thoughts on this?  The two-pass issue gets resolved by #738.

pganssle:
I'm still slowly working my way through the PRs from the PyLondinium sprints. I'm not sure about moving the comma-separated stuff from the lexer into the parser, but I think I'm still inclined to merge whatever version of this one has zero behavioral changes.

PR #735, opened 2018-06-04
use setup_class instead of setUp

jbrockmendel:
Most of the parser tests are written with the older unittest structure.  Now that we're using pytest, some of that can be cleaned up.

The only substantive change is using a classmethod `setup_class` instead of a regular method `setUp`, so it only gets called once instead of for each test.  Small-n it looks like this saves 20-30% of the runtime for that file (which is only 3-5 seconds, so really not a big deal).

There are a lot of `self.assertEqual` statements so getting them all will take a few passes.

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -152,7 +152,7 @@ def test_parser_default(parsable_text, expected_datetime, assertion_message):
     assert parse(parsable_text, default=datetime(2003, 9, 25)) == expected_datetime, assertion_message
 
 
-class TestFormat(unittest.TestCase):
+class TestFormat(object):
 
     def test_ybd(self):
         # If we have a 4-digit year, a non-numeric month (abbreviated or not),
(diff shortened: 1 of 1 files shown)


pganssle:
@jbrockmendel Sorry, I let this sit unreviewed for too long and now it's got some serious merge conflicts. Do you want to resolve them or should I?

jbrockmendel:
I'm happy to resolve them, but it'll be a few days before I can get around to it.

pganssle:
No rush.

jbrockmendel:
This should be ready.  I've identified some more test cleanups to do in another pass, will wait until this is merged.

dateutil/test/test_parser.py
```diff
@@ -152,7 +152,7 @@ def test_parser_default(parsable_text, expected_datetime, assertion_message):
     assert parse(parsable_text, default=datetime(2003, 9, 25)) == expected_datetime, assertion_message
 
 
-class TestFormat(unittest.TestCase):
+class TestFormat(object):
```
pganssle: I've been writing so much Python 3 lately that I was initially taken aback by this. Sooooooon.

pganssle:
Gah, forgot a news entry. No need to make a separate PR for it considering it's an internal refactoring, I'll just do it later.

PR #738, opened 2018-06-07
move comma-microsecond logic from timelex to parser

jbrockmendel:
This looks small, but is somewhat far-reaching.

- Treat commas as a decimal _exclusively_ in seconds.  ATM if parse `'2003 May 4, 14:10,7'` the `10,7` gets treated as "10.7 minutes", which I do not think is intended.

- By getting rid of the comma in `_split_decimal`, we can further simplify #732 to get rid of the second-pass which is the main point of ugliness there.

- If there is no decimal separator in a seconds token, return `None` as the microseconds instead of 0.  Down the road this will be useful for keeping track of resolution information.

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -57,7 +57,7 @@ __all__ = ["parse", "parserinfo"]
 # take off their plate.
 class _timelex(object):
     # Fractional seconds are sometimes split by a comma
-    _split_decimal = re.compile("([.,])")
+    _split_decimal = re.compile("([.])")
 
     def __init__(self, instream):
         if six.PY2:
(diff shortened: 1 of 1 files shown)


jbrockmendel:
This will need tests before being mergeable.  Before that we need to make a decision about the behaviors that this changes, in particular, do we allow commas as a decimal separator for minutes, or only for seconds?  If we allow them for minutes, I'll need to patch this.

dateutil/parser/_parser.py
```diff
@@ -57,7 +57,7 @@
 # take off their plate.
 class _timelex(object):
     # Fractional seconds are sometimes split by a comma
-    _split_decimal = re.compile("([.,])")
+    _split_decimal = re.compile("([.])")
```
pganssle: If we're moving the logic into the parser, we don't have to use a compiled regular expression, since I believe `token.split('.')` will work.

jbrockmendel: Good idea.

IIRC the status of this is that it needs to entail [zero behavioral changes](https://github.com/dateutil/dateutil/pull/732#issuecomment-401887004).  That is not yet the case with this PR; I'll move this up the priority list.

PR #740, opened 2018-06-07
Check README is Rendered correctly

### LINKED ISSUE #734: Add readme_renderer check to docs test
Right now we don't do any automated testing that the README will be rendered correctly on PyPI. PyPI's long description renderer is very finicky and if there are any mistakes, it will fail to render *at all* and just be a bunch of ugly markup.

I think we can test this in the test suite by adding a [`readme_renderer`](https://pypi.org/project/readme_renderer/) check to the `docs` build for `tox`. So add `readme_renderer` to the `deps` for [`testenv:docs`](https://github.com/dateutil/dateutil/blob/master/tox.ini#L53), and then add to the commands `python setup.py -r -s`.

To test that it works, deliberately break the README in some way and make sure that `tox -e docs` fails.

alimcmaster1:
## Summary of changes
Check ReadME.rst is rendered correctly on PyPI.

Closes #734

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -95,5 +95,6 @@ switch, and thus all their contributions are dual-licensed.
 - labrys <labrys@MASKED> (gh: @labrys)
 - ms-boom <ms-boom@MASKED>
 - ryanss <ryanssdev@MASKED> (gh: @ryanss)
+- Alistair McMaster <alistair@MASKED> (gh: @alimcmaster1 ) **D**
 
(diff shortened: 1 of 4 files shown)


pganssle:
@alimcmaster1 Can you add yourself to the authors list? Please add **D** afterwards to indicate that all contributions are dual-licensed.

alimcmaster1:
Sure, have now added thanks!

PR #745, opened 2018-06-08
adopt a standard, automatic and enforced code style

gaborbernat:
Not sure if we want to adopt this. However, this PR creates the tox env ``format``. Calling this ensures code style (PEP-8 and import ordering) not by warning about them, but actually fixing them. I've created the PR to see what changes would this generate and if we want this.

Uses black and  isort.


pganssle:
I am sorta OK with this idea, but I am not yet convinced that we should use `black`, and I don't like how unreadable it makes some of the data arrays in, for example, `test_easter`.

gaborbernat:
We can use ``fmt:off`` and ``fmt:on`` to turn it off in those sections. I also initially had some bad feelings about some style changes, but in the end, I realized those changes makes sense and makes diffs lot nicer.

anthrotype:
I've been using black only for a couple of weeks. From my limited experience I have the say that I am beginning to like it. At first I was also startled by some of its decisions, but gradually I've become accustomed to it and I like that I don't have to think about style anymore. Well, unless when sometimes it looks too odd, in which case instead of `fmt:off` I tend to change variable names, or put some parentheses, so the line breaks where I want.
I tried other autoformatters like yapf, but unlike black these have a huge list of options and configuration. Paradoxically, I actually like that black does _not_ allow me to change things! Perhaps it's me that I'm lazy.
And of course, there's the thing that you don't have to argue with other contributors and maintainers about formatting of PRs. You can just tell them to run black (or tox -e format, or whatever) or even set up a pre-commit hook if they wish.
Anyway, I understand that for a "mature" project like dateutil it's not easy to make the switch. Even pip considered using black but then decided to not do that, as some maintainers didn't like all of blacks style rules, or wished to have more flexibility or control over formatting.

gaborbernat:
@pganssle running pylint would already cause throwing those errors; wouldn't adding that be an orthogonal issue though?

pganssle:
@gaborbernat They are basically the same in the sense that the overall issue is "Decide on a style for this project and document and/or enforce it."

I don't want a lot of churn in formatting-related changes, so I would like to make the decisions once, so this can't be a "Well, this is better than we have so we'll just merge it."

gaborbernat:
What I meant is that I find the formatter a higher level solution. It makes the code conform to the style; instead of warning if the style is not correct. So in some way having a formatter, makes having a linter redundant (as long as the formatter satisfies the linter checks).

tox.ini
```diff
@@ -57,3 +57,20 @@ deps = -r docs/requirements-docs.txt
...
+       isort == 4.3.3
+commands = black dateutil docs
+           isort -rc dateutil docs setup.py release.py
+
+[isort]
+multi_line_output = 3
+include_trailing_comma = True
+force_grid_wrap = 0
+combine_as_imports = True
+line_length = 120
```
jbrockmendel: I think 80 is the standard we've been (roughly) using

dateutil/parser/_parser.py
```diff
@@ -263,34 +260,56 @@ class parserinfo(object):
...
+        "and",
+        "ad",
+        "m",
+        "t",
+        "of",
+        "st",
+        "nd",
+        "rd",
+        "th",
+    ]
```
jbrockmendel: This class of whitespace/alignment change I'm not a fan of.  In some places below it introduces changes that flake8 will complain about.

Issue #736, opened 2018-06-04
Add setup_requires to pyproject.toml


pganssle:
As part of [PEP 518](https://www.python.org/dev/peps/pep-0518/), build dependencies should now go into `pyproject.toml`. I think that only `pip>=10.0` supports this feature at the moment, so I don't think we can remove `setup_requires` entirely, but the two do not conflict, because `pip` will install all the setup requirements *before* invoking `setup.py`, which will thus find all its setup requirements already satisfied.

anthrotype:
I can do this

PR #746, opened 2018-06-08
pyproject.toml: add PEP 518 'build-system'

anthrotype:
Fixes https://github.com/dateutil/dateutil/issues/736

setuptools and wheel are minimum requirements for build-system. I only added setuptools-scm
https://pip.pypa.io/en/stable/reference/pip/#pep-518-support

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,6 @@
+[build-system]
+requires = ["setuptools", "wheel", "setuptools_scm"]
+
 [tool.towncrier]
     package = "dateutil"
     package_dir = "dateutil"


anthrotype:
i don't think this minor patch deserve a changelog entry.. does it?

pganssle:
@anthrotype It definitely does. I'd go as far as saying it is a bugfix, because this actually fixes a real bug I have at work.

pganssle:
@anthrotype Do we need to update `MANIFEST.in` or anything to include `pyproject.toml`?  I want to make  sure that `pyproject.toml` is included in an sdist.

pganssle:
@anthrotype Can you add yourself to `AUTHORS.md`?

anthrotype:
ok I'll add one

anthrotype:
> Do we need to update MANIFEST.in or anything to include pyproject.toml?

I just tried to run python setup.py sdist, unzipped the tarball and I can see pyproject.toml file in the folder. So it's there, I don't think we need to modify the MANIFEST.in

changelog.d/746.bugfix.rst
```diff
@@ -0,0 +1 @@
+Add build requirements to pyproject.toml. (gh pr #746)
```
pganssle: Unless you don't care about getting credit in the changelog, I would word this as:

```
Add build requirements to pyproject.toml. Fixed by @anthrotype (gh issue #736, gh pr #746)
```

If you'd like to retain your commit signatures, please edit the commit history (or make a fixup commit and squash it when you are satisfied with the result).

pganssle: Either way you should add the issue number.

anthrotype:
I added the github issue to the changelog entry. Do you want me to squash all the commits into a single one?

pganssle:
@anthrotype No, I like a separate commit per logical change, so I only squash fixups.

PR #751, opened 2018-06-08
Closes #658 Fix Isoparser misinterprets T24:00

### LINKED ISSUE #658: Isoparser misinterprets T24:00
When I originally wrote the ISO 8601 parser, I misinterpreted the standard. From [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Times):

> Midnight is a special case and may be referred to as either "00:00" or "24:00". The notation "00:00" is used at the beginning of a calendar day and is the more frequently used. At the end of a day use "24:00". "2007-04-05T24:00" is the same instant as "2007-04-06T00:00" (see Combined date and time representations below).

But:

```python
>>> from dateutil.parser import isoparse
>>> isoparse("2007-04-05T24:00")
datetime.datetime(2007, 4, 5, 0, 0)
>>> isoparse("2007-04-06T00:00")
datetime.datetime(2007, 4, 6, 0, 0)
```

This is a bit tricky to handle because time and date parsing happen separately, so there's no obvious way to increment the date based on the value of the time until after it's parsed (at which point there is no difference between `00:00` and `24:00`).  I think the right solution is to allow [`_parse_isotime`](https://github.com/dateutil/dateutil/blob/master/dateutil/parser/isoparser.py#L318) to return `[24, 0, 0, 0, TZINFO]`, and handle that in [`parse_isotime`](https://github.com/dateutil/dateutil/blob/master/dateutil/parser/isoparser.py#L156) and [`parse_isodatetime`](https://github.com/dateutil/dateutil/blob/master/dateutil/parser/isoparser.py#L58) accordingly.

It's not clear to me whether the spec allows a bare time representation of `"24:00"` or if 24:00 is only valid in the context of a `datetime`. We should do whatever the spec says I think.

Cheukting:
## Summary of changes

Fix Isoparser misinterprets T24:00
Update test pytest.mark.parametrize '2014-04-11T24'
Update AUTHORS.md
Closes #658

### Pull Request Checklist
- [X] Changes have tests


pganssle:
Thanks for another great contribution @Cheukting!

Issue #383, opened 2017-05-22
Add mypy static type hinting


pganssle:
To facilitate downstream users who want to use static type hinting in their codebase, it seems like it's a good idea to use the `mypy`-style static type hinting in the `dateutil` codebase.  Some of this work has already been done in the [`typeshed` library](https://github.com/python/typeshed/tree/master/third_party/3/dateutil) ([2](https://github.com/python/typeshed/tree/master/third_party/2/dateutil)), assuming the license there is compatible with 3-clause BSD we may be able to use that.

To Do:
[ ] Add strict mypy checking to the CI.
[ ] Determine what might need to be done in the 2 vs. 3 case (I think they may be slightly different)

Modules to cover:
[ ] easter
[ ] parser
[ ] relativedelta
[ ] rrule
[ ] tz
[ ] zoneinfo

I am not sure if we want to do this entirely inline or with stubs - anyone who knows mypy maybe can chime in on the benefits of each. I would think the ideal would be to use the inline comment-style hinting, but I'm not sure how that works with 2/3 polyglot code.

cc @theghosthucodes who has volunteered to help out with this effort.

TheGhostHuCodes:
Yeah, it looks like adding type hints to the project repo is the preferred way to go (https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#adding-a-new-library) according to `typeshed`.

I've worked mostly with python 2.7 compatible "comment style" type hints within the `*.py` file and that seams to me like the easiest way to display and maintain the type hints in 2/3 polyglot:

 - You don't have to restate the function signature in a separate `*.pyi` file
 - You don't need an extra file open when you're working with code

pganssle:
Yeah, I like the idea of doing it inline. I'm guessing that the biggest 2/3 problem would be things that explicitly will have a return value of `six.text_type` or the like, since `[str,unicode]` is wrong for those. I'm guessing we can actually declare them as `[six.text_type]`?

TheGhostHuCodes:
Yeah, there's a type that's available in `typing` called `Text` that we can use for this. It is available as part of the standard library in python 3.5 I think, but will add a dependency for python 2.7 to the 3rd party `typing` library.

wagner-certat:
> Yeah, there's a type that's available in typing called Text that we can use for this. It is available as part of the standard library in python 3.5 I think, but will add a dependency for python 2.7 to the 3rd party typing library.

And also for 3.x < 3.5

[...]

PR #754, opened 2018-06-08
WIP: use monkeytype for generating typing annotation stubs

anthrotype:
Related to https://github.com/dateutil/dateutil/issues/383

I made a quick python script to call `monkeytype run -m pytest dateutil/test` and then `monkeytype stub <module>` on all the dateutil modules.
This is just an experiment, probably there are better ways.


anthrotype:
hey, this was more like a learning exercise for me, having never used type annotations, mypy (or even dateutil... :grimacing:) before.
Honestly, I don't think I'll have the "bandwith" to finish working on this PR.
If I compare this incomplete patch of mine with what is on typeshed for dateutil right now, I can see the other one seems to be hand-coded by someone with greater experience with both typing and dateutil API.
Feel free to pick up where I left, or close in favour of a better implementation.
Anyway, thanks for the opportunity! :+1:

PR #761, opened 2018-06-08
Add LRU caching to tzoffset, tzstr and gettz

### LINKED ISSUE #691: Add LRU cache to gettz and factories
Now that #635 is resolved, `gettz`, `tzoffset` and `tzstr` satisfies the property that two `tzinfo` objects (`tz1` and `tz2`) constructed with identical arguments will always be the same object (`tz1` is `tz2`), but the cached objects expire whenever no instances of a given timezone exist outside of the cache. This satisfies the "timezone semantics" reasons for adding a cache and will have good performance characteristics so long as at least one object holds a strong reference to the time zone.

But this *does* sacrifice performance for people who would prefer a cache because they are creating the same object over and over again even though it goes out of scope every time, so for example:

```python
def NY_timestamp(dt):
    dt.astimezone(tz.gettz('America/New_York')
    return dt.timestamp()

a = []
for i in range(10000):
    a.append(NY_timestamp(datetime.now()))
```

I think the above would needlessly create 10000 copies of `tz.gettz('America/New_York')`, when even a modest LRU cache would save the trouble.

I think we should leave the `_instances` dictionary as a `WeakValueDictionary` and add a second `_lru_cache` dictionary that holds a limited (and configurable) number of strong references to used time zones. Not sure the best implementation of the cache itself, preferably something where left-append, right-pop and reordering a single element are all O(1), so I think something implemented with a doubly-linked list (maybe `OrderedDict`).

Presumably we can take some cues from the implementation of [`functools.lru_cache`](https://github.com/python/cpython/blob/master/Lib/functools.py#L445), but keep in mind that the Python implementation may not have amazing performance, since CPython has a C implementation and may not have spent much time optimizing the Python implementation.

gokcennurlu:
## Summary of changes

I prepared a benchmark script and here are the results without the patch:

Benchmark:
https://gist.github.com/gokcennurlu/2b4e9d82e255c73e920c00ca5d26dcec

Results without patch:
```bash
3951.2012898921967 μs (tz.gettz() calls without storing values)
12.26398441940546 μs (tz.gettz() calls with storing values)


58.90091247856617 μs (tz.tzoffset() calls without storing values)
9.584207087755203 μs (tz.tzoffset() calls with storing values)


284.80793107300997 μs (tz.tzstr() calls without storing values)
7.534163072705269 μs (tz.tzstr() calls with storing values)
```

Results without patch:
```bash
27.99563854932785 μs (tz.gettz() calls without storing values)
28.635493479669094 μs (tz.gettz() calls with storing values)


22.882843017578125 μs (tz.tzoffset() calls without storing values)
25.380311533808708 μs (tz.tzoffset() calls with storing values)


24.550598114728928 μs (tz.tzstr() calls without storing values)
22.043749690055847 μs (tz.tzstr() calls with storing values)
```

'with storing values' means that a strong reference is kept in the scope (by doing `a = func(..)`) and lower values there explains the optimizing effect of currently used `WeakValueDictionary` based. caching.

Closes #691 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -44,6 +44,7 @@ switch, and thus all their contributions are dual-licensed.
 - Florian Rathgeber (gh: @kynan) **D**
 - Gabriel Bianconi <gabriel@MASKED> (gh: @GabrielBianconi) **D**
 - Gabriel Poesia <gabriel.poesia@MASKED>
+- Gökçen Nurlu <gnurlu1@bloomberg.net> (gh: @gokcennurlu) **D**
 - Gustavo Niemeyer <gustavo@niemeyer.net> (gh: @niemeyer)
 - Holger Joukl <holger.joukl@MASKED> (gh: @hjoukl)
 - Igor <mrigor83@MASKED>
--- a/changelog.d/761.feature.rst
+++ b/changelog.d/761.feature.rst
@@ -1,0 +1,1 @@
+Added a small "strong value" cache into ``tz.gettz``, ``tz.tzoffset`` and ``tz.tzstr`` to improve performance in the situation where transient references are repeatedly created to the same time zones, but no strong reference is continuously held. Patch by Gökçen Nurlu (gh issue #691, gh pr #761)
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -746,12 +746,12 @@ def test_tzoffset_weakref():
     del UTC1
     gc.collect()
 
-    assert UTC_ref() is not None # Because strong cache should keep it
+    assert UTC_ref() is not None    # Should be in the strong cache
     assert UTC_ref() is tz.tzoffset('UTC', 0)
 
     # Fill the strong cache with other items
     for offset in range(5,15):
-        tz.tzoffset('UTC', offset)
+        tz.tzoffset('RandomZone', offset)
 
     gc.collect()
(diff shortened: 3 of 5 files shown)


gokcennurlu:
Fixed the 2.7 `move_to_end` compatibility error, interestingly all seems faster now. I guess `if` branchings there are more expensive, than get/sets, in our cases.

```bash
16.1310862749815 μs (tz.gettz() calls without storing values)
19.910280592739582 μs (tz.gettz() calls with storing values)


14.881940744817257 μs (tz.tzoffset() calls without storing values)
17.241469584405422 μs (tz.tzoffset() calls with storing values)


15.357932634651661 μs (tz.tzstr() calls without storing values)
14.645632356405258 μs (tz.tzstr() calls with storing values)
```

gokcennurlu:
I am aware that CI fails and will fix `tzstr`'s tests by doing the same I've done for `tzoffset`'s tests. I'll cleanup the commits and might add comments, but cache's logic is ready IMO.

We can add test for `set_gettz_cache_size` too, I believe.

gokcennurlu:
Sorry for delaying this. I cleaned up the commits a little bit and I think it is ready for review.

pganssle:
No problem on delay. If you notice the project board review is currently the bottleneck.

PR #763, opened 2018-06-08
Closes #582: Allow partial minutes for python 3.6 or newer

### LINKED ISSUE #582: tzfile does not support non-integer minute offsets
Discovered this bug in #565 in the test for [Monrovia](https://www.timeanddate.com/time/zone/liberia/monrovia). I'm guessing that the problem is that before Python 3.6-ish, `datetime` didn't support this at all:

```python
>>> from dateutil import tz
>>> from datetime import *
>>> tzi = tz.tzoffset(None, timedelta(hours=12, seconds=30))
>>> datetime(2000, 1, 1, tzinfo=tzi)
datetime.datetime(2000, 1, 1, 0, 0, tzinfo=tzoffset(None, 43230))
>>> datetime(2000, 1, 1, tzinfo=tzi).utcoffset()
ValueError: tzinfo.utcoffset() must return a whole number of minutes
```

This should definitely be fixed in all versions of Python that support non-integer offsets, and the main question is what to do in the *other* cases. The options we have are:

1. Continue the current behavior of silently truncating to the nearest minute
2. Truncate the offset and add a warning in the *constructor* when a timezone with a non-integer number of minutes is constructed on a platform that doesn't support it.
3. Truncate the offset and add a warning in *utcoffset* or something when a platform doesn't support it.
4. Don't truncate at all and let `.utcoffset()` throw errors on platforms that don't support the non-truncated version.

I suspect that this affects a vanishingly small number of people, so my preferred solution is the one that has minimal impact on those people who are using timezones with integer number of minute offsets and people using a recent version of Python. I imagine that rules out a check in `utcoffset`. The lowest-cost thing to do is the fourth one, but that's the least friendly way. I'm thinking 2 is the right way to go.

Ideally we'll merge #580 and use that framework to distinguish between the different approaches.

This issue is currently blocking the merge of #565.

cssherry:
## Summary of changes

As described in python bug tracker (https://bugs.python.org/issue5288), datetime now allows offsets that are not an integer number of minutes. Dateutil now accounts for this when running in newer versions of python

- Add "Africa/Monrovia" to test_resolve_imaginary test for python 3.6 or newer
- Test that partial minutes is supported in python 3.6 or newer for tz.tzfile and datetime.utcoffset
- Enable partial minutes in both tz.tzoffset and tz.tzfile
- For now, don't warn user when rounding to nearest minute for python pre-3.6

Closes #582

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -156,7 +156,7 @@ END:VTIMEZONE
 EST_TUPLE = ('EST', timedelta(hours=-5), timedelta(hours=0))
 EDT_TUPLE = ('EDT', timedelta(hours=-4), timedelta(hours=1))
 
-SUPPORTS_PARTIAL_MINUTES = sys.version_info >= (3, 6)
+SUPPORTS_SUB_MINUTE_OFFSETS = sys.version_info >= (3, 6)
 
 
 ###
@@ -761,6 +761,25 @@ def test_tzoffset_singleton(args):
     tz2 = tz.tzoffset(*args)
 
     assert tz1 is tz2
+
+
+@pytest.mark.tzoffset
+@pytest.mark.skipif(not SUPPORTS_SUB_MINUTE_OFFSETS,
+                    reason='Sub-minute offsets not supported')
+def test_tzoffset_sub_minute():
+    delta = timedelta(hours=12, seconds=30)
+    test_datetime = datetime(2000, 1, 1, tzinfo=tz.tzoffset(None, delta))
+    assert test_datetime.utcoffset() == delta
+
+
+@pytest.mark.tzoffset
+@pytest.mark.skipif(SUPPORTS_SUB_MINUTE_OFFSETS,
+                    reason='Sub-minute offsets supported')
+def test_tzoffset_sub_minute_rounding():
+    delta = timedelta(hours=12, seconds=30)
+    test_date = datetime(2000, 1, 1, tzinfo=tz.tzoffset(None, delta))
+    assert test_date.utcoffset() == timedelta(hours=12, minutes=1)
+
 
 @pytest.mark.tzlocal
 class TzLocalTest(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


dateutil/tz/tz.py
```diff
@@ -1769,6 +1770,28 @@ def _datetime_to_timestamp(dt):
     return (dt.replace(tzinfo=None) - EPOCH).total_seconds()
 
 
+def supports_seconds():
```
pganssle: I would not like to change the public API at all with this, there should be no way to query this.

Also, this has a runtime cost when I would prefer an import-time cost.

dateutil/test/test_tz.py
```diff
@@ -1910,9 +1916,19 @@ def testFilestreamWithNameRepr(self):
 
     def testRoundNonFullMinutes(self):
         # This timezone has an offset of 5992 seconds in 1900-01-01.
-        tzc = tz.tzfile(BytesIO(base64.b64decode(EUROPE_HELSINKI)))
-        self.assertEqual(str(datetime(1900, 1, 1, 0, 0, tzinfo=tzc)),
-                             "1900-01-01 00:00:00+01:40")
+        # For python version pre-3.6, this will be rounded
+        set_as_old_python()
```
pganssle: There's no need to mock out the old behavior - we don't need to verify that this behavior is accurate on all versions of Python. It's best to split this into two tests, one for pre-3.6 and one for 3.6+, and then use `pytest.skipif` to run them selectively based on version.

dateutil/test/test_tz.py
```diff
@@ -1908,11 +1910,28 @@ def testFilestreamWithNameRepr(self):
         tzc = tz.tzfile(fileobj)
         self.assertEqual(repr(tzc), 'tzfile(' + repr('foo') + ')')
 
+    @pytest.mark.skipif(not SUPPORTS_PARTIAL_MINUTES, reason='Partial minute offset supported')
+    def testPartialMinuteSupport(self):
+        # If user running python 3.6 or newer, exact offset is used
+        tzc = tz.tzfile(BytesIO(base64.b64decode(EUROPE_HELSINKI)))
+        self.assertEqual(str(datetime(1900, 1, 1, 0, 0, tzinfo=tzc)),
```
pganssle: I think I would prefer a test that looks like this:

```python
assert datetime(1900, 1, 1, 0, 0, tzinfo=tzc).utcoffset() == timedelta(hours=1, minutes=39, seconds=52)
```

Though I am a bit torn because I will admit that the "compare the string" version sure is easy to read.

dateutil/test/test_tz.py
```diff
@@ -1908,11 +1910,28 @@ def testFilestreamWithNameRepr(self):
...
+                         "1900-01-01 00:00:00+01:39:52")
+
+        delta = timedelta(hours=12, seconds=30)
```
pganssle: I would prefer this to be a separate test (or better yet, a parametrized test).

cssherry: sure, made this and second part of testRoundNonFullMinutes separate tests

dateutil/test/test_tz.py
cssherry: The original was with string (see testRoundNonFullMinutes), but I can change them both to timedelta. 

However, note that the two would test slightly different things (`assert datetime(1900, 1, 1, 0, 0, tzinfo=tzc).utcoffset() == timedelta(hours=1, minutes=39, seconds=52)` tests the date offset through utcoffset function, while the current version tests just the date formation from specified offset)

cssherry: I'll make the change, but in separate commit for easy undo in case we decide to go with original method

dateutil/test/test_tz.py
```diff
@@ -1908,11 +1910,31 @@ def testFilestreamWithNameRepr(self):
...
+        assert test_date.utcoffset() == delta
+
+    @pytest.mark.skipif(SUPPORTS_PARTIAL_MINUTES, reason='Partial minute offset not supported')
+    def testRoundNonFullMinutesTzfile(self):
         # This timezone has an offset of 5992 seconds in 1900-01-01.
+        # For python version pre-3.6, this will be rounded
         tzc = tz.tzfile(BytesIO(base64.b64decode(EUROPE_HELSINKI)))
-        self.assertEqual(str(datetime(1900, 1, 1, 0, 0, tzinfo=tzc)),
-                             "1900-01-01 00:00:00+01:40")
+        assert datetime(1900, 1, 1, 0, 0, tzinfo=tzc).utcoffset(), timedelta(hours=1, minutes=40)
```
pganssle: I think this needs to be:

```python
assert datetime(1900, 1, 1, 0, 0, tzinfo=tzc).utcoffset() == timedelta(hours=1, minutes=40)
```

Issue #668, opened 2018-04-12
Use singletons or factories for all time zone classes


pganssle:
Similar issue to #342, which seeks to add caches to time zone functions to make lookups faster, we should *also* try to always return the exact same object, because Python's model is that two `datetime` objects are in the same timezone if and only if `dt1.tzinfo is dt2.tzinfo`. I have two blog posts on why this matters for datetime semantics:

1. [A curious case of non-transitive datetime comparison](https://blog.ganssle.io/articles/2018/02/a-curious-case-datetimes.html)
2. [Semantics of timezone-aware datetime arithmetic](https://blog.ganssle.io/articles/2018/02/aware-datetime-arithmetic.html)

As such, as a general rule we should try and return the same object for the same inputs. The highest priority objects to turn into factories:

- [ ] [`tzlocal`](https://dateutil.readthedocs.io/en/latest/tz.html#dateutil.tz.tzlocal) - See note below
- [ ] [`tzrange`](https://dateutil.readthedocs.io/en/latest/tz.html#dateutil.tz.tzrange)
- [ ] `tzwin`
- [ ] [`tzical`](https://dateutil.readthedocs.io/en/latest/tz.html#dateutil.tz.tzical)

## Note on `tzlocal` and `tzwinlocal`
Both `tzlocal` and `tzwinlocal` bind the time zone information *when constructed*, so if on a linix machine you do:

```python
from dateutil import tz

# Time zone is America/New_York
LOCAL1 = tz.tzlocal()
LOCAL2 = tz.tzlocal()

# change_timezone_to is an imaginary function that changes your machine zone
change_timezone_to('America/Pacific')
LOCAL3 = tz.tzlocal()
```

`LOCAL1` and `LOCAL2` will refer to something like `EST5EDT` and `LOCAL3` will refer to something like `PST8PDT`. This is complicated by the fact that the way `tzlocal` is *currently* implemented, this information is only *partially* bound at compile time, which is a bug.

It is not clear what the correct behavior is when a local time zone changes during the course of a program run. For now, the best way to implement a "tzlocal factory" that preserves the old behavior is to check whether the cached `tzlocal` would be the same as the one that you *would* construct and return a new object if and only if they would not be the same.

Ideally this would be done with minimal overhead, because part of the reason for using singletons is that they are much faster to "construct" than a new object every time.

jbrockmendel:
> when a local time zone changes during the course of a program run

[...]

PR #765, opened 2018-06-09
Add tzlocal singleton

alimcmaster1:
Use a metaclass to implement the singleton pattern for tzlocal class.

Partially closes <#668>

### DIFF
--- a/dateutil/tz/_factories.py
+++ b/dateutil/tz/_factories.py
@@ -1,5 +1,7 @@
 from datetime import timedelta
+import time
 import weakref
+
 
 class _TzSingleton(type):
     def __init__(cls, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


dateutil/tz/_factories.py
```diff
@@ -11,6 +13,29 @@ def __call__(cls):
...
+class _TzChangeSingleton(type):
+
+    """
+    A :class: `tzlocalsingleton`
+               Singleton that guards against TZ changes
+    """
+
+    def __init__(cls, *args, **kwargs):
+        cls.__instance = None
+        cls.__tz = time.tzname
```
pganssle: This guard is not as robust as one would hope, though it's probably a pretty good sentinel, consider:

```python
with TZEnvContext('EST5EDT'):
    l1 = tz.tzlocal()

with TZEnvContext('EST6EDT'):
    l2 = tz.tzlocal()
```

alimcmaster1:
Thanks @pganssle for the review. Updated as per comments and added test case. What do you think?

pganssle:
This is clever, but I'm not sure how fast it is.

That said, it's also got a problem, because setting the `TZ` variable doesn't instantly change the results of `time.tzname` and the other things that `tzlocal` looks for, it only changes after `time.tzset` has been called, so this will fail if you did this:

```python
import os
from dateutil import tz

os.environ['TZ'] = 'PST8PDT'
local0 = tz.tzlocal()
os.environ['TZ'] = 'EST5EDT'
local1 = tz.tzlocal()
time.tzset()
local2 = tz.tzlocal()
```

In that situation, `local0`, `local1` and `local2` will all be PST8PDT, because the cache will be invalidated between `local0` and `local1`, but the actual underlying hooks will not be returning anything different, and `local2` will hit the (incorrect) cache.

pganssle:
Whether or not it is ambiguous (I'm not sure I've ever seen it actually cause a problem), populating the cache based on changes in `os.environ` would severely exacerbate the problem, because normally:

```python
os.environ['TZ'] = 'EST5EDT'
local1 = tz.tzlocal()
time.tzset()
local2 = tz.tzlocal()
```

In the above situation, at the very least `local2` would be correct. With this cache, all subsequent `tzlocal` calls would return the broken version unless you change `TZ` to a third thing and back.

alimcmaster1:
Thanks okay understood definitely agree.

I can’t see a solution here apart from perhaps calling time.tzset() from within the cache if we detect a change in ‘TZ’ environment var since the last call to the cache. 

But this would be a breaking change if anyone has any functionality like the snippet below.

Think I’ll leave this here for now unless anyone else has any ideas?

alimcmaster1:
Spoke offline about this closing PR

PR #766, opened 2018-06-11
ISO parser fixes to seconds & minutes

danchr:
## Summary of changes

Fix handling of seconds an minutes in ISO 8601 timestamps. I ran into these when using 19th century dates in a Danish timezone with PostgreSQL. Apparently, we used [solar time in Copenhagen](https://en.wikipedia.org/wiki/Time_in_Denmark#History), which is UTC+00:50:20…

While at it, I noticed that the code would accept offsets such as UTC+00:-2, which I'm certain is unintentional.

Although this branch allows parsing fractional minutes on all versions of Python, it's only actually useful on Python 3.6, given that all other versions of Python when attempting to use or print the resulting offset. (#763 fixes that.) However, given that it fixes a proper bug, and results in a more helpful errors in these cases, I'd say it's somewhat independent of that change.

### Pull Request Checklist
- [x] Changes have tests


pganssle:
I don't think sub-minute offsets are specified in ISO 8601. Can you try to check if the later revisions (2004, 2014, 2016) support this? At the moment this is a basically standards-compliant parser (with the single exception of the separator, which can be omitted by mutual consent).

At the moment, I have tried to keep the public API somewhat sparse and eventually will allow for some customization of the parsed format to include some common variations on ISO 8601. In Python 3.7, `fromisoformat` supports fractional-minute offsets because `isoformat` can generate them. Likely the default for `isoparse` will end up being some superset of the formats supported by `fromisoformat` for practical reasons, with the ability to opt-in to a stricter parser.

pganssle:
At this point I'd very much like to merge at least the commit that rejects negative minutes/seconds etc, but the reason I'm not just splitting that out into a separate PR is that if a recent version of ISO 8601 *already* supports (even in a "by mutual consent" type way) sub-minute offsets, I'd like to merge the whole thing.

That said, when adding `fromisoformat` to CPython (which *does* have support for sub-minute and sub-second offsets), because both the naive portion of the time and the offset portion of the time ended up having the exact same format requirements (e.g. `HH[:MM[:SS[.fff[fff]]]]`), I just [wrote a single parse function that is used for both of them](https://github.com/python/cpython/commit/09dc2f508c8513e0466a759cc27a09108c1e55c2#diff-25e2d173c84057d069b7890450714eddR314). If it's true that there are just two "time portions" separated by a sign, we should probably do the same in `dateutil.parser.isoparse`.

danchr:
> At this point I'd very much like to merge at least the commit that rejects negative minutes/seconds etc, but the reason I'm not just splitting that out into a separate PR is that if a recent version of ISO 8601 already supports (even in a "by mutual consent" type way) sub-minute offsets, I'd like to merge the whole thing.

Alas, I don't have access to the actual ISO 8601 standard but RFC 3339 seems relevant. It contains [this chunk](https://tools.ietf.org/html/rfc3339#section-4.2):

>       NOTE: Following ISO 8601, numeric offsets represent only time
>       zones that differ from UTC by an integral number of minutes.
>       However, many historical time zones differ from UTC by a non-
>       integral number of minutes.  To represent such historical time
>       stamps exactly, applications must convert them to a representable
>       time zone.

So, it might actually be a technical violation of ISO 8601 to emit these strings. Regardless, they're quite unambiguous, so rejecting them seems to offer little concrete value.

pganssle:
@danchr As I mentioned above, I'm hoping to make it so that `isoparse` can also be used for validating ISO-8601 datetimes, so for the moment I'm not including any extensions to the spec. My long-term vision is to create a series of feature flags that can make the behavior more or less strict than the default, at which point people using `isoparse` for ISO 8601 validation can use a provided `STRICT` object and for everyone else we'll relax the constraints a bit.

danchr:
I'd image it as a `strict` flag, or similar, to `isoparse()` that wasn't on by default — after all, most people want to use a parse function for parsing data that _might_ from some source that slightly deviates from the standard. While at it, non-strict mode could perhaps relax the requirement that the time/date separator is `T`?

The high-level reason for asking for this is that the addition of ISO 8601 support makes `dateutil` the _only_ dependancy you need for handling datetime issues. The parser works very well, your timezone implementation is superior to that in `pytz`, and it even parses ISO 8601 if the Python version doesn't. So, this minor correctness issue bugged me — we're getting a timezone offset we can't parse — even _if_ no-one will care in practice.

pganssle:
I'll note that I have interpreted ISO 8601's "the separator T can be changed by mutual agreement" as "you can put anything for the separator as long as it's exactly one character", despite the fact that in some versions of the standard, a space is *explicitly* disallowed as the separator (I'm guessing this is because ISO8601 was intended to be used in space-delimited data formats). The stricter behavior is available as `dateutil.parser.isoparser(sep='T').isoparse`. I'm willing to take the most liberal interpretation possible for the standard.

pganssle:
OK, I checked ISO/WD 8601:2016 (latest version), here's the relevant section:

> #### 4.2.5.1 Difference between local time and UTC of day
> When it is required to indicate the difference between local time and UTC of day, the representation of
the difference can be expressed in hours and minutes, or hours only. It shall be expressed as positive
(i.e. with the leading plus sign [+]) if the local time is ahead of or equal to UTC of day and as negative
(i.e. with the leading minus sign [-]) if it is behind UTC of day. The minutes time element of the
difference may only be omitted if the difference between the time scales is exactly an integral number of hours.
>
>> Basic format:   ±hhmm, ±hh
>> Example: +0100, +01
>
>> Extended format: ±hh:mm
>> Example: +01:00
>
> Expressions of the difference between local time and UTC of day are a component in the representations defined in 4.2.5.2; they shall not be used as self-standing expressions.

No provision is made for sub-second offsets, probably because ISO 8601 is not amazingly great, so this will have to wait until we have a mechanism to make `isoparse` more strict, which is high on the priority list. I'd like to accept this PR without the sub-second stuff and leave the sub-second stuff for a second PR when we can still allow `isoparse` to be strict.

PR #768, opened 2018-06-12
function annotations for utils.py

jpurviance:
## Summary of changes

First pass on function annotations for tools like `Mypy` and `pyre` for utils.py. 
addresses #383 for utils.py

- Not sure how I feel about the fact that imports now exists at the top of the file just for type checking. 
- Not sure how well this will work for modules that depend on this code (will they know that the annotation `def within_delta(dt1, dt2, delta):  # type: (datetime, datetime, timedelta) -> bool` assumes that `datetime`is coming from the `datetime` module. `Mypy` knows this when it's ran on this module. Will the code that imports this know this?)

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/dateutil/utils.py
+++ b/dateutil/utils.py
@@ -7,10 +7,11 @@ datetimes.
 """
 from __future__ import unicode_literals
 
-from datetime import datetime, time
+from datetime import datetime, time, tzinfo, timedelta
+from typing import Union
 
 
-def today(tzinfo=None):
+def today(tzinfo=None):  # type: (Union[tzinfo, None]) -> datetime
     """
     Returns a :py:class:`datetime` representing the current day at midnight
 
@@ -26,7 +27,7 @@ def today(tzinfo=None):
     return datetime.combine(dt.date(), time(0, tzinfo=tzinfo))
 
 
-def default_tzinfo(dt, tzinfo):
+def default_tzinfo(dt, tzinfo):  # type: (datetime, tzinfo) -> datetime
     """
     Sets the the ``tzinfo`` parameter on naive datetimes only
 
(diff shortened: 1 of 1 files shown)


jpurviance:
Chose `utils.py` because it's always best to start adding annotations to modules that depend on very little and are depended on by many others.

pganssle:
I highly doubt anyone depends on utils.py, FWIW.

dateutil/utils.py
```diff
@@ -7,10 +7,11 @@
 """
 from __future__ import unicode_literals
 
-from datetime import datetime, time
+from datetime import datetime, time, tzinfo, timedelta
+from typing import Union
 
 
-def today(tzinfo=None):
+def today(tzinfo=None):  # type: (Union[tzinfo, None]) -> datetime
```
pganssle: I don't think the None needs to be explicit here.

dateutil/utils.py
```diff
@@ -7,10 +7,11 @@
 """
 from __future__ import unicode_literals
 
-from datetime import datetime, time
+from datetime import datetime, time, tzinfo, timedelta
+from typing import Union
```
pganssle: All typing-related imports needs to go under a guard:

```python
if False:
    from typing import Union
    from datetime import tzinfo, timedelta
```

See [this](https://github.com/dateutil/dateutil/issues/383#issuecomment-396038661) for more details.

pganssle:
It's probably worth having a look at the stubs @anthrotype generated using monkeytype in PR #754. [Here are the ones for utils.py](https://github.com/dateutil/dateutil/pull/754/files#diff-b10503c960155145cf669083bc948102). I think they are approximately correct, except that in the generated ones they haven't abstracted `tzutc` and `tzfile` to `tzinfo`.

dateutil/utils.py
pganssle: That said I believe you don't actually need the `Union` import.

pganssle:
Another thing to note is that we need to add `mypy` to the CI before I can start merging these.

jpurviance:
I agree that this code is probably not depended on by many people. I've chosen to start with utils.py not because its the best code to annotate for those that use dateutil, but instead because its a easy place to start for mypy ci integration. 

When adopting type annotations, its best to control the scope of what mypy checks. If its run on a module that imports from many others its either going to have trouble understanding the types of that code or make some assumptions about it that you don't want. I've started here so that when this code is imported into other modules (even if its just the tests), the correct assumptions are made by mypy and meaningful errors are provided.

PR #749, opened 2018-06-08
Closes #124: Fix remaining doctests

rs2:
## Summary of changes

Fix remaining doctests.

Closes #124 

### Pull Request Checklist
- [x] Changes have tests
- [x] Authors have been added to [AUTHORS.md](https://github.com/dateutil/dateutil/blob/master/AUTHORS.md)
- [x] News fragment added in changelog.d. See [CONTRIBUTING.md](https://github.com/dateutil/dateutil/blob/master/CONTRIBUTING.md#changelog) for details

PR #769, opened 2018-06-12
Fix doctests and run as part of CI

### LINKED ISSUE #124: Fix remaining doctests
@moreati Has fixed a large number of the doctests in #120, but apparently there are still some remaining. We should go through and fix or remove those, as appropriate.

pganssle:
## Summary of changes
Fixes remaining doctests. This supercedes PR #749

Closes #124 

### Pull Request Checklist

### DIFF
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -1728,9 +1728,9 @@ def resolve_imaginary(dt):
         >>> print(tz.resolve_imaginary(datetime(2017, 3, 12, 2, 30, tzinfo=NYC)))
         2017-03-12 03:30:00-04:00
 
-        >>> KIR = tz.gettz('Pacific/Kiritimati')
-        >>> print(tz.resolve_imaginary(datetime(1995, 1, 1, 12, 30, tzinfo=KIR)))
-        1995-01-02 12:30:00+14:00
+        >>> APIA = tz.gettz('Pacific/Apia')
+        >>> print(tz.resolve_imaginary(datetime(2011, 12, 30, 12, tzinfo=APIA)))
+        2011-12-31 12:00:00+14:00
 
     As a note, :func:`datetime.astimezone` is guaranteed to produce a valid,
     existing datetime, so a round-trip to and from UTC is sufficient to get


setup.cfg
```diff
@@ -5,4 +5,5 @@ universal = 1
 license_file = LICENSE
 
 [tool:pytest]
-xfail_strict = true
\ No newline at end of file
+xfail_strict = true
+doctest_optionflags = ALLOW_UNICODE ELLIPSIS
```
jbrockmendel: I think that if "NORMALIZE_WHITESPACE" is added then some long lines can be wrapped.

There's also an option "addopts = --doctest-modules" that I've tried but that has broken for me on non-Windows by trying to import tzwin.  It'd be nice if that was somehow available.

conftest.py
```diff
@@ -0,0 +1,16 @@
+import sys
+
+collect_ignore = []
+if sys.platform != 'win32':
+    collect_ignore += [
+        'dateutil/tzwin.py',
+        'dateutil/tz/win.py',
+    ]
+
```
jbrockmendel: Some of the doctests could be cleaned up by adding e.g.

```

@pytest.fixture(autouse=True)
def add_imports(doctest_namespace):
    doctest_namespace["datetime"] = datetime
    [...]
    doctest_namespace["parser"] = parser
```

jbrockmendel:
Couple comments/suggestions.  LGTM.

pganssle:
Hm... This is a problem, it turns out that `pytest` doesn't support Sphinx's `testsetup` or `testcleanup` blocks, and it seems like it uses the scope of the file it happens to be executed in. That is... not ideal.

We may need to reconsider using `pytest` for this at all and instead do it through `sphinx` directly.

conftest.py
rs2: True that.

AnneTheAgile:
^^ above note that pytest cannot be used was confirmed it seems by 
nicoddemus   on Mar 5, 2018 there who said it could not be easily changed.

Maybe this ticket should revert the CI part and just get the fixes to doctest in the repo.
Ty to everyone it seems like this repo is quite well maintained!

tox.ini
```diff
@@ -9,17 +9,24 @@ envlist = py27,
...
+# Tox 2.9.0 is the last version that supports Python 3.3
 minversion = 2.9.0
 skip_missing_interpreters = true
 
 [testenv]
 description = run the unit tests with pytest under {basepython}
 setenv = COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 passenv = DATEUTIL_MAY_CHANGE_TZ TOXENV CI TRAVIS TRAVIS_* APPVEYOR APPVEYOR_* CODECOV_*
-commands = python -m pytest -m "not xfail" {posargs: "{toxinidir}/dateutil/test" --cov-config="{toxinidir}/tox.ini" --cov=dateutil}
+commands = python -m pytest -m "not xfail" --doctest-modules {posargs: "{toxinidir}/dateutil" --cov-config="{toxinidir}/tox.ini" --cov=dateutil}
```
thisch, line 20: IMO `--doctest-glob=*.rst` is also needed, s.t. the doctests in the rst files are run. Note that pytest only runs the tests that don't use the sphinx-directives. We could use pytest-sphinx to run all off them.

PR #785, opened 2018-07-02
Start testing against master branch of tz project

### LINKED ISSUE #617: Start testing against master on tz's github
To be alerted ahead of time if there are any problems, we should start a weekly test build against master on the [IANA tz database](https://github.com/eggert/tz). Since we compile the sources from zic in the tests anyway, it shouldn't be a big problem to clone the repo and run tests against it.

pganssle:
## Summary of changes
Closes #617.

### Pull Request Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
   include:
     - python: 3.6
       env: TOXENV=docs
+    - python: 3.6
+      env: TOXENV=tz
     - python: 3.7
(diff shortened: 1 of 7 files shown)


pganssle:
This runs the full test suite in case some of the other tests (like the parser tests) indirectly rely on the time zone stuff.

That said, I think if we're diligent about marking stuff that is an actual test of something related to the IANA and zoneinfo stuff, we can probably cut down the number of tests run.

PR #787, opened 2018-07-03
Accept more than 6 fractional digits in `isoparse`

### LINKED ISSUE #786: Accept more than 6 digits for fractional seconds in isoparse
Currently, `isoparse` raises an exception if there are more than 6 digits in the fractional second.

```
>>> dateutil.parser.isoparse("2006-01-02T15:04:05.9999999-07:00")
  File "/home/jay/.local/lib/python2.7/site-packages/dateutil/parser/isoparser.py", line 37, in func
    return f(self, str_in, *args, **kwargs)
  File "/home/jay/.local/lib/python2.7/site-packages/dateutil/parser/isoparser.py", line 136, in isoparse
    components += self._parse_isotime(dt_str[pos + 1:])
  File "/home/jay/.local/lib/python2.7/site-packages/dateutil/parser/isoparser.py", line 363, in _parse_isotime
    raise ValueError('Unused components in ISO string')
ValueError: Unused components in ISO string
```

Go, for example, includes up to 9 digits (nanosecond precision) when marshaling a time into JSON. It would be nice if `isoparse` could accept those without issue, even if precision gets truncated a bit.

jayschwa:
RFC 3339 does not specify a limit to the number of fractional digits
that can be provided. In the wild, languages like Go will provide up
to 9 digits by default.

Since the standard library's `datetime` only supports precision down
to microseconds, any extra digits in the fractional component are
truncated.

Closes #786

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/test/test_isoparser.py
+++ b/dateutil/test/test_isoparser.py
@@ -130,6 +130,15 @@ def test_ymd_hms_micro(dt, date_fmt, time_fmt, tzoffset, precision):
 
     _isoparse_date_and_time(dt, date_fmt, time_fmt, tzoffset, precision)
 
+###
+# Truncation of extra digits beyond microsecond precision
+@pytest.mark.parametrize('dt_str', [
+    '2018-07-03T14:07:00.123456000001',
+    '2018-07-03T14:07:00.123456999999',
+])
+def test_extra_subsecond_digits(dt_str):
+    assert isoparse(dt_str) == datetime(2018, 7, 3, 14, 7, 0, 123456)
+
 @pytest.mark.parametrize('tzoffset', FULL_TZOFFSETS)
 def test_full_tzoffsets(tzoffset):
     dt = datetime(2017, 11, 27, 6, 14, 30, 123456)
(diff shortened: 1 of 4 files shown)


jayschwa:
I referenced RFC 3339 because ISO 8601 appears to be behind a paywall :cry:

dateutil/test/test_isoparser.py
```diff
@@ -473,12 +482,11 @@ def test_isotime_midnight(isostr):
     ('14時30分15秒', ValueError),                # Not ASCII
     ('14_30_15', ValueError),                   # Invalid separators
     ('1430:15', ValueError),                    # Inconsistent separator use
-    ('14:30:15.3684000309', ValueError),        # Too much us precision
     ('25', ValueError),                         # Invalid hours
     ('25:15', ValueError),                      # Invalid hours
     ('14:60', ValueError),                      # Invalid minutes
     ('14:59:61', ValueError),                   # Invalid seconds
-    ('14:30:15.3446830500', ValueError),        # No sign in time zone
+    ('14:30:15.34468305:00', ValueError),       # No sign in time zone
```
jayschwa: I wasn't quite sure about the original intent of this test case. The other case I see that says "No sign in time zone" has a `:` within the zone, but lacks a `-` or `+` prefix, so I followed suit here.

pganssle: Yes that is the right thing to do, though it does bring up an unfortunate ambiguity. Anyone who is trying to validate ISO8601 datetimes will silently accept `%Y-%m-%dT%H:%M:%S.%fhhmm`.

I think this is OK because there are other ways to generate valid ISO8601 strings that do not represent the intended datetime.

pganssle:
Excellent PR. Doesn't look like it slows down the parse very much (if at all), either.

pganssle:
Thanks for your contribution @jayschwa

PR #795, opened 2018-08-09
Issue 523

### LINKED ISSUE #523: Invalid byxxx values allowed in rrule constructor
The current value checking for `byyearday`, `bymonthday`, `bymonth`, etc, only validate that the combinations of `byxxx` rules don't lead to an infinite loop, however, they allow arbitrary values, e.g.:

```python
from dateutil.rrule import YEARLY, DAILY, rrule

rr = rrule(freq=DAILY, bymonthday=32, count=1)
print(list(rr))     # []

rr = rrule(freq=YEARLY, byyearday=377, count=1)
print(list(rr))     # []

rr = rrule(freq=YEARLY, bymonth=13, count=1)
print(list(rr))     # []
```

Actually running each of these takes a while, because the iterator will generate all the possible values before `byxxx` constraints and reject them if invalid, even though we know ahead of time that these constraints are invalid.

This can be fixed one of two ways, depending on what the RFC has to say about it:

1. We raise a `ValueError` when there are out-of-range value components if the RFC restricts all components to be valid values.
2. If the RFC explicitly says that you must accept invalid `byxxx` values, we add in a `_generator_empty` attribute that is set in the constructor, and if it's `True` just return immediately in the iteration step.
3. If the RFC makes no comment either way, we choose one of the above two (I am leaning towards `ValueError` because it's almost certainly a mistake, but that has API consequences that "empty generator" doesn't).

absreim:
## Summary of changes

Adds range checking for byxxx values for rrule. The constructor now raises a ValueError if a value is out of range.

Closes #523 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -1128,7 +1128,8 @@ class rrule(rrulebase):
             else:
                 raise ValueError(bymonth_err_str)
         if bymonthday is not None:
-            bymonthday_err_str = "BYMONTHDAY values must all be integers from 1 to 31 or from -31 to -1."
+            bymonthday_err_str = ("BYMONTHDAY values must all be integers from " 
+                                  "1 to 31 or from -31 to -1.")
             if isinstance(bymonthday, integer_types):
                 if abs(bymonthday) > 31 or abs(bymonthday) < 1:
                     raise ValueError(bymonthday_err_str)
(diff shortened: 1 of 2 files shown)


pganssle:
Hm, this is a lot of code. Is it going to slow things down significantly?

absreim:
Not sure. I'll try to come up with some benchmarks and get some hard numbers.

absreim:
@pganssle 
I just ran some tests using timeit and found that there is a significant performance impact: the change makes it take about 20-40% longer to finish execution for a few moderately complex rrules.
I don't see any obvious ways to make things much faster. There is small amount of duplicated effort between the _validate_rule function and existing code that involves type checking of the parameters but there is no getting around having to check every parameter to see that it is in range.

PR #799, opened 2018-08-11
Pin setuptools version on Python 3.3

pganssle:
Not sure why but `setuptools` seems to have started breaking on Python 3.3


wagner-certat:
Because Python 3.3 is out of support and setuptools dropped support for it: https://setuptools.readthedocs.io/en/latest/history.html#v40-0-0

pganssle:
It's a bit more complicated than that, I'm a maintainer for setuptools as well, `python_requires` should have installed an older version of `setuptools` without the pinning.

PR #801, opened 2018-08-14
Stop trying to add None to the weakref cache

### LINKED ISSUE #800: tz.gettz broken when time zone not found
When we switched over to using weak references for caching (see #635), we didn't take into account the fact that occasionally `tz.gettz` returns `None` (when no `tzinfo` object is found). Since you cannot have a weak reference to `None`, when we try to populate the weak cache, it raises an error.

This can be reproduced with:

```python
from dateutil import tz
tz.gettz('🐼')
```

The result:
> TypeError: cannot create weak reference to 'NoneType' object

This is causing failures in pandas' test suite: https://github.com/pandas-dev/pandas/issues/22337

pganssle:
## Summary of changes
   
None is a singleton and a weak reference to it will never expire, and in this case it is used more as a soft failure mode than as an actual value to return. By special-casing this, the invariant that:

 ```python
gettz(x) is gettz(x)
```

is preserved in all cases *except* when something on the underlying system changes such that `x` does not refer to a valid time zone in the first call, but does refer to a valid time zone in the second call. This could happen if `x` is added to the zoneinfo data, or if the local time zone changes such that `x` is now a valid specifier for the local time.

Closes #800 and fixes pandas-dev/pandas#22337

### Pull Request Checklist
- [X] Changes have tests
- [x] News entry


pganssle:
This is only failing because [gnu.org](https://www.gnu.org) is down, which seems to be a temporary situation.

PR #810, opened 2018-09-02
Fix bug with Samoan offset change, refactor base offset change handling

### LINKED ISSUE #812: utcoffset/tzname incorrect immediately before Samoan offset change
First reported by @MartinThoma in [this SO question](https://stackoverflow.com/questions/52136980), it seems that there is a bug in the way `tzfile` handles the local transition times during a base offset change:

```python
from dateutil import tz

APIA = tz.gettz('Pacific/Apia')
dt0 = datetime.fromisoformat('2011-12-29T20:00-10:00')

for i in range(5):
    dt = (dt0 + timedelta(hours=i))
    dt = dt.astimezone(APIA)
    print(f'{dt.isoformat()}')

## Result:
# 2011-12-29T20:00:00-10:00
# 2011-12-29T21:00:00-10:00
# 2011-12-29T22:00:00-10:00
# 2011-12-29T23:00:00+14:00
# 2011-12-31T00:00:00+14:00
```

Note that while `fromutc` sets the local datetime correctly, `utcoffset` returns `+14:00` an hour early! This is because the assumption in the code is that these base offset shifts don't happen during DST, which leads to the "local" (but not the UTC) transition time being off by 1 DST offset.

PR #810 should fix this.

pganssle:
This is a test and fix for the bug first reported [in this StackOverflow question](https://stackoverflow.com/q/52136980/467366), and migrated to GH in issue #812.

When cleaning that up, I discovered a cleaner way to handle the issue of Portugal's base offset shift without a corresponding total offset shift, obviating the need for a second traversal of the transition list.

Fixes #812.

### DIFF
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -657,37 +657,44 @@ class tzfile(_tzinfo):
         # isgmt are off, so it should be in wall time. OTOH, it's
         # always in gmt time. Let me know if you have comments
         # about this.
-        laststdoffset = None
+        lastdst = None
+        lastoffset = None
+        lastdstoffset = None
+        lastbaseoffset = None
         out.trans_list = []
+
         for i, tti in enumerate(out.trans_idx):
-            if not tti.isdst:
-                offset = tti.offset
-                laststdoffset = offset
-            else:
-                if laststdoffset is not None:
-                    # Store the DST offset as well and update it in the list
-                    tti.dstoffset = tti.offset - laststdoffset
-                    out.trans_idx[i] = tti
+            offset = tti.offset
+            dstoffset = 0
 
-                offset = laststdoffset or 0
+            if lastdst is not None:
+                if tti.isdst:
+                    if not lastdst:
+                        dstoffset = offset - lastoffset
 
-            out.trans_list.append(out.trans_list_utc[i] + offset)
+                    if not dstoffset and lastdstoffset:
+                        dstoffset = lastdstoffset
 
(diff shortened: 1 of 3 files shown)


pganssle:
I believe the issue here is an erroneous assumption that a base offset change will not happen while DST is active. It will take a bit of careful thinking to unravel this.

pganssle:
I don't see anything wrong with this PR, but I did merge this more quickly than I usually do (feel free to blame it on a baby distracting me). If anyone has code review please still submit it and I will follow up with a fix.

I would like to refactor how `trans_idx` works - it can be done much more efficiently if the `ttinfo` objects were cached.

I also need to add some tests around `Europe/Dublin` (with its negative DST), and if there are any examples of base offset change during I'd like to add them to the test suite.

dateutil/tz/tz.py
```diff
@@ -657,37 +657,44 @@ def _read_tzfile(self, fileobj):
...
+            # If a time zone changes its base offset during a DST transition,
+            # then you need to adjust by the previous base offset to get the
+            # transition time in local time. Otherwise you use the current
+            # base offset. Ideally, I would have some mathematical proof of
+            # why this is true, but I haven't really thought about it enough.
+            baseoffset = offset - dstoffset
+            adjustment = baseoffset
+            if (lastbaseoffset is not None and baseoffset != lastbaseoffset
+                    and tti.isdst != lastdst):
+                # The base DST has changed
```
pganssle: Hm, this comment can probably be dropped, doesn't make much sense.

PR #813, opened 2018-09-08
Always use io.BytesIO/StringIO over six; on all supported Pythons

jdufresne:
More forward compatible by slightly reducing unnecessary use of the six library.


pganssle:
LGTM.

PR #814, opened 2018-09-09
setup.py test doesn't work, so exit with 1

hroncok:
## Summary of changes

The error message was printed, but the process exited with 0.
That is extremely error prone in automated environments.

Also, print the error message to standard error.

This is **completely untested** as I'm doing the change from the web interface.

### Pull Request Checklist
- [x] Changes have tests - AFAIK there are no setup.py related tests here
- [x] Authors have been added to [AUTHORS.md](https://github.com/dateutil/dateutil/blob/master/AUTHORS.md) - consider this trivial and Public Domain
- [x] News fragment added in changelog.d. See [CONTRIBUTING.md](https://github.com/dateutil/dateutil/blob/master/CONTRIBUTING.md#changelog) for details - I'm doing this from the GitHub web interface and it starts to be rather complicated :(

I will eventually get to this from a machine where I can clone and rebase and stuff.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ from distutils.version import LooseVersion
 import warnings
 
 import io
+import sys
 
 if isfile("MANIFEST"):
     os.unlink("MANIFEST")
@@ -21,8 +22,9 @@ if LooseVersion(setuptools.__version__) <= LooseVersion("24.3"):
 
 class Unsupported(TestCommand):
     def run(self):
-        print("Running 'test' with setup.py is not supported. "
-              "Use 'pytest' or 'tox' to run the tests.")
+        sys.stderr.write("Running 'test' with setup.py is not supported. "
+                         "Use 'pytest' or 'tox' to run the tests.\n")
+        sys.exit(1)
 
 
 ###
--- a/changelog.d/814.bugfix.rst
+++ b/changelog.d/814.bugfix.rst
@@ -1,0 +1,1 @@
+Changed the ``python setup.py test`` to print an error to ``stderr`` and exit with 1 instead of 0. Reported and fixed by @hroncok (gh pr #814)


setup.py
```diff
@@ -21,8 +22,9 @@
 
 class Unsupported(TestCommand):
     def run(self):
-        print("Running 'test' with setup.py is not supported. "
-              "Use 'pytest' or 'tox' to run the tests.")
+        sys.stderr.write("Running 'test' with setup.py is not supported. "
```
pganssle: Hm, I'm torn - it seems like it would be easier to raise an exception, but that's pretty spammy.

pganssle:
This seems fine I suppose. I wouldn't go out of my way to support this, since anything that is just blindly invoking `setup.py test` is essentially fundamentally broken anyway, but I don't think it hurts anything to use this.

hroncok:
A motivation/story:

We have dateutil as an RPM package in Fedora. It has `setup.py test` in the spec file (a recipe to build the package). I'm not sure when this worked and when not, but once upon a time it stopped running the tests and the packagers doing updates do not always examine the build logs, assuming that tests passed if there was no error.

Today I've seen a Pull Request within Fedora to update dateutil and I've jumped to review it, examined the build logs and have seen the error. Of course, we will fix the invocation in Fedora, yet I thought that such error should not pass silently.

pganssle:
Ah yes, it's been so long I forgot that we actually transitioned from an earlier method that used `python setup.py test`. That makes this much more relevant. Will merge when CI passes.

hroncok:
Thank You for doing that changelog entry!

pganssle:
@hroncok Would you prefer to be credited with your full name instead in the changelog?

pganssle:
CI failure is due to #816, so I'll rebase when that is merged into master.

@hroncok It's not trouble for me to change how you are credited, I'm going to rebase anyway, so I'm happy to change to your full name if you want. The changelog is a mix of real names and github handles.

hroncok:
I mean it, I really don't care.

PR #817, opened 2018-09-14
Remove unnecessary binary_type compatibility shim

jdufresne:
The `bytes` type is available on all supported Pythons. Can remove workaround and make the code more forward compatible with Python 3.


pganssle:
Just needs a changelog - I can add one.

PR #818, opened 2018-09-14
Use simpler hasattr to check if a method exists

jdufresne:


pganssle:
Using `getattr` is a deliberate choice, see [hasattr(): A Dangerous Misnomer](https://hynek.me/articles/hasattr/).

jdufresne:
Very interesting! Thanks for that link.

PR #819, opened 2018-09-19
Change sphinx pin to != 1.8.0

pganssle:
The callable function typing issue should be fixed in sphinx commit bbb4593bfaa65611b6bd600c5, so the next release after 1.8.0 should contain the fix.

Fixed in GH sphinx-doc/sphinx#5431

https://github.com/sphinx-doc/sphinx/commit/bbb4593bfaa65611b6bd600c5e37c4c54af87f1a

Probably best to wait until a new version of Sphinx is released to merge this rather than just having to revert it if the new version still doesn't work.


pganssle:
Version 1.8.1 was released and works.

PR #833, opened 2018-10-30
Changed for bytes/str conversion lines not hit in test_isoparser

### LINKED ISSUE #776: bytes/str conversion lines not hit in test_isoparser
I've noticed that in the [coverage for `test_isoparser`](https://codecov.io/gh/dateutil/dateutil/src/572e358279fb0d90862e2f5fd8d55300a9670fcf/dateutil/test/test_isoparser.py), lines [375](https://github.com/dateutil/dateutil/blob/a41c4c6b498189b1c456e2d8e7a21a308974bc89/dateutil/test/test_isoparser.py#L375), [447](https://github.com/dateutil/dateutil/blob/a41c4c6b498189b1c456e2d8e7a21a308974bc89/dateutil/test/test_isoparser.py#L447) and [449]([447](https://github.com/dateutil/dateutil/blob/a41c4c6b498189b1c456e2d8e7a21a308974bc89/dateutil/test/test_isoparser.py#L449)) are not being hit. I can't tell if this means that the tests aren't being run against both `bytes` and `str` or if it means that I have unnecessary `if` conditions in the tests.

If the former, we need to make sure tests are being run against `bytes` and `str`. If the latter, we can remove the unnecessary branches.

The parts that are not covered because they are under an `xfail` can have `#pragma: no cover` added to them, as can the error handling lines in the test generation part.

jramacha:
## Summary of changes
@pganssle  mentioned some branches were not hit in test_isoparser.py. Ran the test and found that the test weren't hitting the is instance(param, six.text_type) for python2.7 , because the string value returned from strftime was not a unicode string, so converted that to unicode and made sure the test were hitting both the branches. And test weren't hitting the isinstance(param, bytes) for python3+ because the string value returned from strftime always failed that check as it was unicode string. So added a line to encode that string to ascii and made sure the test hit the other branch. 

Also added  #pragma: no cover to lines not covered.

Closes #776


dateutil/test/test_isoparser.py
```diff
@@ -379,11 +379,19 @@ def __make_date_examples():
 @pytest.mark.parametrize('d,dt_fmt', __make_date_examples())
 @pytest.mark.parametrize('as_bytes', [True, False])
 def test_parse_isodate(d, dt_fmt, as_bytes):
-    d_str = d.strftime(dt_fmt)
-    if isinstance(d_str, six.text_type) and as_bytes:
-        d_str = d_str.encode('ascii')
-    elif isinstance(d_str, bytes) and not as_bytes:
-        d_str = d_str.decode('ascii')
+    if six.PY2:
```
pganssle: I would prefer it if this were written in polyglot style, so nothing that makes explicit reference to `six.PY2`.

That said, this is the wrong thing to do. The original code block was this (I've annotated with comments to make it clear what's happening):

```python
if isinstance(d_str, six.text_type) and as_bytes:
    # If d_str is str/unicode, convert it into bytes
    d_str = d_str.encode('ascii')
elif isinstance(d_str, bytes) and not as_bytes:
    # If d_str is bytes, convert it to str/unicode
    d_str = d_str.decode('ascii')
```

The problem is that I'm expecting to run the same code with `as_bytes=True` and `as_bytes=False`, which means that whatever `d_str` is, it should have to be converted to the other thing half the time. In Python 2 I expect it to be `bytes` and in Python 3 I expect it to be `str`, so *both* of these branches should *already* be hit.

`unicode(b'example')` is the wrong way to convert from bytes to string, the correct way is `b'example'.encode()`, which is what I'm already doing.

Anyway, either something has changed since the original issue was raised or I made a mistake originally, because it seems like these lines are indeed covered, and only the later lines (with `time_val`) are having the coverage problem, so these changes can be removed entirely.

dateutil/test/test_isoparser.py
```diff
@@ -399,7 +407,7 @@ def test_parse_isodate(d, dt_fmt, as_bytes):
     ('2014-04-19T', ValueError),                # Unknown components
 ])
 def test_isodate_raises(isostr, exception):
-    with pytest.raises(exception):
+    with pytest.raises(exception):              #pragma: no cover
```
pganssle: This should be covered, it is a normal test that is testing that these things raise errors.

dateutil/test/test_isoparser.py
```diff
@@ -453,11 +461,18 @@ def __make_time_examples():
 @pytest.mark.parametrize('time_val,time_fmt', __make_time_examples())
 @pytest.mark.parametrize('as_bytes', [True, False])
 def test_isotime(time_val, time_fmt, as_bytes):
-    tstr = time_val.strftime(time_fmt)
-    if isinstance(time_val, six.text_type) and as_bytes:
-        tstr = tstr.encode('ascii')
-    elif isinstance(time_val, bytes) and not as_bytes:
-        tstr = tstr.decode('ascii')
+    if six.PY2:
```
pganssle: As with the other changes, this is not the correct approach.

dateutil/test/test_isoparser.py
```diff
@@ -501,16 +516,16 @@ def test_isotime_midnight(isostr):
 ])
 def test_isotime_raises(isostr, exception):
     iparser = isoparser()
-    with pytest.raises(exception):
+    with pytest.raises(exception):              #pragma: no cover
```
pganssle: This is a normal test, please remove the `#pragma: nocover`

dateutil/test/test_isoparser.py
```diff
@@ -501,16 +516,16 @@ def test_isotime_midnight(isostr):
...
         iparser.parse_isotime(isostr)
 
 
-@pytest.mark.xfail()
+@pytest.mark.xfail()                            #pragma: no cover
 @pytest.mark.parametrize('isostr,exception', [
     ('14:3015', ValueError),                    # Inconsistent separator use
     ('201202', ValueError)                      # Invalid ISO format
 ])
 def test_isotime_raises_xfail(isostr, exception):
```
pganssle:
```suggestion
def test_isotime_raises_xfail(isostr, exception):    # pragma: nocover
```

This is the only line in this function that needs `# pragma; nocover`. The pragma goes after the function signature line on each of the `xfail` tests, and not in the body or on the decorator.

dateutil/test/test_isoparser.py
```diff
@@ -306,14 +306,14 @@ def test_isoparser_invalid_sep(sep):
 
 
 # This only fails on Python 3
-@pytest.mark.xfail(six.PY3, reason="Fails on Python 3 only")
-def test_isoparser_byte_sep():
-    dt = datetime(2017, 12, 6, 12, 30, 45)
-    dt_str = dt.isoformat(sep=str('T'))
+@pytest.mark.xfail(six.PY3, reason="Fails on Python 3 only")    #pragma: no cover
+def test_isoparser_byte_sep():                                  #pragma: no cover
```
pganssle: This whole function does not need `# pragma: nocover` because it is indeed already covered by the Python 2 tests.

dateutil/test/test_isoparser.py
```diff
@@ -286,15 +286,15 @@ def test_iso_with_sep_raises(sep_act, valid_sep, exception):
         parser.isoparse(isostr)
 
 
-@pytest.mark.xfail()
+@pytest.mark.xfail()                            #pragma: no cover
 @pytest.mark.parametrize('isostr,exception', [
     ('20120425T01:2000', ValueError),           # Inconsistent time separators
 ])
 def test_iso_raises_failing(isostr, exception):
```
pganssle:
```suggestion
def test_iso_raises_failing(isostr, exception):    # pragma: nocover
```

Only this line needs `# pragma: nocover`. As I mention below, you can mark an entire function as not covered by a single pragma on the `def` line.

dateutil/test/test_isoparser.py
pganssle: In this case, the problem is that in the original code, I was using:

```python
if isinstance(time_val, six.text_type) and as_bytes:
    ...
```
But what I *wanted* was:

```python
if isinstance(tstr, six.text_type) and as_bytes:
    ...
```
Because `time_val` is the time input, not the output of `time_val.strftime` (which will either be `bytes` or `str` depending on the Python version).

dateutil/test/test_isoparser.py
jbrockmendel: Can xfails be made strict? Makes sure we get alerted if something is accidentally fixed.

pganssle: They are already strict. It's configured globally in the pytest section of setup.cfg

PR #837, opened 2018-11-11
Update pyproject.toml for version 3.3

pganssle:
The CI is broken once again on Python 3.3, trying to fix this.

I believe this is a problem with `python_requires` when installing `setuptools` on Python 3.3.

This does not bode well for dropping support for python 2, though this may be just the fact that it's using a fairly old version of `pip`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,10 @@
 [build-system]
-requires = ["setuptools", "wheel", "setuptools_scm"]
+requires = [
+    "setuptools; python_version != '3.3'",
+    "setuptools<40.0; python_version == '3.3'",
+    "wheel",
+    "setuptools_scm"
+]
 
 [tool.towncrier]
     package = "dateutil"


jbrockmendel:
> This does not bode well for dropping support for python 2

There are still a non-trivial number of people using py27, but is the same true of py33?

pganssle:
> There are still a non-trivial number of people using py27, but is the same true of py33?

No, very few if any people use python 3.3. The reason I continue to keep it in the test suite is that I would like to be notified if anything we do actually breaks Python 3.3 support, because I want `python_requires` to reflect which versions of `python` the software *works with*, rather than the versions of `python` that are supported.

The reason I think it doesn't bode well for the transition from near-universal 2/3 support to many 3-only libraries is that things like `python_requires` are designed to make it so that dropping support for a version of Python easy and cheap (because doing so isn't supposed to break the users who haven't upgraded yet).

I see the comparatively minor shift of "dropping Python 3.3" as a test run for the big show, which is dropping Python 2 support, where there will be a *much* longer period of having large numbers of users stuck on Python 2. The fact that very few people are concerned with the fact that the things they are doing breaks Python 3.3 makes me think that the mechanisms we've put in place to make the 2->3 transition smooth are *not being adequately tested*, because no one is fixing them when they break under the reasoning that "very few people use Python 3.3" or "you shouldn't be supporting 3.3 beyond EoL anyway".

pganssle:
Oops, forgot a changelog entry.

PR #843, opened 2018-12-05
Ensure tests pass on Python 3 if long is redefined as int

cclauss:
## Summary of changes

When most projects add support for Python 3, they conditionally add the redefinition __long = int__ so instead of skipping these tests on Python 3, let's run them with that redefinition in place just to be sure.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -13,10 +13,15 @@ from dateutil.parser import UnknownTimezoneWarning
 
 from ._common import TZEnvContext
 
-from six import assertRaisesRegex, PY3
+from six import assertRaisesRegex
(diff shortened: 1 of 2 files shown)


dateutil/test/test_parser.py
```diff
@@ -303,11 +308,10 @@ def testDateCommandFormatReversed(self):
                                   tzinfo=self.brsttz))
 
     def testDateCommandFormatWithLong(self):
```
pganssle: I think maybe just add `@pytest.mark.skipif(six.PY3, reason="Long does not exist in Python 3")`

This test is specifically for support of Python 2 longs.

PR #850, opened 2018-12-31
Raise exception if not implemented comparison operators are used

jettify:
## Summary of changes
`relativedelta` has unexpected behavivior with `>`, `<`, `>=`, `<=` operators, it does not fail and sometimes shows correct result. Having `NotImplementedException` can save lot of debugging time. 

This is relevant for `python2`, `python3` is doing the right thing when those methods are not implemented.

Closes 

### Pull Request Checklist
- [x] Changes have tests


pganssle:
@jettify First off, thank you for taking the time to make a pull request.

In this particular case, I think it would be preferable to leave this as is. The standard way to implement these comparators is to `return NotImplemented` rather than `raise NotImplementedError` - the reason for this is that the standard comparison operators have some fallback logic, and someone could define a class like this:

```python
class AlwaysLessThan:
    def __lt__(self, other):
        return True
    def __gt__(self, other):
        return False
```

With the current implementation, `AlwaysLessThan() < relativedelta(years=1)` returns `False`. If this PR were merged, it would start raising an exception, because the comparison would never fall through to `AlwaysLessThan`'s implementation.

Yes, it is true that this means that Python 2 will be harder to debug, but this is one of the improvements to the language semantics that you get by upgrading to Python 3. If it were possible to improve the semantics in Python 2 *without* breaking Python 3 in the process, I'd be more receptive to it, but even then it may be a breaking change for Python 2 users who may be, for whatever reason, counting on this behavior. It's not clear that on balance it's worth it.

jettify:
I updated my implementation based on your feedback:
* new behavior only applied only in case of python2
* comparison throws `TypeError`, like in python3
* TypeError is raised only in case when we compare `relativedelta` to `relativedelta` otherwise default behaviour


Python3
```
In [3]: relativedelta(month=1) > relativedelta(month=1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-40d5d314bbca> in <module>
----> 1 relativedelta(month=1) > relativedelta(month=1)

TypeError: '>' not supported between instances of 'relativedelta' and 'relativedelta'
```

Python2
```
In [2]: relativedelta(month=1) > relativedelta(month=1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-40d5d314bbca> in <module>()
----> 1 relativedelta(month=1) > relativedelta(month=1)

/Users/nick/sources/dateutil/dateutil/relativedelta.pyc in __gt__(self, other)
    611             if isinstance(other, self.__class__):
    612                 name = self.__class__.__name__
--> 613                 raise TypeError(self._msg.format('>', name, name))
    614             return NotImplemented
    615

TypeError: '>' not supported between instances of 'relativedelta' and 'relativedelta'
```

jettify:
Anyway, I see your point and agree on python3 push, feel free to close this issue.

dateutil/relativedelta.py
```diff
@@ -592,6 +592,33 @@ def __repr__(self):
         return "{classname}({attrs})".format(classname=self.__class__.__name__,
                                              attrs=", ".join(l))
 
+    _msg = "'{}' not supported between instances of '{}' and '{}'"
```
pganssle: This can be under the `if PY2`, since it's not used in Python 3.

dateutil/relativedelta.py
```diff
@@ -592,6 +592,33 @@ def __repr__(self):
...
+
+    if PY2:
+        def __lt__(self, other):
```
pganssle: I wonder if it's worth refactoring this as:

```python
if PY2:
    def __mk_cmp(op):
        def not_implemented(self, other):
            if isinstance(other, self.__class__):
                msg = "{} not supported between instances of {name} and {name}"
                raise TypeError(msg.format(op, self.__class__.__name__)
            return NotImplemented
        return not_implemented

...

class relativedelta(...):
    ...
    if PY2:
        # Backport Python 3-like failure semantics into this class
        __lt__ = __mk_cmp('<')
        __le__ = __mk_cmp('<=')
        __gt__ = __mk_cmp('>')
        __ge__ = __mk_cmp('>=')
```

That will cut down on the copy-paste.

dateutil/test/test_relativedelta.py
```diff
@@ -303,6 +303,41 @@ def testComparison(self):
...
+            relativedelta(month=1) >= relativedelta(days=30)
+
+        with self.assertRaises(TypeError):
+            relativedelta(month=1) < relativedelta(days=30)
+
+        with self.assertRaises(TypeError):
+            relativedelta(month=1) <= relativedelta(days=30)
+
+    def textCustomComparison(self):
+        if PY2:
```
pganssle: Generally speaking, for tests that fail on Python 3 for compatibility reasons, we mark the test as skipped on Python 3 with the `skipif` marker, like so:

`@pytest.mark.skipif(six.PY3, reason="Whatever reason")`

But in this case, these tests should pass regardless, so you can just remove the `if PY2` block here.

dateutil/test/test_relativedelta.py
```diff
@@ -303,6 +303,41 @@ def testComparison(self):
...
+            class AlwaysLessThan:
```
pganssle: I believe you can achieve essentially the same thing using an existing object I created, [`ComparesEqual`](https://github.com/dateutil/dateutil/blob/master/dateutil/test/_common.py#L267), which compares equal to everything, so this test can be rewritten as:

```python
rd = relativedelta(month=1)
assert not rd < ComparesEqual
assert not rd > ComparesEqual
assert rd >= ComparesEqual
assert rd <= ComparesEqual
```

pganssle:
@jettify Thanks for your work on this issue. I've made a few comments on how the code can be cleaned up, but I think this is more or less the correct approach.

I am going to consider this a breaking change, so I've slated it for merging in the 3.x series, which will be the last "LTS" version of `dateutil`, intended to ease the transition to Python 3 by providing a consistent interface with the 4.x series (which will be Python 3-only), so that it is still possible to write Python 2/3 compatible code that uses dateutil 3/4.

Unfortunately, I am the blocker on a few things needed for the 3.x series (and even the next release in the 2.8 series), and it's quite delayed. I am hoping to get one or two more 2.x releases done, then start merging the breaking changes, so it will likely be a few months before this gets merged.

pganssle:
@jettify Please add yourself to the AUTHORS file (with a `**D**` after your name, to indicate that your contributions are dual-licensed Apache 2.0 and BSD), and if you can add a news fragment. Thanks!

AnneTheAgile:
Latest release is 2.8 so this ticket still needs to wait 
Release 2.8.0 · dateutil/dateutil · GitHub: https://github.com/dateutil/dateutil/releases/tag/2.8.0

jettify:
Do we still want to merge this? Since py2 is no longer supported.

PR #858, opened 2019-01-31
Fixed get_marker API change introduced by pytest version >= 4.0

### LINKED ISSUE #857: Tests are broken because a change of API in pytest
Since https://github.com/pytest-dev/pytest/pull/4564 `get_marker` does not exist anymore which renders the tests unusable when running them through `python -m pytest`.

nicoe:
## Summary of changes

I simply added a default value to the getattr call that is used only on python3.3 which should hopefully not be used anymore but in case it happens and the getattr fails (which it shouldn't) then the marker is skipped.

Closes #857 

### Pull Request Checklist

### DIFF
--- a/changelog.d/857.misc.rst
+++ b/changelog.d/857.misc.rst
@@ -1,0 +1,2 @@
+Fixed issue with get_marker API change introduced by pytest version >= 4.0
+Reported and fixed by @nicoe (gh issue #857, pr #858)
--- a/dateutil/test/conftest.py
+++ b/dateutil/test/conftest.py
@@ -6,9 +6,11 @@ import pytest
 # See: https://stackoverflow.com/a/53198349/467366
 def pytest_collection_modifyitems(items):
     for item in items:
+        marker_getter = getattr(item, 'get_closest_marker', None)
+
         # Python 3.3 support
-        marker_getter = getattr(item, 'get_closest_marker',
-                                getattr(item, 'get_marker'))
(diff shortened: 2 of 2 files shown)


pganssle:
@nicoe Thanks for the contribution.

I think I've lost track a bit about what the Python 3.3 compatibility is here. I can obviously see that it's broken in the build of PR #859, but I'm trying to unravel why so I can understand this change:

According to the latest documentation, [`get_closest_marker`](https://docs.pytest.org/en/latest/reference.html#_pytest.nodes.Node.get_closest_marker) should still be an attribute of `item` in this scenario, and in the old version (that only Python 3.3 supports) should have a `get_marker` method, so I would expect that on anything except Python 3.3, `marker_getter` would resolve to `item.get_closest_marker`, which does not  seem to be happening, so it's falling back to `item.get_marker`, which is failing - this is the correct behavior, because something about the logic has failed.

I don't think falling back to setting `marker = None` is the right behavior, because that's just disabling the feature where `xfail` tests don't contribute to the coverage - I want to retain that behavior.

pganssle:
Oh I see what the problem is. In my head I was thinking that the default argument to `getattr` wouldn't even evaluate if not necessary, but obviously that's not how function calls work in Python and `pytest>=4.0` is the first version where `get_marker` is actually *removed*, thus raising an error.

I think we can do this a slightly different way, I'll push an update to your branch.

PR #859, opened 2019-01-31
Add parameters parsing to EXDATE

### LINKED ISSUE #410: Add support of VALUE=DATE and TZID when parsing EXDATE rule

nicoe:
## Summary of changes
Quite simply add parameters parsing to EXDATE

Closes #410 

### Pull Request Checklist
- [X] Changes have tests

### DIFF
--- a/dateutil/test/test_rrule.py
+++ b/dateutil/test/test_rrule.py
@@ -2853,6 +2853,16 @@ class RRuleTest(WarningTestMixin, unittest.TestCase):
                           datetime(1997, 9, 9, 9, 0),
                           datetime(1997, 9, 16, 9, 0)])
 
+    def testStrSetExDateMultiple(self):
+        rrstr = ("DTSTART:19970902T090000\n"
+                 "RRULE:FREQ=YEARLY;COUNT=6;BYDAY=TU,TH\n"
+                 "EXDATE:19970904T090000,19970911T090000,19970918T090000\n")
+
+        rr = rrulestr(rrstr)
+        assert list(rr) == [datetime(1997, 9, 2, 9, 0),
+                            datetime(1997, 9, 9, 9, 0),
+                            datetime(1997, 9, 16, 9, 0)]
+
     def testStrSetExDateWithTZID(self):
         BXL = tz.gettz('Europe/Brussels')
         rr = rrulestr("DTSTART;TZID=Europe/Brussels:19970902T090000\n"
@@ -2986,6 +2996,11 @@ class RRuleTest(WarningTestMixin, unittest.TestCase):
 
         self.assertEqual(list(rr), [datetime(1997, 9, 2, 0, 0, 0),
                                     datetime(1998, 9, 2, 0, 0, 0)])
+
+    def testStrMultipleDTStartComma(self):
+        with pytest.raises(ValueError):
+            rr = rrulestr("DTSTART:19970101T000000,19970202T000000\n"
+                          "RRULE:FREQ=YEARLY;COUNT=1")
 
     def testStrInvalidUntil(self):
         with self.assertRaises(ValueError):
--- a/dateutil/test/test_rrule.py
+++ b/dateutil/test/test_rrule.py
@@ -2853,7 +2853,6 @@ class RRuleTest(WarningTestMixin, unittest.TestCase):
                           datetime(1997, 9, 9, 9, 0),
                           datetime(1997, 9, 16, 9, 0)])
 
-    @pytest.mark.xfail
     def testStrSetExDateWithTZID(self):
         BXL = tz.gettz('Europe/Brussels')
         rr = rrulestr("DTSTART;TZID=Europe/Brussels:19970902T090000\n"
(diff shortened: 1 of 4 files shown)


nicoe:
* Paul Ganssle [2019-02-02 16:21:15]:
> Thank you for working on this. This code is a bit complicated, and
> it seems like it's exactly duplicated or nearly so a little further
> on in the function. I think maybe it would be a good idea to factor
> out the common code into a `_parse_date_value` function or method
> and use it for both `DTSTART` and `EXDATE`. I'm assuming it will
> also be useful for `RDATE`.

OK I'll do it this way.
I also noticed that the code is exactly the same as the one below but
usually I make a function once I repeat the code 3 times and not twice
:).

-- 
(°> Nicolas Évrard
( ) Liège
 `¯

nicoe:
I factorized the code as said. It needs some more thoughts to include RDATE as its value can also be a `PERIOD`. I'll get back to this later.

pganssle:
OK, I've made the following changes:

1. Refactored `_parse_date_value` into a method
2. Started raising an error when multiple comma-separated values are passed to `DTSTART` (previously this was interpreting the second one as a time zone and ignoring it).
3. Added more tests to cover more of the use cases specified in issue #410.
4. Rewrote the git history a bit.

I think this is ready for merge. I'm hoping to get the 2.8.0 release wrapped up today, so I'm going to target merging this before the release.

@nicoe If you have time today, can you look over my changes and see if you object to anything?

CC: @mlorant

dateutil/test/test_rrule.py
```diff
@@ -2853,6 +2853,74 @@ def testStrSetExDate(self):
...
+        rr = rrulestr(rrstr)
+        assert list(rr) == [datetime(1997, 9, 4, 9), datetime(1997, 9, 11, 9)]
+
+    def testStrSetExDateValueDateTimeWithTZID(self):
+        BXL = tz.gettz('Europe/Brussels')
+        rrstr = '\n'.join([
+            "DTSTART;VALUE=DATE-TIME;TZID=Europe/Brussels:19970902T090000",
+            "RRULE:FREQ=YEARLY;COUNT=4;BYDAY=TU,TH",
+            "EXDATE;VALUE=DATE-TIME;TZID=Europe/Brussels:19970902T090000",
+            "EXDATE;VALUE=DATE-TIME;TZID=Europe/Brussels:19970909T090000",
```
nicoe: Would it be sensible to invert the order of VALUE and TZID for the second EXDATE?
It shouldn't change anything but it's an additional test somehow.

dateutil/test/test_rrule.py
```diff
@@ -2853,6 +2853,74 @@ def testStrSetExDate(self):
...
+        rr = rrulestr(rrstr)
+        assert list(rr) == [datetime(1997, 9, 4, 9, tzinfo=BXL),
+                            datetime(1997, 9, 11, 9, tzinfo=BXL)]
+
+    def testStrSetExDateValueDate(self):
+        rrstr = '\n'.join([
+            "DTSTART;VALUE=DATE:19970902",
+            "RRULE:FREQ=YEARLY;COUNT=4;BYDAY=TU,TH",
+            "EXDATE;VALUE=DATE:19970902",
+            "EXDATE;VALUE=DATE:19970909",
```
nicoe: This test makes me realize that the code handling TZID shouldn't be triggered when it's a VALUE=DATE, should it?

pganssle: I'm not actually sure. It's certainly the case that the `datetime.date` has no way to represent a time zone, but what `dateutil.rrule` does is to parse `DATE` values as "datetime at midnight", which means they *can* have a time zone.

It depends on what the spec says, I suppose. [The grammar for EXDATE](https://tools.ietf.org/html/rfc5545#section-3.8.5.1) seems to indicate it's syntactically valid to have a `TZID` for `VALUE=DATE`, but maybe it's forbidden at a semantic level elsewhere in the spec.

PR #880, opened 2019-02-23
Convert setup.py to declarative build file when possible

### LINKED ISSUE #871: Move all simple constants from setup.py into setup.cfg
Nearly all the options in [our `setup.py`](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) can be moved over to `setup.cfg`.

The only things that need to stay are:

1. `use_scm_version` (I believe there is no declarative option for this)
2. The README configuration, which involves modifying the README
3. `cmdclass`, which I think has no declarative alternative

The trickiest part will be converting the `find_packages` option over, but that is covered in the setuptools documentation I linked to, you specify it as `packages = find:`, and then pass the options in `[options.packages.find]`.

For bonus points, we may be able to switch over the `README` thing by switching `README` over to using `.. code-block :: python3` directly, and then doing a search-and-replace as part of the [docs build](https://github.com/dateutil/dateutil/blob/master/docs/conf.py) instead. This will be a bit tricky, though, because `README.rst` is included in the docs with an RST include, which means we would have to create a temporary file that would be included. For now, this ticket can be closed if we do everything *except* that.

FakeNameSE:
## Summary of changes
Moved simple constants from setup.py to setup.cfg.

Closes #871

### Pull Request Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,7 +2,46 @@
 universal = 1
 
 [metadata]
+name = python-dateutil
+description = Extensions to the standard Python datetime module
+author = Gustavo Niemeyer
+author_email = gustavo@niemeyer.net
+maintainer = Paul Ganssle
+maintainer_email = dateutil@python.org
+url = https://dateutil.readthedocs.io
+long_description_content_type = text/x-rst
+license = Dual License
 license_file = LICENSE
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    License :: OSI Approved :: BSD License
+    License :: OSI Approved :: Apache Software License
+    Programming Language :: Python
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: Software Development :: Libraries
+
+[options]
+zip_safe = True
+setup_requires = setuptools_scm
+install_requires = six >= 1.5
+python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -2,7 +2,47 @@
...
 
 [metadata]
+name = python-dateutil
+description = Extensions to the standard Python datetime module
+author = Gustavo Niemeyer
+author_email = gustavo@niemeyer.net
+maintainer = Paul Ganssle
+maintainer_email = dateutil@python.org
+url = https://dateutil.readthedocs.io
+long_description = file: README.rst
```
pganssle: You should revert this change, because we're modifying the `README` variable in `setup.cfg`.

FakeNameSE: Thank you, problem solved!

kapilt:
this has caused a fairly large problem for downstream users, due to the need to upgrade setup tools for this inconsequential change, which has caused some downstream dependencies to freeze, resulting in a large amount of churn for thousands of downstream packages and developers as their forced to pin to 2.8.0. a large portion of that is due to boto3's own issues with its installers, but packaging decisions that cause deprecation of extant releases really need careful consideration in libraries. given this is effectively a non functional change, is there any interest in a pr that would revert it, its unclear what value the change was delivering, especially measured by its negative impact to users.

pganssle:
@kapilt No thank you. `dateutil` ships a universal wheel, so it shouldn't really matter *what* version of `setuptools` you have, and if you insist on building `dateutil` from scratch, you should use a recent version.

If for some reason you need to build `dateutil` from source (which should be rare), the version of `setuiptools` required to build `dateutil` is documented in the [`pyproject.toml`](https://github.com/dateutil/dateutil/blob/master/pyproject.toml), and the current version of `dateutil` is intended to be built with a PEP 517-compatible front-end. Deviations from this are basically unsupported configurations.

I would also note that it is dangerous to use an old version of `setuptools`, as you are likely to generate bad metadata, or cause problems with your install environment.

PR #881, opened 2019-02-25
Use custom ParserError class in parser exceptions

gfyoung:
If multiple arguments are passed in, it raises a tuple, not one string.

xref https://github.com/pandas-dev/pandas/pull/25430#discussion_r259614777

### DIFF
--- a/changelog.d/881.bugfix.rst
+++ b/changelog.d/881.bugfix.rst
@@ -1,0 +1,3 @@
+Parsing errors will now raise ``ParserError``, a subclass of ``ValueError``,
+which has a nicer string representation.
+Patch by @gfyoung (gh pr #881)
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -651,7 +651,10 @@ class parser(object):
         if len(res) == 0:
             raise ParserError("String does not contain a date: %s", timestr)
 
-        ret = self._build_naive(res, default)
+        try:
+            ret = self._build_naive(res, default)
+        except ValueError as e:
+            six.raise_from(ParserError(e.args[0] + ": %s", timestr), e)
 
         if not ignoretz:
             ret = self._build_tzaware(ret, res, tzinfos)
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -49,7 +49,7 @@ from warnings import warn
 from .. import relativedelta
 from .. import tz
 
-__all__ = ["parse", "parserinfo"]
+__all__ = ["parse", "parserinfo", "ParserError"]
 
 
 # TODO: pandas.core.tools.datetimes imports this explicitly.  Might be worth
@@ -626,7 +626,7 @@ class parser(object):
             first element being a :class:`datetime.datetime` object, the second
             a tuple containing the fuzzy tokens.
 
-        :raises ValueError:
+        :raises ParserError:
             Raised for invalid or unknown string format, if the provided
             :class:`tzinfo` is not in a valid format, or if an invalid date
             would be created.
@@ -646,10 +646,10 @@ class parser(object):
         res, skipped_tokens = self._parse(timestr, **kwargs)
 
         if res is None:
-            raise ValueError("Unknown string format:", timestr)
+            raise ParserError("Unknown string format: %s", timestr)
 
         if len(res) == 0:
-            raise ValueError("String does not contain a date:", timestr)
+            raise ParserError("String does not contain a date: %s", timestr)
 
         ret = self._build_naive(res, default)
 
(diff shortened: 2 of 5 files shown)


pganssle:
There was a similar issue in `setuptools` recently and I realized that someone might be actually using the `args` parameter here to more cleanly pull out the `timestr`.

@jbrockmendel Do you remember what happened with the plan to switch these `ValueError`s over to a custom Parser error class? I'm thinking maybe a better solution would be something like this:

```python
class ParserError(ValueError):
    def __repr__(self):
        if len(self.args) > 1:
            return self.__class__.name + ": " + (self.args[0] % self.args[1:])
        else:
            return super(ParserError, self).__repr__()
```

Then we can have our cake and eat it, too.

jbrockmendel:
> Do you remember what happened with the plan to switch these `ValueError`s over to a custom Parser error class?

Nothing comes to mind.  I've been going through older issues/PRs, will mention if I come across it

gfyoung:
@pganssle @jbrockmendel : Friendly ping here.  Any more thoughts on this one?

jbrockmendel:
LGTM.  @pganssle go for it?

pganssle:
@jbrockmendel I keep going back and forth on this because I do think it might be useful for someone to be able to extract the `timestr` simply with `.args[1]`, and someone may be counting on that.

I'm leaning more towards the solution of switching these from `ValueError` to `ParserError`, where `ParserError` is defined [as above](https://github.com/dateutil/dateutil/pull/881#issuecomment-467473945). Not sure if it should be `__repr__` or `__str__`, though.

gfyoung:
@pganssle : Should the error be defined in the same file, or is there a better place for it?

pganssle:
@gfyoung Well, the thing is I generally don't want to make any guarantees about the contents of an error message, which is why I don't generally test with any sort of regex match on the errors. The reason is that I don't consider that format part of the public interface.

I *would* be more comfortable with the idea that `ValueError(msg, timestr)` *could* have the "timestr" part of the exception as something that *is* part of the interface, because it's something that is cleanly accessible by some code that attempts to recover from the error. I'm not sure I was *deliberately* trying to make that happen, but the original contributor may have been and people may be relying on it now.

Given that there *is* a way to both make the error look nice when printed *and* make `timestr` easily accessible without parsing the error message, I'm inclined to do so rather than risk having to quibble about what is and is not considered a backwards-incompatible change.

pganssle:
@gfyoung Ah, missed the question, I think same file is best, and you can add it to `__all__`.

gfyoung:
@pganssle : Have a look.  I had to make some adjustments to your suggested `ParserError` implementation because it wasn't rendering properly when put `raise` in front of it.  However, it should be good now.

PTAL.

gfyoung:
@pganssle @jbrockmendel : Friendly ping here. Any more thoughts on this one?

gfyoung:
@pganssle @jbrockmendel : Another friendly ping here. Any more thoughts on this one?

jbrockmendel:
@pganssle call the ball

dateutil/parser/_parser.py
```diff
@@ -49,7 +49,17 @@
...
+
+
+class ParserError(ValueError):
+    def __str__(self):
+        try:
+            return self.args[0] % self.args[1:]
+        except (TypeError, IndexError):
+            return super(ParserError, self).__str__()
+
+    __repr__ = __str__
```
pganssle: This is not quite appropriate, because `str` returns only the message portion of the error and does not include the class:

```python
>>> str(ValueError("Test", "test"))
"('Test', 'test')"
>>> repr(ValueError("Test", "test"))
"ValueError('Test', 'test')"
```

So this should be maybe:

```python
def __repr__(self):
    return "%s(%s)" % (self.__class__.__name__, str(self))
```

changelog.d/881.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed a bug in errors raised in ``_parser.py`` in which the ``ValueError``
```
pganssle: This should not mention `_parser.py`, which is a private module, and should be updated to reflect the fact that parsing errors will now raise `ParserError`, which is a subclass of `ValueError`, and which has a cleaner string representation when passed a `msg`, `timestr` tuple.

gfyoung:
> I'm very sorry that I've been remiss about merging this.

No worries.  I'm no stranger to the time crunch that maintainers can have 🙂 

> I have noted two minor changes that need to be made, but I have done this primarily for my own notes...I can update and merge this tomorrow or the next day.

Sounds good.  I'll ping on Tuesday if necessary.  Thanks!

pganssle:
Hm.. Build failure looks unrelated. We can merge this one after the tz master build is fixed.

PR #882, opened 2019-02-25
REF: collect parser methods by function

jbrockmendel:
The methods within `parser` are a bit of a jumble.  This separates out two sections for methods that a) act on individual tokens and b) are for creating the datetime output after the string is fully parsed.

No hurry on this if it would cause rebasing conflicts elsewhere.

Methods are unchanged, with the exception of a long line in a comment in to_decimal that is wrapped.

### DIFF
--- a/changelog.d/735.misc.rst
+++ b/changelog.d/735.misc.rst
@@ -1,0 +1,2 @@
+Converted many tests in ``test_parser`` over to using ``pytest`` style tests.
+Patch by @jbrockmendel (gh pr #735)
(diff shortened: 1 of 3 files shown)


jbrockmendel:
failed build is pypy, looks unrelated

jbrockmendel:
> Can you add a `misc` item to the changelog?

Done.  Is there a better word than "Fixed" in this context?

pganssle:
@jbrockmendel I often go with "Patch by ..." when contributing to CPython, but no need to worry. For the moment, these MISC things don't actually generate entries directly in the changelog (though I'm thinking maybe that should change).

PR #883, opened 2019-02-25
implement next_jump, prev_jump to fix first-century cases

jbrockmendel:
Fixes (with tests) currently-failing cases "0031 Jun 11" and "0031Jun11".

`next_jump` and `prev_jump` are from a fork I made ages ago to handle some corner cases.  If accepted, they can be used to clean up a bunch of existing checks in a follow-up.


dateutil/parser/_parser.py
```diff
@@ -571,6 +571,74 @@ def resolve_ymd(self, yearfirst, dayfirst):
         return year, month, day
 
 
+def next_jump(tokens, idx, skip_comma=True):
```
pganssle: I think this should be a method somewhere, no?

I definitely think it shouldn't be a public, top-level function yet. Can you move it onto either `parser` or `parserinfo` (whichever you think is best), and start it out as private (`_next_jump`)?

dateutil/test/test_parser.py
```diff
@@ -675,6 +675,17 @@ def test_era_trailing_year(self):
         res = parse(dstr)
         assert res.year == 2001, res
 
+    def test_first_century(self):
```
pganssle: Can you make this a parametrized test? I prefer "one assert per test" wherever possible.

dateutil/parser/_parser.py
```diff
@@ -571,6 +571,74 @@ def resolve_ymd(self, yearfirst, dayfirst):
...
+    """
+    Get the next sibling and its index, skipping over whitespace
+    and (if specified) commas.
+
+    Parameters
```
pganssle: This is inconsistent with the rest of the documentation, where we're using the standard sphinx style. I'm not opposed to the idea of converting over the documentation to using this numpy-style, but I don't know if I want it to be a mishmash, which confuses new contributors.

dateutil/parser/_parser.py
```diff
@@ -571,6 +571,74 @@ def resolve_ymd(self, yearfirst, dayfirst):
...
+              tokens[idx + 2].isspace()):
+            # e.g. if tokens[idx] is the "30" in "Dec 30, 2016" we want to
+            #  return the "2016"
+            sib_idx = idx + 3
+            sib = tokens[sib_idx]
+
+    return sib, sib_idx
+
+
+def prev_jump(tokens, idx, skip_comma=True):
```
pganssle: It doesn't seem like you're actually using this function anywhere, other than in the function itself. Is it necessary to keep it?

dateutil/parser/_parser.py
```diff
@@ -571,6 +571,74 @@ def resolve_ymd(self, yearfirst, dayfirst):
         return year, month, day
 
 
+def next_jump(tokens, idx, skip_comma=True):
+    """
+    Get the next sibling and its index, skipping over whitespace
```
pganssle: An example of what this is looking for would probably be useful to future software archaeologists. I think you have one somewhere in the comments of the function itself.

dateutil/parser/_parser.py
```diff
@@ -981,6 +1049,15 @@ def _parse_numeric_token(self, tokens, idx, info, ymd, res, fuzzy):
...
             idx += 1
 
+        elif (not ymd.has_year and len_li == 4 and
+              info.month(next_jump(tokens, idx)[0] or '')):
+            # e.g. this is the "0003" in "0003 Nov 06"
+            # unambiguous year, use value_repr instead of value to avoid
+            #  dropping leading zeros, e.g. "0031" represents 31 AD,
+            #  not 2031
+            # Note: this check must come before the check for info.jump below
+            ymd.append(value_repr, 'Y')
```
pganssle: I'm mildly surprised that we can put strings in the `ymd` tuple like this without it breaking anything.

dateutil/test/test_parser.py
jbrockmendel: Will do.

dateutil/parser/_parser.py
jbrockmendel: That's fair, will try to convert (I'm much less familiar with the sphinx style, so will bear review).  Other than "I like it better" Im not sure if there is a compelling reason to convert to the numpy-style.  Regardless, that would be its own Issue.

dateutil/parser/_parser.py
jbrockmendel: Good idea, will do.

dateutil/parser/_parser.py
jbrockmendel: _ymd.append starts with a check for `hasattr(val, '__len__')` which I guess is intended to catch strings (not sure why it doesnt just check for strings directly...)

dateutil/parser/_parser.py
jbrockmendel: We don't need it at the moment, but I thought I'd port it since it goes with next_jump.  I'll remove it now and can re-introduce it if/when it becomes necessary.

dateutil/parser/_parser.py
jbrockmendel: > I think this should be a method somewhere, no? [...] Can you move it onto either `parser` or `parserinfo`

Downstream I actually have these as methods in a `TokenStream` class (subclasses `list`) and `tokens = TokenStream(timestr)` takes the place of `l = _timelex(timestr)`.  In #884 I'll describe some of the other virtues of the `TokenStream` class.  If this is made into a method, I think `TokenStream` is the right abstraction for where it belongs.

For the time being I've privatized the function.

PR #890, opened 2019-02-26
Remove duplicate tests

jbrockmendel:
Add a check so that duplicates don't sneak back in.

Rename TestInputFormats to the more accurate TestInputTypes.

A couple more test-cleanup PRs coming up (one with an actual bugfix), trying to keep them small and orthogonal.


dateutil/test/test_parser.py
```diff
@@ -180,7 +182,7 @@ def test_ybd(self):
             assert res == actual
 
 
-class TestInputFormats(object):
+class TestInputTypes(object):
```
pganssle: Why `Types` instead of `Formats`?

jbrockmendel: Because whats being tested here is different types: bytes vs unicode vs bytearray vs StringIO...

pganssle: Ah, I was a bit thrown by the fact that the first test is just an empty string. I see now.

jbrockmendel:
Looks like no rebase needed.

pganssle:
@jbrockmendel Can you add a changelog?

jbrockmendel:
done

PR #891, opened 2019-02-26
Add+collect tests for tzinfos input types, fix missed case of invalid…

jbrockmendel:
Implement tests for types of inputs that can be passed as `tzinfos` (two of these are moved from further down in the file, the rest ported from downstream).  Fixes the case where something bizarre is passed like `tzinfos={"UTC": something_totally_invalid}` which at the moment raises `UnboundLocalError` instead of `TypeError`

### DIFF
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -360,6 +360,14 @@ class TestTzinfoInputTypes(object):
         expected = datetime(2017, 2, 3, 12, 40)
         assert result == expected
 
+    @pytest.mark.xfail(reason="Incorrect error raised with bad tzinfo")
+    def test_invalid_tzinfo_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+        # Pass an absurd tzinfos object
+        tzinfos = {"UTC": ValueError}
+        with pytest.raises(TypeError):
+            parse(dstr, tzinfos=tzinfos)
+
     def test_valid_tzinfo_tzinfo_input(self):
         dstr = "2014 January 19 09:00 UTC"
         tzinfos = {"UTC": tz.UTC}
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -360,7 +360,6 @@ class TestTzinfoInputTypes(object):
         expected = datetime(2017, 2, 3, 12, 40)
         assert result == expected
 
-    @pytest.mark.xfail(reason="Incorrect error raised with bad tzinfo")
     def test_invalid_tzinfo_input(self):
         dstr = "2014 January 19 09:00 UTC"
         # Pass an absurd tzinfos object
--- a/dateutil/test/test_parser.py
+++ b/dateutil/test/test_parser.py
@@ -347,6 +347,46 @@ class TestInputTypes(object):
         assert res == expected
 
 
+class TestTzinfoInputTypes(object):
+    def test_tzinfo_dict_could_return_none(self):
+        dstr = "2017-02-03 12:40 BRST"
+        result = parse(dstr, tzinfos={"BRST": None})
+        expected = datetime(2017, 2, 3, 12, 40)
+        assert result == expected
+
+    def test_tzinfos_callable_could_return_none(self):
+        dstr = "2017-02-03 12:40 BRST"
+        result = parse(dstr, tzinfos=lambda *args: None)
+        expected = datetime(2017, 2, 3, 12, 40)
+        assert result == expected
+
+    def test_valid_tzinfo_tzinfo_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+        tzinfos = {"UTC": tz.UTC}
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.UTC)
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
+        assert res.tzinfo is expected.tzinfo
+
+    def test_valid_tzinfo_unicode_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+        tzinfos = {u"UTC": u"UTC+0"}
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.UTC)
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
+
+    def test_valid_tzinfo_callable_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+
+        def tzinfos(*args, **kwargs):
+            return u"UTC+0"
+
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.tzstr("UTC+0"))
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
+        assert res.tzinfo is expected.tzinfo
+
+
 class ParserTest(unittest.TestCase):
 
     @classmethod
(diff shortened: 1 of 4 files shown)


dateutil/test/test_parser.py
```diff
@@ -238,6 +238,60 @@ def test_parse_bytearray(self):
...
+        dstr = "2014 January 19 09:00 UTC"
+
+        def tzinfos(*args, **kwargs):
+            return u"UTC+0"
+
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.tzutc())
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
+
+    @pytest.mark.xfail(reason="Doesn't check for UTC in tzinfos dict, "
```
pganssle: This is a weird test but I think you're right that it should be possible.

I think a more plausible use of this functionality would be something like returning a custom UTC type, e.g. `datetime.timezone.utc`, but for now let's leave it as is.

dateutil/test/test_parser.py
```diff
@@ -238,6 +238,60 @@ def test_parse_bytearray(self):
...
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
+
+    def test_valid_tzinfo_callable_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+
+        def tzinfos(*args, **kwargs):
+            return u"UTC+0"
+
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.tzutc())
```
pganssle: This is a confusing test, because what's actually happening here is that you are getting back `tzstr("UTC+0")` as the `tzinfo`, but that compares equal to `tz.UTC` because it's an inter-zone comparison and both resolve to the same value. 

Can you change the test to use: `tz.tzstr("UTC+0")` in the "expected" part of the test?

dateutil/test/test_parser.py
```diff
@@ -238,6 +238,60 @@ def test_parse_bytearray(self):
...
+    def test_invalid_tzinfo_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+        # Pass an absurd tzinfos object
+        tzinfos = {"UTC": ValueError}
+        with pytest.raises(TypeError):
+            parse(dstr, tzinfos=tzinfos)
+
+    def test_valid_tzinfo_tzinfo_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+        tzinfos = {"UTC": tz.tzutc()}
```
pganssle: I prefer to use `tz.UTC` over `tz.tzutc()` as a general rule in `dateutil` itself. I realize that in `pandas` there are backwards-compatibility problems with that, but in `dateutil`'s test suite that's not the case ;)

dateutil/test/test_parser.py
```diff
@@ -238,6 +238,60 @@ def test_parse_bytearray(self):
...
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.tzutc())
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
+
+    def test_valid_tzinfo_unicode_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+        tzinfos = {u"UTC": u"UTC+0"}
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.tzutc())
```
pganssle: Same comment as below.

Also, I realize that these tests would still succeed regardless of whether or not we're even *using* the `tzinfos` parameter. That makes me think that we should either be using `res.tzinfo is expected.tzinfo` or we should be using some none-UTC value here, or both.

One option is to have two tests - one for "UTC" as the offset specifically, that tests `is`, and one for something more general like `"LXQ"` where we are unlikely to accidentally hit the correct answer via another code path.

dateutil/test/test_parser.py
jbrockmendel: Changed.

When did tz.UTC become available?  I'll open an issue in pandas to update the min dependencies; tz.UTC will be non-trivially more performant in the cython code.

pganssle: Looks like [2.7.0](https://github.com/dateutil/dateutil/blob/master/NEWS#L266).

dateutil/test/test_parser.py
```diff
@@ -240,6 +240,60 @@ def test_parse_bytearray(self):
...
+
+    def test_valid_tzinfo_callable_input(self):
+        dstr = "2014 January 19 09:00 UTC"
+
+        def tzinfos(*args, **kwargs):
+            return u"UTC+0"
+
+        expected = datetime(2014, 1, 19, 9, tzinfo=tz.tzstr("UTC+0"))
+        res = parse(dstr, tzinfos=tzinfos)
+        assert res == expected
```
pganssle: I think we should also add `res.tzinfo is expected.tzinfo`, to ensure that the `tzinfos` function actually worked correctly.

pganssle: Same with a few of the other tests, I think. Even if you don't specify `tzinfos` all of these tests will still pass.

jbrockmendel:
I'm travelling the next couple days, so may not get around to this until Wednesday.  In the interim, can we merge the other two?

pganssle:
@jbrockmendel I merged #892, but I think #894 is still blocked on changes from you. If you want I can make the changes myself and merge that one.

jbrockmendel:
I think the requested change here is just adding `assert res.tzinfo is expected.tzinfo` where appropriate.  I'm skipping that for the case where both are `None`.  That leaves `test_valid_tzinfo_tzinfo_input`, `test_valid_tzinfo_callable_input` where that passes and `test_valid_tzinfo_unicode_input` where that assertion fails since that assertion amounts to `tz.tzstr('UTC+0') is tz.UTC`.  I'm leaving that assertion out in the newest commit.

jbrockmendel:
@pganssle gentle ping.  I think this is the last of my test-refactoring PRs.

Also definitely-ready is gfyoung's #881.  I definitely don't want to merge any of my own PRs, but would be comfortable taking on merge responsibilities for things like that.

pganssle:
@jbrockmendel Sorry about all the delays. I'll review again today or tomorrow. Assuming any changes I might want are sufficiently simple, I'll just push them directly to your branch and merge rather than bother you with another round of addressing review comments.

pganssle:
@jbrockmendel I've rebased this and unified the "check if equal + check if `tzinfo` is identical" logic into a common function, but looking at it, I think I misunderstood the `xfail` test:

```python
    def test_valid_tzinfo_int_input(self):
        dstr = "2014 January 19 09:00 UTC"
        tzinfos = {u"UTC": -28800}
        expected = datetime(2014, 1, 19, 9, tzinfo=tz.UTC)
        res = parse(dstr, tzinfos=tzinfos)
        self.assert_equal_same_tz(res, expected)
```

The current behavior is that `tz.tzoffset("UTC", -28800)` gets attached to the result, which seems like the right thing to do, since specifying `UTC` in `tzinfos` is more explicit than an automatic mapping of `"UTC"` to `tz.UTC`.

I'll note that the fact that `tzinfos` can override what `"UTC"` gets parsed to *also* enables you to do things like `tzinfos={"UTC": datetime.timezone.utc}`, which could be very useful for people who want to use the standard library's objects.

jbrockmendel:
Yes, I agree that if the user explicitly specifies a non-standard UTC, that should take precedence.

pganssle:
Thanks @jbrockmendel for the patience. Now to continue working through the backlog on all my projects...

PR #892, opened 2019-02-27
Add several parser test cases

jbrockmendel:


dateutil/test/test_parser.py
```diff
@@ -675,6 +675,41 @@ def test_era_trailing_year(self):
...
+
+    def test_no_year_zero(self):
+        with pytest.raises(ValueError):
+            parse("0000 Jun 20")
+
+    def test_out_of_bound_day(self):
+        with pytest.raises(ValueError):
+            parse("Feb 30, 2007")
+
+    @pytest.mark.parametrize('fuzzy', [True, False])
```
pganssle: I think this is probably fine, but a parametrized fixture may be able to do this with less repetition? I am by no means an expert, but one option is:

```python
@pytest.fixture(scope="class", params=[True, False])
def fuzzy(self, arg):
    return arg.param
```

Then you can just have "fuzzy" be an argument in all the other tests.

pganssle:
@jbrockmendel Great tests, thanks for this. I have the one comment that you can take or not take about the parametrized fixture, and this also needs a changelog. I'm about to merge #890 so you may want to rebase if you're making the changes after that.

dateutil/test/test_parser.py
jbrockmendel: Sure.  Easier to just put it at the module level and not need the `scope="class"` or `self`

jbrockmendel:
Rebased and edited with requested changes.

jbrockmendel:
@pganssle gentle ping for this, #891, and #894 (and if were really ambitious, #883 and #737).  I'm a dateutil-productivity hot-streak and want to get the most out of it before inevitably getting distracted.

Issue #724, opened 2018-05-21
Tests failing on Python 3.3


pganssle:
The tests seem to be failing on Python 3.3 due to a wheel incompatibility.

My best guess for what is happening is that `wheel` dropped support for Python 3.3 and something in the testing pipeline (only seems to be happening on Travis as far as I can tell) is using an old version of `pip`.

My suspicion is that Travis has some code that creates a `virtualenv`, and it does so with an un-upgraded version of `pip`, which breaks on Python 3.3. It might be best to kick this over to Travis's github issues and see if we can get it fixed there. Otherwise we'll probably have to modify the way Travis creates its virtualenvs.

In the end we'll just drop Python 3.3 support soon enough, but I'd like to make sure I can do this in a pinch if need be and I don't like the idea of being forced to drop Python 3.3 by a misconfigured CI. Generally I don't drop support for a version of Python unless it's burdensome to continue supporting it in the code itself (i.e. a descriptive approach to version support). That hasn't happened for Python 3.3 yet.

pganssle:
It seems this is a tox-specific problem. I have a MWE repo [here](https://github.com/pganssle/travis-test) with a failing Travis run [here](https://travis-ci.org/pganssle/travis-test). I think it's the part where `tox` creates the virtualenv that needs to have `wheel` pinned.

hugovk:
By the way, to get some idea of usage, here's the pip installs for python-dateutil from PyPI for May 2018:


| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  76.50% |      9,055,420 |
| 3.6            |  15.65% |      1,853,163 |
| 3.5            |   4.40% |        520,941 |
| 3.4            |   2.62% |        309,889 |
| 2.6            |   0.77% |         91,410 |
| 3.7            |   0.04% |          4,983 |
| 3.3            |   0.01% |          1,504 |
| 3.8            |   0.00% |            101 |
| 3.2            |   0.00% |             78 |
| None           |   0.00% |             32 |
| Total          |         |     11,837,521 |

Source: `pypinfo --start-date 2018-05-01 --end-date 2018-05-31 --percent --markdown python-dateutil pyversion`

PR #897, opened 2019-03-06
Drop support for Python 3.3

michael-k:
## Summary of changes

Python 3.3 reached its EOL in September 2017 and now even Python 3.4
will reach its EOL in a few days.

See https://github.com/dateutil/dateutil/issues/724#issuecomment-396908268 for “the pip installs for python-dateutil from PyPI for May 2018”.

Python 3.3 EOL: https://www.python.org/dev/peps/pep-0398/#x-end-of-life
Python 3.4 EOL: https://www.python.org/dev/peps/pep-0429/#release-schedule

### Pull Request Checklist
- [ ] ~~Changes have tests~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: python
 cache: pip
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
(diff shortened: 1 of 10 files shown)


pganssle:
See #611. For the moment the support policy is *descriptive* rather than prescriptive. As long as `dateutil` works on Python 3.3, I want to test it, so that I can be sure that `python_requires` remains accurate.

If we ever want to use a feature that is not available in Python 3.3 (or any other EOL version), but *is* available in all other supported versions, we can "drop support" for Python 3.3, but as the library still works correctly on Python 3.3, the metadata should accurately reflect that.

PR #911, opened 2019-04-25
Unify PyPy into normal flow

vtbassmatt:
## Summary of changes

Hi there, PM from Azure Pipelines here. We added `pypy2` and `pypy3` as Python "versions" for the `UsePythonVersion` task. That means you no longer have to special-case PyPy in your pipelines.


vtbassmatt:
Yes, those are valid on Windows too.

PR #912, opened 2019-04-28
Fix tzwinlocal() with non-ASCII timezone names

### LINKED ISSUE #210: tzwinlocal() does not work with non-ASCII characters in timezone name
First reported as a comment on #197 by @labrys, this is actually unrelated to that issue, so it's getting its own issue:

> since updating to dateutil 2.5.0 in sickrage, windows users are getting this (possibly related) timezone error.  It seems to be a unicode related bug as its causing:
> ```python
>     sb_timezone = tz.tzwinlocal() if tz.tzwinlocal else tz.tzlocal()
>   File "C:\SickRage\SickRage\lib\dateutil\tz\win.py", line 239, in __init__
>     handle, "%s\%s" % (TZKEYNAME, self._stdname)) as tzkey:
> UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 57: ordinal not in range(128)"
> ```
> The limited tests we've tried so far do not cause this error with `tz.tzlocal()`.  Changing timezones to e.g EST or to some other european timezones do not cause the error with `tz.tzwinlocal()`.
> This was confirmed using python 2.7.10 (portable) on Server 2012 R2 and Windows 7 ultimate.

I was able to confirm this issue myself. I tried a first pass mitigation in PR #208 (which fixed things on our end), but it turns out that this is a bug in Python 2.x where the `_winreg` module does not properly handle unicode.

Ideally this would be handled upstream, but we may be able to come up with a mitigation by retrieving the key another way. I'm not 100% sure, but given that there are apparently [canonical time zone keys](http://stackoverflow.com/a/32974430/467366) in English on Windows, maybe we can try and retrieve the time zone that way on Python 2.7.

Alternately, we can maybe use an approach similar to what [`winreg_unicode`](https://github.com/DanielStutzbach/winreg_unicode) is doing (I'd rather not add winreg_unicode as a dependency if it can be helped).

dagwieers:
## Summary of changes
This is the fix as suggested in https://github.com/dateutil/dateutil/issues/210#issuecomment-320348647

Closes #210 (at least for Windows 7 and newer)

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/dateutil/tz/win.py
+++ b/dateutil/tz/win.py
@@ -282,8 +282,14 @@ class tzwinlocal(tzwinbase):
             self._dst_abbr = keydict["DaylightName"]
 
             try:
-                tzkeyname = text_type('{kn}\\{sn}').format(kn=TZKEYNAME,
-                                                          sn=self._std_abbr)
+                try:
+                    # NOTE: This works from Windows 7 onwards
+                    tzkeyname = text_type('{kn}\\{sn}').format(kn=TZKEYNAME,
+                                                               sn=keydict["TimeZoneKeyName"])
+                except KeyError:
+                    # NOTE: Backward compatibility for pre Windows 7 (but suffers from #210)
+                    tzkeyname = text_type('{kn}\\{sn}').format(kn=TZKEYNAME,
+                                                               sn=self._std_abbr)
                 with winreg.OpenKey(handle, tzkeyname) as tzkey:
                     _keydict = valuestodict(tzkey)
                     self._display = _keydict["Display"]


pganssle:
If this works, I think it's better than nothing that it only fixes the problem on Win 7+.

Two things, one easy and one hard:

1. Can you add a comment that we can remove this hack when we drop support for Python 2.7?
2. Can you think of any way to test this? I have trouble even replicating the issue, so it's hard for me to tell if this has worked without a test (or to prevent regressions).

I think we can use [`TZWinContext`](https://github.com/dateutil/dateutil/blob/master/dateutil/test/_common.py#L178), but I think we may also have to modify the locale for the test.

dagwieers:
@pganssle I don't know how you are currently testing for Windows, but changing the locale to German and using CET should trigger it easily. I have to look into this myself for Kodi.

pganssle:
@dagwieers I don't access have any Windows machines. I think I was able to replicate it at one point when I had a Windows VM, but I think the best thing to do is to put it in one of the Windows CI jobs.

dagwieers:
@pganssle Wrt. Python 2.7 support, I am not convinced we want to roll this back if Python 2.7 support is no longer needed. IMO it makes more sense to return the internal (C) name rather than a localized/translated one.

That is also why I am doing this regardless of the Python version, otherwise I would have singled Python 2 out for the workaround.

PR #914, opened 2019-05-04
Fix tz environment

### LINKED ISSUE #908: Build against tz master failing in caching logic
I have been able to reproduce [this failure](https://travis-ci.org/dateutil/dateutil/jobs/523478272) locally, but I can't do it consistently, no idea why that is. It seems to be consistently happening on both Travis and Azure Pipelines, but when I pin `tzdata` to an old commit from before the breakage started, it's still broken.

I suspect maybe something in the testing stack had an upgrade between March 21st and 28th, and that may be breaking it specifically in the `tz` env for some reason.

pganssle:
## Summary of changes

Turns out that in [this commit](https://github.com/tox-dev/tox/commit/43d65b177f6cb9a8757554bf8988), `tox` stopped using absolute paths for `toxenvdir` and `toxworkdir` and such, which the script was relying on.

Regardless of whether that is changed back in `tox`, this should work, so we can now drop the pin and close the issue.

Closes #908 

### Pull Request Checklist


pganssle:
This is failing in a very weird way on Python 3.5, I can't reproduce it locally. Seems like `tox` is trying to pin both `setuptools==39.2.0` and `setuptools==41.0.1` for some reason. I cleared the cache and re-ran the job with the same result.

@gaborbernat Any idea what's going on there?

pganssle:
> Seems like `tox` is trying to pin both `setuptools==39.2.0` and `setuptools==41.0.1` for some reason.

Not sure why it's only happening on Python 3.5, but it might be related to [these lines from our `pyproject.toml`](https://github.com/dateutil/dateutil/blob/40924bd0392308914b01815f56d311c8478217b8/pyproject.toml#L3-L4):

```toml
[build-system]
requires = [
    "setuptools<40.0; python_version == '3.3'",
    "wheel",
    "setuptools_scm"
]
```

Perhaps `tox` is not respecting the environment specifications for some reason?

gaborbernat:
We do some deduplication, do you have the build log?

pganssle:
@gaborbernat Build log is here: https://travis-ci.org/dateutil/dateutil/jobs/534480063

gaborbernat:
GLOB sdist-make: /home/travis/build/dateutil/dateutil/setup.py... so this one does not uses isolated build system at all

pganssle:
@gaborbernat That doesn't really answer the question of why it's failing only on Python 3.5, but I suppose I can let it remain a mystery if turning on build isolation works.

gaborbernat:
No it doesn't, it just makes it a setuptools bug rather than tox bug. tox will build with whatever setuptools you have alongside tox in its host env.

pganssle:
@gaborbernat Not sure I understand what you mean by that, `setuptools` is definitely not invoking `pip`, and the build log says that this fails while installing the build dependencies:

```
Processing ./.tox/.tmp/package/1/python-dateutil-2.8.1.dev79+gd9defb0.d20190519.zip
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
  Complete output from command \
    /home/travis/build/dateutil/dateutil/.tox/py/bin/python -m pip install \
    --ignore-installed --no-user --prefix /tmp/pip-build-env-j6i54z3n \
    https://.../setuptools-41.0.1-py2.py3-none-any.whl#... \
    https://.../setuptools-39.2.0-py2.py3-none-any.whl#... \
    https://.../wheel-0.33.4-py2.py3-none-any.whl#... \
    https://.../setuptools_scm-3.3.3-py2.py3-none-any.whl#...
```

I suppose that could be something `pip` is doing or something `tox` is doing, but there's no way it's `setuptools`. It's not clear why it's starting just now.

gaborbernat:
Let me rephrase we do the sdist creation with an old setuptools probably as we do it with whatever is alongside tox. Installing the sdist into the virtualenv is done with your pyproject file which is failing.

pganssle:
Seems it's happening with the isolated builds, too: https://travis-ci.org/dateutil/dateutil/jobs/534498021#L1041

gaborbernat:
Can you run it with triple verbosity?

pganssle:
@gaborbernat Is there a way to turn off isolated builds for just one environment? I think there's no version of `pip` that supports both PEP 517/518 and Python 3.3.

gaborbernat:
Sadly no way to do that ☹️

pganssle:
From what I can tell from the [verbose build log](https://travis-ci.org/dateutil/dateutil/jobs/534500969), it seems like it's ignoring the environment specifiers on Python 3.5 for some reason. Hard to tell why, considering I can't reproduce locally.

Upgrading all VMs to xenial seems to fix the problem, so maybe it's a bug but it's hard to reproduce and easy to work around, so :man_shrugging:, guess I'll just do this.

For now I'll drop the `isolated_build` thing, unfortunately. Maybe I can configure the CI to work with it correctly at some point later on.

pganssle:
JFC. It breaks only on Python 3.6 for xenial. Whyyyyy.

pganssle:
Pinning `tox` seems to fix it, so this seems to be an actual regression in `tox`, but I can't reproduce it locally. Hopefully I'll find a bit of time later to find a MWE, but I strongly suggest it has to do with the duplicate requirement in `pyproject.toml`.

gaborbernat:
Feels like we keep the first occurrence of key... https://github.com/tox-dev/tox/blob/master/src/tox/package/builder/isolated.py#L34-L39 Can you create an issue for tox?

PR #915, opened 2019-05-17
Register pytest markers

pganssle:
The latest version of pytest has started warning on unregistered markers, which the test suite uses fairly extensively, causing the tests to fail. This explicitly registers all the markers 
used in the test suite.

In the future we can re-assess whether we want to replace these markers with something else, and also probably we'll want to start pinning all test and documentation dependencies.

### Pull Request Checklist


pganssle:
Test failure is unrelated, merging.

PR #934, opened 2019-07-06
Pin requirements more thoroughly on py33

pganssle:
## Summary of changes
This does two things to alleviate the current (and hopefully all future) problems we're having testing Python 3.3:

1. Upgrade to the latest supported `pip` on Python 3.3, version 10.0.1, which actually *does* seem to support `python_requires`
2. Add a constraints file to all `pip` operations for Python 3.3 pinning all direct and transitive dependencies.

With all dependencies pinned, the 3.3 build should be more stable.

### Pull Request Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,6 +19,15 @@ passenv = DATEUTIL_MAY_CHANGE_TZ TOXENV CI TRAVIS TRAVIS_* APPVEYOR APPVEYOR_* C
 commands = python -m pytest {posargs: "{toxinidir}/dateutil/test" --cov-config="{toxinidir}/tox.ini" --cov=dateutil}
 deps = -rrequirements-dev.txt
 
+[testenv:py33]
+description = run the unit tests with pytest under Python 3.3
+setenv = COVERAGE_FILE={toxworkdir}/.coverage.{envname}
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -19,6 +19,15 @@ passenv = DATEUTIL_MAY_CHANGE_TZ TOXENV CI TRAVIS TRAVIS_* APPVEYOR APPVEYOR_* C
 commands = python -m pytest {posargs: "{toxinidir}/dateutil/test" --cov-config="{toxinidir}/tox.ini" --cov=dateutil}
 deps = -rrequirements-dev.txt
 
+[testenv:py33]
```
pganssle: This line creates a new py33-only `testenv` just so that I can use different requirements/constraints files, everything else is the same.

@gaborbernat Do you think there's a more effective way to do this? I tried using environment selectors on the requirements.txt files, but tox didn't understand it.

In addition to the DRY violation, the big downside to doing this is that apparently tox -e py won't work properly on Python 3.3, because it won't automatically select the py33 environment.

gaborbernat:
I can't think of a better way than dropping support for python 3.3 😁

pganssle:
> I can't think of a better way than dropping support for python 3.3

I don't understand, you are saying that this is the best way *other* than dropping support for Python 3.3?

As I've said many times before, these tests are in place so that I can accurately determine when we actually add something that doesn't work in Python 3.3. So far, this has never happened, because all the things you need to support Python 2.7 also tend to give you Python 3.3 for free. Once we have a Python 3.3-specific breaking PR, we'll drop support for Python 3.3 (and 3.4, probably 3.5 as well).

PR #935, opened 2019-07-08
Fix error message for gettz when passed bytes

### LINKED ISSUE #927: gettz on Python 3 fails with TypeError when passed bytes
Given the following code on Python 3.7:
```python
import dateutil.tz
dateutil.tz.gettz(b'America/New_York')
```
A `TypeError` is raised:
```
  File "lib/python3.7/site-packages/dateutil/tz/tz.py", line 1551, in __call__
    rv = self.nocache(name=name)
  File "lib/python3.7/site-packages/dateutil/tz/tz.py", line 1612, in nocache
    if name.startswith(":"):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

labrys:
## Summary of changes

Fix error message for gettz when passed bytes

Closes #927 

### Pull Request Checklist
- [X] Changes have tests

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -115,7 +115,7 @@ switch, and thus all their contributions are dual-licensed.
 - bjv <brandon.vanvaerenbergh@MASKED> (@bjamesvERT)
 - gl <gl@MASKED>
 - gfyoung <gfyoung17@gmail.com> **D**
-- labrys <labrys@MASKED> (gh: @labrys) **R**
+- Labrys <labrys.git@gmail.com> (gh: @labrys) **R**
 - ms-boom <ms-boom@MASKED>
 - ryanss <ryanssdev@MASKED> (gh: @ryanss) **R**
 
--- a/changelog.d/935.misc.rst
+++ b/changelog.d/935.misc.rst
@@ -1,0 +1,1 @@
+Raise a more helpful TypeError message when passing a bytes zonename to gettz in Python 3.  Reported by @labrys (gh issue #927). Fixed by @labrys (gh pr #935)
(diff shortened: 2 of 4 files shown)


labrys:
## Results of timeit
### Initial Test
```python
python -m timeit -n 1000 -s "from dateutil.tz import gettz; gettz('America/New_York')"

# Before change
1000 loops, best of 5: 3.76 usec per loop
:0: UserWarning: The test results are likely unreliable. The worst time (241 usec) was more than four times slower than the best time (3.76 usec).

# After change
1000 loops, best of 5: 3.78 usec per loop
:0: UserWarning: The test results are likely unreliable. The worst time (204 usec) was more than four times slower than the best time (3.78 usec).
```
### Optimized test
The initial test experiences a slow initial call (i believe during loading of tzfile) so I optimized it to only count subsequent gettz calls.
```python
python -m timeit -n 1000 -s "from dateutil.tz import gettz; gettz('America/Los_Angeles')" "gettz('America/New_York')"

# Before change
1000 loops, best of 5: 2.42 usec per loop

# After change
1000 loops, best of 5: 2.28 usec per loop
```

## Conclusion
The timeit differences are statistically insignificant

pganssle:
@labrys Sorry for the delay in the review here, this mostly looks good, I've gone ahead and rebased this and pushed a few changes to it. We generally don't use `xfail` to enforce that a given exception is raised (instead use `pytest.raises`. Instead, we use `xfail` to write tests that *should* pass but don't because there's a bug (that way you get the regression test imposed immediately when you fix the bug).

I've gone ahead and refactored your tests. I also think that your example of `object()` being passed to `gettz` should probably also raise `TypeError`, since the `AttributeError` is exposing an arbitrary implementation detail, but I'll leave that for a different PR to fix.

labrys:
@pganssle Sounds good to me.

PR #947, opened 2019-08-20
Stop using deprecated Travis container environment

jdufresne:
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: python
 cache: pip
 python:


jbrockmendel:
@jdufresne two of the Travis failures look like network issues.  can you re-push to see if that is the case?

pganssle:
@jbrockmendel I don't think it's an intermittent failure, everything has been failing like that for a while. I spent a little time trying to fix, it, but I have been too swamped lately to do much and the Travis documentation wasn't the best.

I suspect what's happening is that we're pinning a too-old version of `pypy` that Travis no longer supports or it's aliased wrong or something. Presumably it's similar with `3.3`, where `3.3` doesn't have native support in Travis anymore.

We do have coverage for these in Azure pipelines, so it may be the best thing to do is to just make Azure pipelines required and drop these two requirements. The biggest problem with that is that right now Azure pipelines doesn't contribute to the code coverage (it requires setting up an upload token and such), and I haven't had time to configure it. I know that it's possible to do so, though, because we set up codecov on [grid-strategy](https://github.com/matplotlib/grid-strategy/blob/master/azure-pipelines.yml#L44)] and we're *only* using azure pipelines there.

hugovk:
You can also remove the other one:

```diff
     - python: 3.7
       # This is required until Travis has a default image that
       # can run Python 3.7
       dist: xenial
-      sudo: required
```

PR #955, opened 2019-08-27
Allow customizing TZFILES and TZPATHS through env vars 

### LINKED ISSUE #943: Allow configuring TZPATH through an environment variable
To facilitate the use of dateutil in platforms where the tzdata is not placed in an "standard location", it would be useful if the user could provide the directory where to look for tzdata through an environment variable. I think we might just need to look that variable up at import time [here](https://github.com/dateutil/dateutil/blob/cbcc0871792e7eed4a42cc62630a08ec7a78be30/dateutil/tz/tz.py#L1461).

This is done in other packages like [pytz](https://github.com/stub42/pytz/blob/62f872054dde69e5c510094093cd6e221d96d5db/src/pytz/__init__.py#L91) or [as a way to test dateutil](https://github.com/dateutil/dateutil/blob/cbcc0871792e7eed4a42cc62630a08ec7a78be30/dateutil/test/conftest.py#L27).

This could be controlled through the variable `dateutil.tz.TZPATHS` that can behave *a la sys.path* as suggested by @pganssle and try to look things up in the order provided by the list.

mariocj89:
## Summary of changes

Allow users to customize this hardcoded value through environment
variables to facilitate using dateutil on installations where the
hardcoded values are not desired.

Closes #943

### Pull Request Checklist
- [ ] Changes have tests


dateutil/tz/tz.py
```diff
@@ -1456,15 +1456,37 @@ def __repr__(self):
...
-if sys.platform != "win32":
-    TZFILES = ["/etc/localtime", "localtime"]
-    TZPATHS = ["/usr/share/zoneinfo",
-               "/usr/lib/zoneinfo",
-               "/usr/share/lib/zoneinfo",
-               "/etc/zoneinfo"]
-else:
-    TZFILES = []
-    TZPATHS = []
+def __get_tz_files():
```
mariocj89: Let me know if you want this change at all. When I was refactoring getting the path into a function as done for other global variables in this file it felt that the same change that we are applying to TZPATH should be applied to TZFILES. If you disagree, let me know and I can leave TZPATHS only.

dateutil/tz/tz.py
```diff
@@ -1456,15 +1456,37 @@ def __repr__(self):
...
+        return env_value.split(':')
+
+    if sys.platform == "win32":
+        return []
+
+    return ["/etc/localtime", "localtime"]
+
+
+def __get_tz_paths():
+    env_value = os.environ.get('DATEUTIL_TZPATH', None)
```
mariocj89: @pganssle what do you think we should do if an empty env var is set?

I was going to consider empty as not set, but I guess this can be used as a way to force dateutil to pick up the vendored tzdata information.

mariocj89: I think most people would want only the first one, at least in my experience.

If I am an end-user I will know where to look at and just set it. If I am a framework developer or the like I will probably set it as a default and allow users to reset it if they want.

In short: I'll change the  PR to make `''` be equal to not set. For `DATEUTIL_DEFAULT_TZPATH` I think YAGNI.

dateutil/tz/tz.py
```diff
@@ -1456,15 +1456,37 @@ def __repr__(self):
...
-               "/usr/lib/zoneinfo",
-               "/usr/share/lib/zoneinfo",
-               "/etc/zoneinfo"]
-else:
-    TZFILES = []
-    TZPATHS = []
+def __get_tz_files():
+    env_value = os.environ.get('DATEUTIL_TZFILES', None)
+    if env_value is not None:
+        return env_value.split(':')
```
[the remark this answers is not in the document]
mariocj89: Good point for windows! I would probably just use `os.pathsep` to keep consistency across paths, but happy to go with your proposal if you prefer it.

mariocj89:
@pganssle added a commit as fixup to ease review, let me know if you want me to squash them together. Happy to do it :)

mariocj89:
@pganssle I've updated  the PR to include the `_EXTEND` env var. 

I guess we can then leave this on hold until the PEP for `cpython` happens and a decision is taken for the naming of variables there and how it happens. 🤷‍♂ 

If you want to land this earlier with any change, let me know :)

mariocj89:
Any update on the final variable names or are we OK with having them being specific to dateutil for the time being? We could add the new variable afterwards if needed.

pganssle:
@mariocj89 Good timing on this, I have actually just gotten my proof of concept `zoneinfo` implementation into a workable state, and should be [preparing a PEP if necessary](https://github.com/python/steering-council/issues/20) this weekend.

I think we can wait a *little* bit longer in order to allow `dateutil` to be more consistent with the way the proposed new module will work.

mariocj89:
Commit squashed, removed `DATEUTIL_TZPATH_PREPEND` and updated the name of the variable to follow PEP615.

mariocj89:
Gentle ping @pganssle

mariocj89:
Oh no! We missed the anniversary of this PR, sorry little #995, I hope you had a great day 😆 .

@pganssle

mariocj89:
Gentle ping @pganssle

mariocj89:
@pganssle should we re-evaluate landing this? The zoneinfo patch seems to had stalled. We could land this and remove py2 support, as with this users will be able to load custom tzdata if they want to stay in py2.

If you want I can update this to use the same env var as cpython.

PR #966, opened 2019-10-04
Fix typo in setup.cfg warning configuration

pganssle:
## Summary of changes
It appears that we were relying on the "errors" filter to treat PendingDeprecationWarning as an error due to a typo in setup.cfg. This line is apparently redundant with the "error" line anyway, but it's best to have a working version of it if we're going to have it at all. I am leaving it in place to be explicit that even these "ignored by default" warnings should be errors, in case a default changes later.

### Pull Request Checklist


jbrockmendel:
LGTM

PR #971, opened 2019-11-02
Updates to the release process

pganssle:
## Summary of changes

The `RELEASING` document had gotten quite out of sync with the actual release procedure, so I have updated it.

I have also added new `tox` environments to handle the creation of virtual environments for some of the tasks around building and releasing.

### Pull Request Checklist


tox.ini
```diff
@@ -78,11 +78,32 @@ commands = sphinx-build -d "{toxworkdir}/docs_doctree" docs "{toxworkdir}/docs_o
...
+description = Build and sign an sdist and bdist
 basepython = python3.7
 skip_install = true
 passenv = *
 deps = click >= 7.0
       pep517 >= 0.5.0
-commands = python release.py
+commands =
+    python release.py build
+    python release.py sign
```
pganssle: Probably we should make this conditional on `gpg` being installed with a default key set or something, because I just realized that this probably will (or at least *should*) fail on Travis.

PR #973, opened 2019-11-03
Lock around strong cache updates

### LINKED ISSUE #901: Intermittent parse() KeyError failure with 2.8.0
We are seeing a KeyError exception with 2.8.0 intermittently. We have reverted to 2.7.5 for now. Seems like a thread safety issue with the new strong cache? Here's the traceback:

```
  File "site-packages\dateutil\parser\_parser.py", line 1358, in parse
  File "site-packages\dateutil\parser\_parser.py", line 657, in parse
  File "site-packages\dateutil\parser\_parser.py", line 1192, in _build_tzaware
  File "site-packages\dateutil\tz\_factories.py", line 40, in __call__
  File "collections.py", line 157, in pop
  File "collections.py", line 86, in __delitem__
KeyError: (None, -25200)
```

pganssle:
## Summary of changes
Since the strong cache uses operations in OrderedDict that are not thread-safe in Python 2, it is necessary to acquire a lock while updating the dictionary.

When Python 2 support is dropped, we can likely refactor this to avoid locking.

I have a small script that reproduces the issue on Python 2.7, but unfortunately it only reproduces the issue intermittently, and in the current implementation seems to be prone to deadlocks (possibly that is another issue with the tz code..), so it is not really an ideal unit test.

Still, I'll include it here in case anyone wants to verify that this solution works:

<details>

```python
from dateutil import tz

import threading
try:
    from threading import Barrier
except ImportError:
    from threading import Semaphore

    class Barrier(object):
        def __init__(self, n):
            self.n = n
            self.count = 0
            self.mutex = Semaphore(1)
            self.barrier = Semaphore(0)

        def wait(self):
            self.mutex.acquire()
            self.count = self.count + 1
            self.mutex.release()
            if self.count == self.n:
                self.barrier.release()
            self.barrier.acquire()
            self.barrier.release()

N = 2500
b = Barrier(N)
b2 = Barrier(N-1)

def constructor():
    b.wait()
    tz.tzstr("EST5EDT")
    b2.wait()

for i in range(N):
    threading.Thread(target=constructor).start()

b2.wait()
print("Done")
```

</details>

Closes #901.

### Pull Request Checklist

### DIFF
--- a/changelog.d/973.bugfix.rst
+++ b/changelog.d/973.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a race condition in the ``tzoffset`` and ``tzstr`` "strong" caches on Python 2.7. Reported by @kainjow (gh issue #901).
--- a/dateutil/tz/_factories.py
+++ b/dateutil/tz/_factories.py
@@ -2,6 +2,8 @@ from datetime import timedelta
 import weakref
 from collections import OrderedDict
 
+from six.moves import _thread
+
 
 class _TzSingleton(type):
     def __init__(cls, *args, **kwargs):
@@ -26,6 +28,8 @@ class _TzOffsetFactory(_TzFactory):
         cls.__strong_cache = OrderedDict()
         cls.__strong_cache_size = 8
 
+        cls._cache_lock = _thread.allocate_lock()
+
     def __call__(cls, name, offset):
         if isinstance(offset, timedelta):
             key = (name, offset.total_seconds())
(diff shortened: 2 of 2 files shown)


jbrockmendel:
How confident are you that this is only needed in py2?  it wouldnt be difficult to make the lock py2-only.

pganssle:
@jbrockmendel Not *terribly* confident. It seems like it's from [this closed bug in CPython](https://bugs.python.org/issue24667), but I'm not sure if PyPy3 is still affected by it (if they ever were). This is also fixing another race condition that is definitely there but only causes unnecessary cache misses as opposed to throwing errors.

If we had a reasonable test for this, I'd probably just change `cls.__cache_lock` to be `contextmanager.nullcontext` on Python 3 or something similar, but I don't think it hurts much to lock here, and if we find that it's a terrible performance burden we can cut a new release that is more clever about avoiding locks.

PR #976, opened 2019-11-18
migrate away from unittest and adopt pytest for test_imports

jpurviance:
## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/dateutil/test/test_imports.py
+++ b/dateutil/test/test_imports.py
@@ -1,166 +1,176 @@
 import sys
-import unittest
+import pytest
 
-class ImportVersionTest(unittest.TestCase):
+
+HOST_IS_WINDOWS = sys.platform.startswith('win')
+
+
+def test_import_version_str():
     """ Test that dateutil.__version__ can be imported"""
-
-    def testImportVersionStr(self):
-        from dateutil import __version__
-
-    def testImportRoot(self):
-        import dateutil
-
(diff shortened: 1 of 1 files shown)


dateutil/test/test_imports.py
```diff
@@ -1,166 +1,176 @@
...
-        from dateutil.tz import gettz
-        from dateutil.tz import tzwin
-        from dateutil.tz import tzwinlocal
-        from dateutil.tz import UTC
-        from dateutil.tz import datetime_ambiguous
-        from dateutil.tz import datetime_exists
-        from dateutil.tz import resolve_imaginary
+# Test that dateutil.tzwin related imports work properly
+@pytest.mark.skipif(not HOST_IS_WINDOWS, reason="Requires Windows")
+def test_import_tz_windows_direct(self):
```
pganssle:
```suggestion
def test_import_tz_windows_direct():
```

pganssle:
@jpurviance Can you rebase and add a news fragment for this?

Also there may be conflicts with #1007, though I think they will be minor if this gets merged first.

I'll also note that I think we can save it for a separate PR, but the "clean imports" fixture I suggested in #1007 would be useful for I think all of these functions. I wonder if it would make sense to leave these tests on one or more classes (just remove the `unittest.TestCase`) so that we can easily set up an autouse fixture or a class-based setup/teardown?

PR #982, opened 2019-11-20
Drop Python 3.3 on Appveyor

pganssle:
# Summary of changes

Appveyor has been failing recently because of Python 3.3. We're mainly still testing Python 3.3 as a sort of "canary" to determine when we can drop it from the `python_requires`, so I believe it's sufficient to just test on Travis.

I have also added Python 3.7 and 3.8 tests. For now we can do x86/x64 on 3.7 and 3.8 until 3.8 is the more common platform.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,6 @@ build: false
 environment:
   matrix:
     - PYTHON_VERSION: 27
-    - PYTHON_VERSION: 33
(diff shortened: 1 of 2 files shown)


jbrockmendel:
lgtm pending green

pganssle:
Thanks @jbrockmendel!

PR #989, opened 2019-12-28
Bug fix for tzoffset.

rajeee:
Since self._utczone stores UTC names in lower case, the argument 'name' should be cast to lower case to make the comparison, just like in other functions.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests


pganssle:
Thanks for catching this @rajeee. Would you mind adding a regression test for this?

At some point we should migrate these all to using `casefold` instead of `lower`.

PR #993, opened 2020-01-02
Fix custom repr for ParserError

### LINKED ISSUE #991: ParserError custom __repr__ is not reached
There is an indentation error in [the definition for `__str__`/`__repr__` in `dateutil.parser.ParserError`](https://github.com/dateutil/dateutil/blob/f6c6d5cb97ce460c9510846bf14e945b2b16ead0/dateutil/parser/_parser.py#L1603), which means that `__repr__` is never even defined for the class:

```
>>> from dateutil.parser import ParserError
>>> ParserError("Hi")
dateutil.parser._parser.ParserError('Hi')
```

I think the definition of the repr is also a bit problematic, in that even if you fix the indentation issue, it would return `ParserError(Hi)` instead of `ParserError('Hi')` in the above situation. I will make a PR shortly.

pganssle:
## Summary of changes
This was originally dead code, because an indentation error had `__repr__` defined after the `return` statement in `__str__`. The definition of the `__repr__` is also changed to be more in line with the conception of a repr as "what you would have to evalue to get this object".

Even though this is not guaranteed behavior, this commit also adds a regression test to avoid simple errors like this in the future.

Closes #991 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/991.bugfix.rst
+++ b/changelog.d/991.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the custom ``repr`` for ``dateutil.parser.ParserError``, which was not defined due to an indentation error. (gh issue #991, gh pr #993)
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -1600,8 +1600,9 @@ class ParserError(ValueError):
         except (TypeError, IndexError):
             return super(ParserError, self).__str__()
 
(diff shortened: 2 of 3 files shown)


jbrockmendel:
LGTM

pganssle:
Thanks for the review @jbrockmendel!

Issue #998, opened 2020-01-14
fuzzy_with_tokens parse should return the original date string elements


mcarans:
Currently fuzzy_with_tokens parse returns a datetime and the list of strings that were not part of the date. It would be very valuable to get the list of strings used to form the datetime.

eg. parse 'Myanmar Town July 2019 dataset' could return:
(datetime.datetime(2019, 7, 14, 0, 0), ('Myanmar Town ', ' ', 'dataset'), ('July', '2019'))

mcarans:
For those who need this functionality and find this issue, you can either use higher level functions I've implemented here:
https://github.com/OCHA-DAP/hdx-python-utilities#date-parsing-utilities

Or if you want the same interface as dateutils, then you use hdx.utilities.dateparse.parse which is modified for ambiguous dates to be day first D/M/Y where there are values in front of the year and day last Y/M/D where there are values after the year and for fuzzy_with_tokens, returns datetime, skipped tokens, date tokens.

pganssle:
Per my reasoning in #999, I think we will decline this functionality.

I think we have general vague plans to add an interface that will do something like this, by adding an intermediate form of `dateutil.parser.parse` that returns something closer to a `re.match` object including metadata about the parse (and hopefully one that can be re-used in a cache to make additional parses quicker).

I will leave this issue open for the moment until I have a chance to either find the existing issue to track this or open a new issue.

PR #999, opened 2020-01-15
Return datetime, skipped tokens, date tokens if fuzzy_with_tokens is True (Added date tokens)

mcarans:
## Summary of changes
Added return of date tokens eg.

        s1 = "Today is 25 of September of 2003, exactly " \
            "at 10:49:41 with timezone -03:00."
        self.assertEqual(parse(s1, fuzzy_with_tokens=True),
                         (datetime(2003, 9, 25, 10, 49, 41,
                                   tzinfo=self.brsttz),
                         ('Today is ', 'of ', ', exactly at ',
                          ' with timezone ', '.'),
                         ('25 ', 'September of 2003', '10:49:41',
                          '-03:00')))

Closes https://github.com/dateutil/dateutil/issues/998

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -79,6 +79,7 @@ switch, and thus all their contributions are dual-licensed.
 - Michael Aquilina <michaelaquilina@MASKED> (gh: @MichaelAquilina)
 - Michael J. Schultz <mjschultz@MASKED>
 - Mike Gilbert <floppym@MASKED>
+- Mike Rans  (gh @mcarans) **D**
 - Nicholas Herrriot <Nicholas.Herriot@gmail.com> **D**
 - Nicolas Évrard (gh: @nicoe) **D**
 - Nick Smith <nick.smith@MASKED>
--- a/changelog.d/999.feature.rst
+++ b/changelog.d/999.feature.rst
@@ -1,0 +1,1 @@
+Return datetime, skipped tokens, date tokens if fuzzy_with_tokens is True (Added date tokens).
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -643,7 +643,7 @@ class parser(object):
             default = datetime.datetime.now().replace(hour=0, minute=0,
                                                       second=0, microsecond=0)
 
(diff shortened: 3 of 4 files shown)


pganssle:
This would be a backwards-incompatible change, since it changes the signature of the function.

Theoretically we could add a `fuzzy_with_date_tokens` or something, but I don't really like how much the parser interface has grown. There are a few proposals for changing what the parser returns floating around. The most likely to be fruitful would be something that generates an intermediate structure containing information about where the parser got the datetime values it pulled from the string which would be used to both parse the string and for manual inspection.

As hard as it is for me to say when you've obviously put hard work into this fix, but I think we will probably not accept any contribution like this. I really appreciate your taking the time to work on this and to help improve dateutil, so I do not want to waste your time by misleading you into thinking that it's likely that this would get merged with some tweaks.

I'll leave this PR open for a bit to give you a chance to reply - I try to keep an open mind about this sort of thing, I just want to make it clear that the chances are slim to none, since a few of us have been hitting our heads against the wall trying to provide some reasonable functionality like this for a long time now :(.

mcarans:
@pganssle My suggestion would be to allow the user to supply an Optional[Dict] parameter called something like intermediate_results. By default it is None so that no intermediate results are returned (and existing code doesn't need to change). If the user supplies a dict, then intermediate results will be put in the dict. If more intermediate results need to be output in future, the function signature won't change eg. 

    intermediate_results = dict()
    parse(..., intermediate_results = intermediate_results)

intermediate_results would be returned with {'skipped_tokens': skipped_tokens, 'date_tokens': date_tokens} (and anything else that might be useful to return).

mcarans:
For those who need this functionality and find this PR, you can either use higher level functions I've implemented here:
https://github.com/OCHA-DAP/hdx-python-utilities#date-parsing-utilities

Or if you want the same interface as dateutil, then you use hdx.utilities.dateparse.parse which is modified for ambiguous dates to be day first D/M/Y where there are values in front of the year and day last Y/M/D where there are values after the year and for fuzzy_with_tokens, returns datetime, skipped tokens, date tokens.

Issue #1019, opened 2020-03-29
CI/CD Pipeline Status


ffe4:
I have been wondering if there are still plans to switch to Azure pipelines. The Azure badge says that the project was never built, but there are quite a few issues referencing Azure. In addition, GitHub actions has become an alternative for CI on Windows and OSX.

I was thinking about trying out GitHub Actions and Azure pipelines, and the Travis hiccup on #1017 made me think that now might be as good a time as any. I might as well use this repo for my tests on one of these platforms and maybe see if I can close some related issues at the same time (e.g. #873, #888, #889, #396).

My question would be whether there are already any plans or preferences, or whether there are other issues that might influence decisions on the pipeline.

pganssle:
We had Azure pipelines working for a while, but apparently Microsoft randomly decided that I was a bot or something and refused to allow me access to the account associated with `dateutil` without verifying my identity with a phone number or something, so I dropped Azure pipelines.

I've been using Github Actions for [zoneinfo](https://github.com/pganssle/zoneinfo) and I'm *somewhat* happy with it, but I'm not sure I'd be amazingly happy with it for our entire test matrix on dateutil for the following reasons:

1. It creates an annoying number of status hooks.
2. There is no option to re-run individual jobs - you can re-run *all* jobs or one job.
3. There's no way to re-run jobs that have passed, which means that if a job failed to upload to codecov or something, you have to make an actual *change to the PR* to get it to run again.

I'll note that I've seen problems as bad as the current Travis problems with GHA or AP, so I don't think it would be a cure-all. These CI providers just have random glitches sometimes, and I haven't found one to be amazingly more reliable than another.

I'm not sure if any of these things are major deal-breakers, though. Maybe we can start by moving the PyPy jobs over to GHA?

ffe4:
I would still expect quirks, but the situation right now, with three pipelines defined in the project just did not seem optimal. I think GHA might also be faster than Travis since jobs run in parallel.

If you don't really like the idea of running the entire test matrix on GHA, is there a platform that you think might be better suited for the project? I'd be somewhat hesitant about just adding yet another pipeline to the project. Or would you rather stick with Travis?

pganssle:
> I would still expect quirks, but the situation right now, with three pipelines defined in the project just did not seem optimal. I think GHA might also be faster than Travis since jobs run in parallel.

Travis jobs also run in parallel, it's just a different number of them running in parallel. I would not be terribly surprised if GHA were faster, though, that's true.

[...]

PR #1020, opened 2020-03-31
Migrate PyPy CI to GitHub Actions

### LINKED ISSUE #888: Add code coverage for Azure pipelines
I think `codecov` doesn't have direct interactions with Azure pipelines right now, so we'll need to maybe use the token method? I feel very uncomfortable checking any sort of authentication token in to CI, though.

I think maybe the best thing to do is to set it up as some sort of variable in the Azure pipelines console, and access that in the `azure-pipelines.yml`. Unfortunately I think that means I have to do it myself, though if someone knows of a good tutorial or blog post or something on how to do it that would help.

### LINKED ISSUE #889: Use Azure pipelines as main pypy tester
The Travis PyPy runs have been getting flaky lately, but we *are* testing them already on Azure pipelines. I'm thinking maybe we can switch them over to at least allowed failures on Travis in favor of the Azure builds.

Also, I'm wondering if we can solve #451 at the same time by adding Windows + PyPy builds for Azure pipelines.

This is semi-blocked on #888, since we still need coverage numbers from pypy.

ffe4:
## Summary of changes

As discussed in #1019, here is a PR for migrating the PyPy tests over to GHA.

- Add GHA workflow for PyPy tests
  I have used the official Codecov action for coverage reporting, and I think you shouldn’t even need to set the token for it to work. I have also unpinned tox, since it seems to work for these tests. Otherwise I have mostly reused what was already in the other pipelines, with a slight version bump on Postgres.
- Update `tox.ini` to run on Windows
  Just added `python -m` before commands.
- Remove PyPy jobs from Travis
  Note that I am not running a pypy3.5 test since it’s not installed on the GHA runner.

It would also be nice if you could have a look at whether the warnings for `updatezinfo.py` are expected on macos.

Assuming everything works satisfactory, could I also remove the `azure-pipelines.yml`? From your comments it didn't seem likely that it's ever gonna be used again. It'll still be in the commit history if it's needed later.

Closes #888, closes #889

### Pull Request Checklist

### DIFF
--- a/.github/workflows/test_pypy.yml
+++ b/.github/workflows/test_pypy.yml
@@ -1,0 +1,50 @@
+name: PyPy
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        python-version: [pypy2, pypy3]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    env:
+      TOXENV: py
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }} on ${{ matrix.os }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: python -m pip install -U tox
+      - name: Install zic (Windows)
+        run: |
+          curl https://get.enterprisedb.com/postgresql/postgresql-9.5.21-2-windows-x64-binaries.zip --output $env:GITHUB_WORKSPACE\postgresql9.5.21.zip
+          unzip -oq $env:GITHUB_WORKSPACE\postgresql9.5.21.zip -d .postgresql
+        if: runner.os == 'Windows'
+      - name: Run updatezinfo.py (Windows)
+        run: |
+          $env:Path += ";$env:GITHUB_WORKSPACE\.postgresql\pgsql\bin"
+          ci_tools/retry.bat python updatezinfo.py
+        if: runner.os == 'Windows'
+      - name: Run updatezinfo.py (Unix)
+        run: ./ci_tools/retry.sh python updatezinfo.py
+        if: runner.os != 'Windows'
+      - name: Run tox
+        run: python -m tox
+      - name: Generate coverage.xml
+        run: python -m tox -e coverage
+      - name: Report coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          file: ./.tox/coverage.xml
+          name: ${{ matrix.os }}:${{ matrix.python-version }}
+          fail_ci_if_error: true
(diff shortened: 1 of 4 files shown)


pganssle:
@ffe4 Would you mind sending me an e-mail? My address is in the README.

tox.ini
```diff
@@ -34,16 +34,16 @@ deps = coverage
 skip_install = True
 changedir = {toxworkdir}
 setenv = COVERAGE_FILE=.coverage
-commands = coverage erase
-           coverage combine
-           coverage report --rcfile={toxinidir}/tox.ini
-           coverage xml
+commands = python -m coverage erase
```
pganssle, line 37: Why change these?

ffe4: Because they failed on Windows without.

.github/workflows/test_pypy.yml
```diff
@@ -0,0 +1,44 @@
+name: PyPy
+
+on: push
```
pganssle, line 3: I think you need this to be:

```yaml
on:
    pull_request:
        branches: [master]
```

Or some variation on that.

ffe4: Right, that might make sense, my bad. It ran on the PRs on my repo, but it probably just reused the runs from my second branch. I'll try force push a change in a sec.

ffe4:
The Codecov step [timed out it seems](https://github.com/dateutil/dateutil/pull/1020/checks?check_run_id=552584516#step:10:1). I had that once as well on my end. I'm not sure if it's that important on every PR, so there's also the option of just letting it pass the Codecov step even if there is an error. I don't think there's a retry option.

ffe4:
The Codecov timeout seems to be a server-side issue and they're already [working on it](https://github.com/codecov/codecov-action/issues/37). Should hopefully be resolved soon.

.github/workflows/test_pypy.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Run tox
+        run: python -m tox
+      - name: Generate coverage.xml
+        run: python -m tox -e coverage
+      - name: Report coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          file: ./.tox/coverage.xml
+          name: ${{ matrix.os }}:${{ matrix.python-version }}
+          fail_ci_if_error: true
```
ffe4, line 50: Since Codecov has a server-side issue right now we could decide not to fail here, at least for now.

ffe4:
I just had to do another force-push because the GHA badge link in the README was apparently broken. My linter did not give me a syntax error and the badge loaded just fine in the browser, so I'd rather add it later. I also dropped the commit to remove azure pipelines that I forced earlier to try and trigger GHA.

Just waiting on Travis reporting and appveyor finishing now, but unless you want to allow Codecov to fail I'd be happy with this as it is.

pganssle:
This status hook thing is getting crazy annoying.

tox.ini
pganssle: I don't think this should be necessary, but I also don't think it hurts as long as it works, so I'm fine with it.

ffe4: I couldn't find the log anymore, but I think it was necessary for the steps to work on pypy3 but not for pypy2 for some reason. I just assumed it being one of those Windows things.

Issue #653, opened 2018-03-26
Python 2 deprecation schedule


pganssle:
I've been giving a lot of thought as to the Python 2 deprecation schedule for `dateutil`, and I think I have come up with a tentative plan. Before I go ahead and put it on [Python 3 statement](https://github.com/python3statement/python3statement.github.io).

Here is my current plan and timeline:

1. August-September 2018: **3.0 release** - Supporting Python 2.7, 3.4+
       We have a lot of deprecated behavior in the 2.x branch that I'd like to remove, and a few small breaks to backwards compatibility that I'd like to make that aren't really something that can be deprecated (e.g. the behavior of `dayfirst` and `yearfirst` that has so many people pinned on 2.4.x). Ideally, we'd also have the new parser up and running before then.
2. January 2019: **4.0 release** - Supporting only Python 3.4+. Version 3.0 becomes LTS, bugfixes only.
    This release should make basically no breaking changes other than dropping support for Python 2, so that people can write Python 2/3 compatible code pinned to `python-dateutil <= 5.0.0`, rather than doing more complicated backports.
3. July 2019 - Final bugfix release for version 3.0. Zoneinfo support continues indefinitely.

After the 4.0 release, I think we'll want to make at least a 6 month deprecation period before any more backwards-compatibility breaking changes are made.

Somewhere in here I'm also planning to put in a compiled backend, but I think that will be opt-in for at least the 2.x branch.

CC @jbrockmendel @jdufresne @jreback

mgedmin:
I think you meant "pinned to python-dateutil < 4.0.0".

pganssle:
@mgedmin No, I meant pinned to `python-dateutil<5.0.0` (though you're right, I didn't mean `<=`). The point would be that code written for 3.0 would still work with code written for 4.0, except that the 4.x branch doesn't work with Python 2. If you pin to `python-dateutil<5.0.0` on your app that supports both Python 2 and 3, your Python 2 users will get something from the 3.x line and your Python 3 users would get something from the 4.x line, assuming they use a `pip` that supports `Python-Requires`.

If you have an app or library that is Python 2-only, you'd pin to `python-dateutil<4.0.0`.

mgedmin:
Ah, I see!  I'm still not used to `Python-Requires` (which needs a very recent `pip`, IIRC).

Zac-HD:
[...]

PR #1024, opened 2020-04-03
Fix tz.gettz() not returning local time for empty string

### LINKED ISSUE #925: gettz('') fails on windows
Given the following (tested with Python 3.7 and dateutil 2.8.0):
```python
import dateutil.tz
print(dateutil.tz.gettz(''))
```

On Linux:
```
None
```

On Windows:
```
  File "<input>", line 1, in <module>
  File "H:\src\env\desc\lib\site-packages\dateutil\tz\tz.py", line 1551, in __call__
    rv = self.nocache(name=name)
  File "H:\src\env\desc\lib\site-packages\dateutil\tz\tz.py", line 1635, in nocache
    tz = tzwin(name)
  File "H:\src\env\desc\lib\site-packages\dateutil\tz\win.py", line 223, in __init__
    self._std_abbr = keydict["Std"]
KeyError: 'Std'
```

### LINKED ISSUE #926: gettz('') result is incorrect per the docs
Per the docs:
> If no argument or an empty string is passed to gettz, local time is returned:

On Linux it returns `None` instead of `tzlocal()`.

Related: #925

ffe4:
## Summary of changes

- `tz.gettz()` now properly checks for empty string and returns local time
- added test for `tz.gettz()` returning the same result when passed empty string or `None`

`nocache` checks the `gettz(name)` parameter with `if not name` and tries to get the local time from the `TZ` environment variable. When the environment variable does not exist, it only checked for `name is None`, skipping the logic that searches for the local time `tzfile` when passed an empty string.

Closes #925 
Closes #926 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/test/property/test_tz_prop.py
+++ b/dateutil/test/property/test_tz_prop.py
@@ -1,0 +1,35 @@
+from datetime import datetime, timedelta
+
+import pytest
+import six
+from hypothesis import assume, given
+from hypothesis import strategies as st
+
+from dateutil import tz as tz
+
+EPOCHALYPSE = datetime.fromtimestamp(2147483647)
+NEGATIVE_EPOCHALYPSE = datetime.fromtimestamp(0) - timedelta(seconds=2147483648)
+
+
+@pytest.mark.gettz
+@pytest.mark.parametrize("gettz_arg", [None, ""])
+# TODO: Remove bounds when GH #590 is resolved
+@given(
+    dt=st.datetimes(
+        min_value=NEGATIVE_EPOCHALYPSE, max_value=EPOCHALYPSE, timezones=st.just(tz.UTC),
+    )
+)
+def test_gettz_returns_local(gettz_arg, dt):
+    act_tz = tz.gettz(gettz_arg)
+    if isinstance(act_tz, tz.tzlocal):
+        return
+
+    dt_act = dt.astimezone(tz.gettz(gettz_arg))
+    if six.PY2:
+        dt_exp = dt.astimezone(tz.tzlocal())
+    else:
+        dt_exp = dt.astimezone()
+
+    assert dt_act == dt_exp
+    assert dt_act.tzname() == dt_exp.tzname()
+    assert dt_act.utcoffset() == dt_exp.utcoffset()
--- a/dateutil/test/test_tz.py
+++ b/dateutil/test/test_tz.py
@@ -1080,6 +1080,15 @@ class GettzTest(unittest.TestCase, TzFoldMixin):
 
 
 @pytest.mark.gettz
+def test_gettz_same_result_for_none_and_empty_string():
+    local_from_none = tz.gettz()
+    local_from_empty_string = tz.gettz("")
+    assert local_from_none is not None
+    assert local_from_empty_string is not None
+    assert local_from_none == local_from_empty_string
+
+
+@pytest.mark.gettz
 @pytest.mark.parametrize('badzone', [
     'Fake.Region/Abcdefghijklmnop',  # Violates several tz project name rules
 ])
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -1596,7 +1596,7 @@ def __get_gettz():
                     name = os.environ["TZ"]
                 except KeyError:
                     pass
-            if not name or name == ":":
+            if name is None or name in ("", ":"):
                 for filepath in TZFILES:
                     if not os.path.isabs(filepath):
                         filename = filepath
--- a/dateutil/tz/tz.py
+++ b/dateutil/tz/tz.py
@@ -1596,7 +1596,7 @@ def __get_gettz():
                     name = os.environ["TZ"]
                 except KeyError:
                     pass
-            if name is None or name == ":":
+            if not name or name == ":":
                 for filepath in TZFILES:
                     if not os.path.isabs(filepath):
                         filename = filepath
--- a/changelog.d/1024.misc.rst
+++ b/changelog.d/1024.misc.rst
@@ -1,0 +1,2 @@
+Fixed tz.gettz() not returning local time when passed an empty string.
+Reported by @labrys (gh issues #925, #926). Fixed by @ffe4 (gh pr #1024)


dateutil/test/test_tz.py
```diff
@@ -1077,6 +1077,12 @@ def testGettzCacheTzLocal(self):
         local2 = tz.gettz()
 
         assert local1 is not local2
+    
+    def test_gettz_same_result_for_none_and_empty_string(self):
```
pganssle, line 1081: I think we want to assert something specific about the result - that it's not None and that it is a local time.

I'd refactor this as a parameterized test parameterizing `args` over `() and ("",)`, then check that `tz.gettz(*args)` returns a `tz.tzlocal` or whatever it's supposed to return.

ffe4: Since it's platform dependent I was not sure if there is a good way to get local time other than using `gettz`. I actually had the following test:
```python
@pytest.mark.skipif(IS_WIN, reason='requires Unix')	
def test_gettz_returns_local_time(self):	
    with TZEnvContext('/etc/localtime'):	
        assert tz.gettz() == tz.tzfile('/etc/localtime')
```
However, since not all unix systems will have a tzfile at this location, this test is not really reliable. I don't really understand what `tz.tzlocal` is/does, so I was not sure if there is a way to use it here.

ffe4: I also thought `tz.gettz()` returning local time would hopefully have been tested already, although I did not find a test for it.

pganssle: Ah, that's a good point. We can start with just checking that it's not `None` and that it's the same as `tz.gettz()`.

I think we can add a test for the "is it the local time" behavior with some variation on this:

```python
def test_gettz_returns_local(self, example_datetime):
    dt_act = example_datetime.replace(tzinfo=tz.gettz())
    if six.PY2:
        dt_exp = example_datetime.replace(tzinfo=tz.tzlocal())
    else:
        dt_exp = example_datetime.astimezone()

    assert dt_act == dt_exp
    assert dt_act.tzname() == dt_exp.tzname()
    assert dt_act.utcoffset() == dt_exp.utcoffset()
    assert dt_act.dst() == dt_exp.dst()
```

We can either make this a property test and have `example_datetime` be generated by a `datetime` strategy or we can just parametrize over `example_datetime` and pick a few random datetimes and call it a day.

dateutil/test/test_tz.py
```diff
@@ -1079,6 +1082,28 @@ def testGettzCacheTzLocal(self):
...
+    assert local_from_empty_string is not None
+    assert local_from_none == local_from_empty_string
+
+
+@pytest.mark.gettz
+@pytest.mark.parametrize("gettz_arg", [None, ""]) # could add colon here
+@given(dt=st.datetimes(max_value=datetime.now()))
+def test_gettz_returns_local(gettz_arg, dt):
+    dt_act = dt.replace(tzinfo=tz.gettz())
+    dt_exp = dt.replace(tzinfo=tz.tzlocal())
```
ffe4, line 1099: I mostly used your suggestion, but used only `tzlocal()` since `astimezone()` sets a `timezone` instance that returns `None` for `dst()`

dateutil/test/test_tz.py
```diff
@@ -1079,6 +1082,28 @@ def testGettzCacheTzLocal(self):
...
+    local_from_none = tz.gettz()
+    local_from_empty_string = tz.gettz("")
+    assert local_from_none is not None
+    assert local_from_empty_string is not None
+    assert local_from_none == local_from_empty_string
+
+
+@pytest.mark.gettz
+@pytest.mark.parametrize("gettz_arg", [None, ""]) # could add colon here
+@given(dt=st.datetimes(max_value=datetime.now()))
```
ffe4, line 1096: Without setting a `max_value` the tests fail for me because they return standard and not summer time. Could this maybe be expected for dates further into the future (e.g. `datetime(2038, 4, 1, 0, 0)`) due to different data sources of `tzlocal` and `gettz`? I'm not sure about the whole timezone stuff yet.

dateutil/test/test_tz.py
```diff
@@ -1079,6 +1082,28 @@ def testGettzCacheTzLocal(self):
...
+def test_gettz_returns_local(gettz_arg, dt):
+    dt_act = dt.replace(tzinfo=tz.gettz())
```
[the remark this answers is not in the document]
ffe4: Since `datetime` bounds have to be naive I instead initialized the constant from a timestamp.

The test now also has a minimum bound, since I ran into an issue when `dt.astimezone()` started returning LMT (solar time?) in the 19th century.

This looks like a similar issue to the one with the upper bound, although I assume the problem in the case is probably that my local zone file starts at 1901 (`zdump -v /etc/localtime | head -n 1`), after the transition from LMT was already over, while `astimezone()` gets more historical information?

If that is the case I am not sure which date would be an appropriate lower bound, other than 1970.

dateutil/test/test_tz.py
ffe4: Since `datetime` bounds have to be naive I instead initialized the constant from a timestamp.

The test now also has a minimum bound, since I ran into an issue when `dt.astimezone()` started returning LMT (solar time?) in the 19th century.

This looks like a similar issue to the one with the upper bound, although I assume the problem in the case is probably that my local zone file starts at 1901 (`zdump -v /etc/localtime | head -n 1`), after the transition from LMT was already over, while `astimezone()` gets more historical information?

If that is the case I am not sure which date would be an appropriate lower bound, other than 1970.

pganssle: > The test now also has a minimum bound, since I ran into an issue when `dt.astimezone()` started returning LMT (solar time?) in the 19th century.

This is what it's supposed to do. LMT = Local Mean Time. I think you may be right that this is again a Version 1 vs. Version 2 issue. I believe the appropriate lower bound is something close to the negative equivalent of the epochalypse: `1901-12-13T20:45:52`.

If it works within that range (or even within a range starting ~24 hours after that), that is probably best.

1970-01-01 is also an *OK* lower bound in the sense that the TZ database doesn't make any guarantees about data quality before that, but I think that if we see any deviations in timestamps within the range `[-2**31, 2**31]` between `.astimezone()` and `tz.gettz()`, it may be because the two are using different data sources and not just different implementations reading the same data source; in that case, we may need to scrap the whole idea.

ffe4: I've just checked and you are correct, the first date in my zone file is in fact the negative epochalypse. I guess that's another date to keep in mind for when I run into time zone issues.

I had to initialize it with `timedelta` since I got a `ValueError` on Windows when initializing with `fromtimestamp()`.
```
ValueError: timestamp out of range for platform localtime()/gmtime() function
```

dateutil/test/property/test_tz_prop.py
```diff
@@ -0,0 +1,35 @@
...
+
+from dateutil import tz as tz
+
+EPOCHALYPSE = datetime.fromtimestamp(2147483647)
+NEGATIVE_EPOCHALYPSE = datetime.fromtimestamp(0) - timedelta(seconds=2147483648)
+
+
+@pytest.mark.gettz
+@pytest.mark.parametrize("gettz_arg", [None, ""])
+# TODO: Remove max_value when GH #590 is resolved
```
pganssle, line 16:
```suggestion
# TODO: Remove bounds when GH #590 is resolved
```

dateutil/tz/tz.py
```diff
@@ -1596,7 +1596,7 @@ def nocache(name=None):
                     name = os.environ["TZ"]
                 except KeyError:
                     pass
-            if name is None or name == ":":
```
pganssle, line 1599: I'll note that this will suddenly allow `tz.gettz(0)` without throwing an exception.

Maybe we should go with `if name is None or name in ("", ":"):`?

In a potential future version we may want to explicitly raise `TypeError` if `name` is not an instance of `str` or `None`, but I'm inclined to defer the discussion about duck typing/leniency and just implement this in such a way that we don't have to address it right now.

ffe4: Sure, makes sense. Although I would be interested to hear what you think about how we should do typing (#383) after the deprecation #653 

Also, should I squash my commits to clean up the commit history, will you just squash them all when merging, or what are your preferences in that regard?

pganssle: I personally rewrite my history so that each commit is an atomic change where the tests pass. If that's too much of a pain, squashing it down is also a valid approach.

PR #1037, opened 2020-04-27
Fix country typo in exercises

diceroll123:
It says Ireland, but the rest of the code says Israel. Fixed accordingly.


pganssle:
Good catch, thanks!

PR #1044, opened 2020-05-27
Do not require setuptools.command.test as it may not be available

### LINKED ISSUE #1043: setup.py requires setuptools.command.test which may be not available
On macOS the `test` package may not always be available (at least it happens in our case).
This is only used to give a hint about not running `python setup.py test`, which is not supported.
It should be possible to use the `setuptools.Command` class instead, which should always be available.
I'll submit a PR about this in a moment.

eldruin:
## Summary of changes

Use `setuptools.Command` class instead of `setuptools.command.test` since this may not be available in certain cases.

Closes #1043 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/1043.bugfix.rst
+++ b/changelog.d/1043.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed not requiring `setuptools.command.test` in setup.py and use the `setuptools.Command` class instead,
+as the former one may not be available. (gh issue #1043, gh pr #1044). Reported and fixed by @eldruin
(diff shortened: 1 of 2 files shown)


pganssle:
@eldruin Do you mind if we re-open this (I don't want to just re-open and use your contribution without your consent)?

I think that even though this really only occurs in unsupported situations, it would be better to give a good error message in these cases than to fail at the import stage.

eldruin:
Sure, no problem.

Issue #605, opened 2018-01-02
rrule drops microseconds


rbu:
If you have an rrule start on a datetime with non-zero microseconds, the microseconds of the generated datetimes are zero.

Steps to reproduce:
1. Generate an rrule like so: ```rrule(dtstart=datetime(2007,1,1,23,59,59,999999), ...)```

Expected result:
* First yielded item is ```datetime.datetime(2007, 1, 2, 23, 59, 59, 999999)```

Actual result:
* First yielded item is ```datetime.datetime(2007, 1, 2, 23, 59, 59)```

Reproducer code:
```python
>>> from dateutil.rrule import *
>>> from datetime import *
>>> list(rrule(DAILY, count=3, byweekday=(TU,TH), dtstart=datetime(2007,1,1,23,59,59,999999)))
[datetime.datetime(2007, 1, 2, 23, 59, 59), datetime.datetime(2007, 1, 4, 23, 59, 59), datetime.datetime(2007, 1, 9, 23, 59, 59)]
>>> [d.microsecond for d in rrule(DAILY, count=3, byweekday=(TU,TH), dtstart=datetime(2007,1,1,23,59,59,999999))]
[0, 0, 0]
```

pganssle:
Not sure why, but this is apparently a [deliberate feature](https://github.com/dateutil/dateutil/blob/master/dateutil/rrule.py#L428), first added [here](https://github.com/dateutil/dateutil/commit/46c8930ed776214161167a2b655125a73e89b9b6).

It's not actually tested, though. I'm not sure what [RFC 5545](https://tools.ietf.org/html/rfc5545) has to say about the precision of the dates specified. It's not hard to fix, but it's a bit of a software archaeology question as to whether or not we *should* fix it.

rbu:
Thank you for the the archaeological excavations. I see that from the RFC side, this behaviour is undefined and as such, resetting microseconds is somewhat consistent.
From an software API perspective, I do not see why parts of the dtstart would be dropped (hence the bug).

Let me know if there's something I can do to help.

pganssle:
By the way, if I were to make a guess as to why this was added, it would be that `datetime.now()` gives you the time with microsecond precision, but often that's just random detritus and you don't really care about it. Presumably it was seen that carrying around microsecond precision was more likely to be an annoyance to users than a benefit. It will need some thought as to how we can fix this without breaking those expectations.

ffe4:
[...]

PR #1055, opened 2020-06-17
Fix parser not using microseconds from default

### LINKED ISSUE #1032: parser doesn't include default microsecond when seconds are parsed
I'm trying to parse a datetime from a string using parser, while also supplying a default datetime, but I notice that when seconds are present in my input string, the microsecond value of the default datetime is not used.

Minimal working example:
```
>>> from dateutil import parser
>>> parser.parse("2020-02-15T12:34", default=datetime(2020,12,31,23,59,59,999999))
datetime.datetime(2020, 2, 15, 12, 34, 59, 999999)
>>> parser.parse("2020-02-15T12:34:23", default=datetime(2020,12,31,23,59,59,999999))
datetime.datetime(2020, 2, 15, 12, 34, 23)
```

python version: 3.8.2
dateutil version: 2.8.1

I've been looking through the parser code a little bit, and it seems like line 919, 951 and 1060 are interesting. seconds and microseconds are assigned in the same statement.

ffe4:
Parser was setting microseconds to zero when parsing seconds, even if no microseconds were part of the date string.

Closes #1032

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests


ffe4:
A similar issue exists for `rrule` #605. If we decide that we want to preserve microseconds here, I think it would make sense to implement the same behavior for `rrule` as well.

mariocj89 (approved):

pganssle:
I'm tempted to say we should go ahead with this, but I wouldn't be surprised if this would break someone's workflow because it's fairly common to truncate off the sub-second portions of a datetime if they are 0, so for example, someone generates a list of datetimes like this:

```python
dt_strs  = list(map(datetime.datetime.isoformat,
    [datetime.datetime(2020, 1, 1, 12, 45, 0, 1),
     datetime.datetime(2020, 1, 1, 12, 45, 0, 250000),
     datetime.datetime(2020, 1, 1, 12, 45, 0, 750000),
     datetime.datetime(2020, 1, 1, 12, 45, 1, 0),
    ]
))
```

Then parses them with:

```python
for dt_str in dt_strs:
    dt = dateutil.parser.parse(dt_str, default=datetime.datetime(2008, 12, 31, 0, 0, 0, 999999))
    print(dt)
```

The input is:
```
['2020-01-01T12:45:00.000001',
 '2020-01-01T12:45:00.250000',
 '2020-01-01T12:45:00.750000',
 '2020-01-01T12:45:01']
```

The current output matches the input:

```
2020-01-01 12:45:00.000001
2020-01-01 12:45:00.250000
2020-01-01 12:45:00.750000
2020-01-01 12:45:01
```

But after this change, the microseconds will match the default date's:

```
2020-01-01 12:45:00.000001
2020-01-01 12:45:00.250000
2020-01-01 12:45:00.750000
2020-01-01 12:45:01.999999
```

I still think it may be worthwhile, since the default value for `default` is basically `datetime.combine(datetime.now().date(), time(0, 0))`, so you'd only be affected by this if you are _actively specifying a default datetime with microseconds_, and if you are doing that, there's a good chance that you want truncated microseconds to be pulled from the default even if seconds are specified, but I'm really not sure.

samuel-coapps:
I ran into #1032, and I made very similar code changes, thinking to file a fix, before discovering this pull request.

I think that action ought to be taken since this behavior clearly contradicts the documentation for the API.

A concern raised was that correcting the behavior has a small chance of breaking existing application code. E.g., Joe the Python greenhorn wants to parse time strings and attach them to today's date. Joe provides a default of `datetime.now()`, but then Joe encounters unexpected behavior parsing strings like '12h 30s' (since the minutes will be pulled from the current timestamp, not set to zero). Joe then changes his default to `datetime.now().replace(hour=0,minute=0,second=0)`. Since Joe isn't very familiar with Python, he didn't realize that what he really wants to use is `datetime.now().date()`. If we fix the discrepancy between the API documentation and its actual behavior with respect to default microseconds, Joe's application will be broken!

One way of addressing this concern would be to apply default microseconds only when the user explicitly enables them (perhaps by passing in a boolean `default_microseconds` keyword argument which normally defaults to `False`). It's not the most elegant API design, but it might be the best compromise between backward compatibility and allowing the proper application of default microseconds. @pganssle and @ffe4, what do you think? I'm happy to work on implementing this in my spare time if you think that this could close #1032.

PR #1057, opened 2020-06-17
Fix unimplemented parser tests

ffe4:
The test cases for unimplemented parser cases were expected to fail, but some of them would not pass even if dateutil was able to parse them. I fixed some of these issues and also refactored one of the tests to be parametrized.

### Pull Request Checklist
- [ ] Changes have tests


dateutil/test/test_parser.py
```diff
@@ -833,7 +834,7 @@ def test_on_era(self):
     def test_extraneous_year(self):
         # This was found in the wild at insidertrading.org
         dstr = "2011 MARTIN CHILDREN'S IRREVOCABLE TRUST u/a/d NOVEMBER 7, 2012"
-        res = parse(dstr, fuzzy_with_tokens=True)
+        res, _ = parse(dstr, fuzzy_with_tokens=True)
```
pganssle, line 837: If we're not using the tokens, should we use `fuzzy=True` instead? Or is the bug only triggered when you capture the tokens?

dateutil/test/test_parser.py
```diff
@@ -775,8 +775,8 @@ def test_somewhat_ambiguous_string(self):
         # The parser is choosing the wrong part for hour
         # causing datetime to raise an exception.
         dtstr = '1237 PM BRST Mon Oct 30 2017'
-        res = parse(dtstr, tzinfo=self.tzinfos)
-        assert res == datetime(2017, 10, 30, 12, 37, tzinfo=self.tzinfos)
+        res = parse(dtstr, tzinfos={"BRST": gettz("America/Chicago")})
```
pganssle, line 778: I suppose `America/Sao_Paolo` would be a better choice for this (though it doesn't really matter — and it matters less since apparently this is one of the "made up" abbreviations that has since been dropped from the tz database).

dateutil/test/test_parser.py
ffe4: You are of course correct. I only noticed these by chance and quickly made the smallest change necessary to make them pass without even thinking about it.

PR #1058, opened 2020-06-19
Moved setuptools_scm config into pyproject.toml

KOLANICH:


pyproject.toml
```diff
@@ -1,12 +1,15 @@
 [build-system]
 requires = [
     "setuptools; python_version != '3.3'",
-    "setuptools<40.0; python_version == '3.3'",
+    "setuptools>=42; python_version == '3.3'",
```
pganssle, line 4: Python 3.3 is not supported in `setuptools >=42`.

I personally don't care much about where the setuptools_scm configuration lives *or* about continuing to support Python 3.3, so I'm ambivalent about this change, but if we go ahead with it, we should also drop all the 3.3-specific configuration, update the `python_requires ` and pin the minimum version in the non-3.3 case as well.

KOLANICH: Please note that setuptools is needed when building. Wheels can be consummed without that version of setuptools.

KOLANICH: I guess wheels still can be used by old pythons, it's only they cannot build wheels out of the box (but may be able with slight tweaks)

pganssle:
I'm going to say we'll leave this as is until we drop Python 3.3 (which will probably be when we drop Python 2).

PR #1060, opened 2020-06-22
Call .decode() only for binary data

frenzymadness:
## Summary of changes

I believe that the `__init__` method might be simplified in this way. In my opinion, it makes sense to call `decode()` method only on bytes/bytearray objects in both major Pythons.

In Python 2, `bytes` is builtin alias for `str` so `decode()` call does not happen for unicode.

Do you know about any corner case this simplified code is not aware of?

### Pull Request Checklist
- [ ] Changes have tests — this class is hopefully tested well enough
- [ ] Authors have been added to [AUTHORS.md](https://github.com/dateutil/dateutil/blob/master/AUTHORS.md) — not necessary, nothing really important contributed
- [ ] News fragment added in changelog.d. See [CONTRIBUTING.md](https://github.com/dateutil/dateutil/blob/master/CONTRIBUTING.md#changelog) for details — not important user-facing change

### DIFF
--- a/dateutil/parser/_parser.py
+++ b/dateutil/parser/_parser.py
@@ -60,14 +60,8 @@ class _timelex(object):
     _split_decimal = re.compile("([.,])")
 
     def __init__(self, instream):
-        if six.PY2:
-            # In Python 2, we can't duck type properly because unicode has
-            # a 'decode' function, and we'd be double-decoding
-            if isinstance(instream, (bytes, bytearray)):
-                instream = instream.decode()
-        else:
-            if getattr(instream, 'decode', None) is not None:
-                instream = instream.decode()
+        if isinstance(instream, (bytes, bytearray)):
+            instream = instream.decode()
 
         if isinstance(instream, text_type):
             instream = StringIO(instream)
--- a/changelog.d/1060.misc.rst
+++ b/changelog.d/1060.misc.rst
@@ -1,0 +1,1 @@
+Simplify handling of bytes and bytearray in _parser._timelex. Reported and fixed by @frenzymadness (gh issue #1060).


dateutil/parser/_parser.py
```diff
@@ -60,14 +60,8 @@ class _timelex(object):
...
     def __init__(self, instream):
-        if six.PY2:
-            # In Python 2, we can't duck type properly because unicode has
-            # a 'decode' function, and we'd be double-decoding
-            if isinstance(instream, (bytes, bytearray)):
-                instream = instream.decode()
-        else:
-            if getattr(instream, 'decode', None) is not None:
-                instream = instream.decode()
+        if isinstance(instream, (bytes, bytearray)):
```
vstinner, line 63: I suggest to also accept memoryview. For exampel, asyncio uses:

```
        if not isinstance(data, (bytes, bytearray, memoryview)):
            raise TypeError(f'data argument must be a bytes-like object, '
                            f'not {type(data).__name__!r}')
```

vstinner: I don't know if dateutil still supports Python 2. In that case, you may need to only use memoryview on Python 3.

frenzymadness: Implemented, thank you!

dateutil/parser/_parser.py
```diff
@@ -60,14 +60,8 @@ class _timelex(object):
...
     def __init__(self, instream):
-        if six.PY2:
-            # In Python 2, we can't duck type properly because unicode has
-            # a 'decode' function, and we'd be double-decoding
-            if isinstance(instream, (bytes, bytearray)):
-                instream = instream.decode()
-        else:
-            if getattr(instream, 'decode', None) is not None:
-                instream = instream.decode()
+        if isinstance(instream, (bytes, bytearray)) or six.PY3 and isinstance(instream, memoryview):
```
dmach, line 63: Wouldn't be better to do duck-typing and test for the presence of the decode() method?

```
if hasattr(instream, "decode"):
    instream = instream.decode()
```

mariocj89: If you pass in Python 2 a unicode string it will still attempt to decode it by encoding it first. And if that string has a non-ascii char, it'd fail to encode.

mariocj89, line 63: I think that what @vstinner meant with supporting memoryview was not this. This will fail when a memoryview is passed, as it does not have a `decode` method AFAIK.

If you want to add support for memoryview, please add tests for it. Alternatively, feel free to just remove `or six.PY3 and isinstance(instream, memoryview)` from this PR.

mariocj89 (changes requested):

frenzymadness:
I must admit that I kinda lost the context of this change after the eleven months. Give me a while or, if you know the solution, feel free to open a new PR based on this one and finish the fix.

mariocj89:
> Give me a while or, if you know the solution, feel free to open a new PR based on this one and finish the fix.

Sure, I'll push a commit with the additions.

dateutil/parser/_parser.py
mariocj89, line 63: At the same time, I see that commit 8e786465437b51123ab7644b8ce94b51d947c5b8 says "Add support for bytesarray and other bytes-like", not sure if there is any other case we are missing here than `bytes`, `bytearray` and `unicode`.

Let's see if the test suite catches anything.

mariocj89 (approved):
LGTM, but would be good to get another review as I rewrote your commit.

frenzymadness:
@vstinner could you please take a look?

vstinner (approved):
LGTM.

Ignore my remark about memoryview, memoryview objects have no decode() method.

PR #1069, opened 2020-07-26
Throw meaningful error when unsupported format for rrule

### LINKED ISSUE #960: unsupported property: FREQ=YEARLY
Using the following code:
```
rule = "FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59+01:00"
start = "2014-03-29 00:00:00+01:00"
rrulestr(date,dtstart=start)
```
I get: `ValueError: unsupported property: FREQ=YEARLY`
Is this a bug? Am I doing something wrong?

krithikabalu:
## Summary of changes

Throw meaningful error when rrule is in unsupported format.

Closes #960

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/dateutil/rrule.py
+++ b/dateutil/rrule.py
@@ -1697,7 +1697,7 @@ class _rrulestr(object):
                                          value)
                     dtstart = dtvals[0]
                 else:
-                    raise ValueError("unsupported property: "+name)
+                    raise ValueError("unsupported rule: "+s)
             if (forceset or len(rrulevals) > 1 or rdatevals
                     or exrulevals or exdatevals):
                 if not parser and (rdatevals or exdatevals):
--- a/dateutil/test/test_rrule.py
+++ b/dateutil/test/test_rrule.py
@@ -4912,3 +4912,8 @@ class WeekdayTest(unittest.TestCase):
 
         for repstr, wday in zip(with_n_reprs, with_n_wdays):
             self.assertEqual(repr(wday), repstr)
+
+    def testUnsupportedRuleErrorWhenNoMatchFound(self):
+        rule = "FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59+01:00"
+        with self.assertRaises(ValueError, msg="unsupported rule: " + rule):
+            rrulestr(rule, dtstart=datetime(1997, 9, 2, 9, 0))


krithikabalu:
I would like to clarify that failing build step is not because of my changes. I can see it failing without my changes too.

dateutil/test/test_rrule.py
```diff
@@ -4912,3 +4912,19 @@ def testWeekdayReprWithN(self):
...
+    def testUnsupportedRuleErrorWhenTimeFormatIsNotCorrect(self):
+        rule = "FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59+01:00"
+        with pytest.raises(ValueError) as exception:
+            rrulestr(rule, dtstart=datetime(1997, 9, 2, 9, 0))
+        self.assertEqual("Unsupported rule: " + rule + ". Check if property is supported and time format does not have "
+                                                       "delimiters according to RFC5545",
+                         str(exception.value))
+
+    def testUnsupportedRuleErrorWhenRuleNameIsNotCorrect(self):
+        rule = "XYZ:19970902T090000"
```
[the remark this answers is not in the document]
krithikabalu: thank you for sharing. Please have a look in my next commit.

krithikabalu:
@ffe4 Please check if my new change is aligned to the outcome you expect and let me know. Thank you!

dateutil/rrule.py
```diff
@@ -1664,6 +1664,8 @@ def _parse_rfc(self, s,
                     name = "RRULE"
                     value = line
                 else:
+                    if line.count(':') > 1:
+                        raise ValueError("Malformed rule: " + line)
```
ffe4, line 1668: While checking for the number of colons works, I am not sure it is the best approach. If I only use a colon for the timezone offset, we get the confusing error again:

```pycon
>>> rrulestr("FREQ=YEARLY;UNTIL=20190329T005959+01:00",dtstart=start)
  File "/Users/Daniel/ffe4/dateutil/dateutil/rrule.py", line 1734, in __call__
    return self._parse_rfc(s, **kwargs)
  File "/Users/Daniel/ffe4/dateutil/dateutil/rrule.py", line 1702, in _parse_rfc
    raise ValueError("unsupported property: " + name)
ValueError: unsupported property: FREQ=YEARLY
```
After looking at the code again, I think making this a separate error message might be a little more work than I thought at first. I think it would be okay to go back to the previous version, replacing the catch-all error message, but simplify the message a bit. Rather than referring to the RFC and some unspecified delimiter, we can just say that there cannot be any colon other than the one separating the property name. I also think it makes more sense to return the specific line, as you did here, rather than the complete rule.

dateutil/test/test_rrule.py
```diff
@@ -4912,3 +4912,17 @@ def testWeekdayReprWithN(self):
 
         for repstr, wday in zip(with_n_reprs, with_n_wdays):
             self.assertEqual(repr(wday), repstr)
+
+    def testMalformedRuleErrorWhenTimeFormatIsNotCorrect(self):
+        rule = "FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59+01:00"
+        with pytest.raises(ValueError) as exception:
+            rrulestr(rule, dtstart=datetime(1997, 9, 2, 9, 0))
+        self.assertEqual("Malformed rule: " + rule,
+                         str(exception.value))
```
ffe4, line 4921: Also, if we replace the catch-all error message, we probably should not add any tests for the message content either, but just check for a `ValueError`.

dateutil/rrule.py
krithikabalu: I understand your point. One thought I have: In addition to checking count of ':', can I also check the pattern for name before checking if it is equal to "RRULE" like this:
```
valid_name_pattern = re.compile("^[A-Z]+$")
if not valid_name_pattern.match(name):
      raise ValueError("Malformed rule: " + line)
elif name == "RRULE":
```

ffe4: I think you could do something similar. When we arrive at the catch-all error, we can add a check for whether we have a valid property name. If it is a valid property name we return that it is not supported, otherwise we return that we could not correctly parse the line (likely because of colons).

The pattern for the property name should also contain the hyphen `-` and maybe lower case letters for good measure (see [section 3.1](https://tools.ietf.org/html/rfc5545#section-3.1) and section [3.8.8.2](https://tools.ietf.org/html/rfc5545#section-3.8.8.2) for valid property names)

Of course, we could also check for a colon-separated datetime via the same method and return a helpful error message for that as well. But I think it is enough do mention datetimes in the error message for invalid property names.

krithikabalu: I have checked for specific cases as per your comment and made a commit. Please review. Thank you

dateutil/rrule.py
```diff
@@ -1728,6 +1730,19 @@ def _parse_rfc(self, s,
...
+    def __raise_error_if_unsupported_property(name):
+        valid_pattern = re.compile("^([A-Z]|[a-z]|-)+$")
+        if valid_pattern.match(name):
+            raise ValueError("unsupported property: " + name)
+
+    @staticmethod
+    def __raise_error_if_incorrect_timestamp(line):
+        hour_min_sec_format = ".*([\\d]{2}:[\\d]{2}:[\\d]{2}).*"
+        time_pattern = re.compile(hour_min_sec_format)
+        if time_pattern.match(line):
```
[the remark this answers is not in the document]
krithikabalu: Have looked through the comments. I will keep date time out of picture.

dateutil/test/test_rrule.py
```diff
@@ -4912,3 +4912,11 @@ def testWeekdayReprWithN(self):
 
         for repstr, wday in zip(with_n_reprs, with_n_wdays):
             self.assertEqual(repr(wday), repstr)
+
+    def testRaisesCorrectErrorForUnsupportedProperty(self):
+        rule = "X-ABC-MMSUBJ:19970902T090000"
+        self.assertRaisesRegex(ValueError, "unsupported property", rrulestr, rule, dtstart=datetime(1997, 9, 2, 9, 0))
```
ffe4, line 4918: The CI test fails because `assertRaisesRegex` was only added in python 3. Since we still support python 2.7 we need to add an import and use that instead.

```python
from six import assertRaisesRegex
```

```suggestion
        assertRaisesRegex(ValueError, "unsupported property", rrulestr, rule, dtstart=datetime(1997, 9, 2, 9, 0))
```

Also, the test is in the class `WeekdayTest`. Move it to `RRuleTest` instead.

ffe4:
Looks good to me. You might also want to add your name to the AUTHORS.md, however.

dateutil/test/test_rrule.py
```diff
@@ -4614,6 +4614,14 @@ def testReplaceIfNotSet(self):
         self.assertEqual(list(newrr),
                              [datetime(1997, 1, 6)])
 
+    def testRaisesCorrectErrorForUnsupportedProperty(self):
+        rule = "X-ABC-MMSUBJ:19970902T090000"
+        assertRaisesRegex(self, ValueError, "unsupported property", rrulestr, rule, dtstart=datetime(1997, 9, 2, 9, 0))
```
pganssle, line 4619: I don't care about this too much, but I'd prefer if we used `pytest` here instead of the unittest-specific stuff. I also strongly prefer a context manager form. The `pytest` version of this is:

```python
with pytest.raises(ValueError, match="unsupported property"):
    rrulestr(rule, dtstart=datetime(1997, 9, 2, 9, 0))
```

That said, there are two other issues here:

1. We care less about the "unsupported property" message than that we've picked out the _correct_ unsupported property.
2. The `dtstart=datetime(1997, 9, 2, 9, 0)` portion of this is unnecessary, and including it might confuse the reader into thinking that it's an important part of the test.

So, I suggest refactoring this test like so:

```python
def testRaisesCorrectErrorForUnsupportedProperty(self):
    with pytest.raises(ValueError, match=re.escape("X-ABC-MMSUBJ")):
        rrulestr("X-ABC-MMSUBJ:19970902T090000")
```

dateutil/rrule.py
```diff
@@ -1697,7 +1697,9 @@ def _parse_rfc(self, s,
                                          value)
                     dtstart = dtvals[0]
                 else:
-                    raise ValueError("unsupported property: "+name)
+                    if re.match("^[A-Za-z0-9-]+$", name):
+                        raise ValueError("unsupported property: " + name)
+                    raise ValueError("RRULE line contains invalid colon or property: " + line)
```
pganssle, line 1702: This is still pretty cryptic. I haven't read #960 in a while, and I tried running this and got:

```
ValueError: RRULE line contains invalid colon or property: FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59
```

It was very unclear what "invalid colon or property" meant in that situation.

How about instead we just go very wordy instead:

```python
                    if re.match("^[A-Za-z0-9-]+$", name):
                        raise ValueError("unsupported property: " + name)
                    raise ValueError(
                        "RRULE contains invalid property or stray colon:\n"
                        + line + "\n\n"
                        + "This commonly occurs when specifying a datetime "
                        + "using delimiters, e.g. 2025-06-07T12:34:56; RFC "
                        + "5545 requires datetimes specified with no "
                        + "delimiters, e.g. 20250607T123456")
```

dateutil/test/test_rrule.py
```diff
@@ -4614,6 +4614,14 @@ def testReplaceIfNotSet(self):
...
+
+    def testRaisesCorrectErrorIfDatetimeContainsColons(self):
+        rule = "FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59"
+        self.assertRaises(ValueError, rrulestr, rule, dtstart=datetime(1997, 9, 2, 9, 0))
```
pganssle, line 4623: I don't know if there is a good test that isn't too tightly bound to the specific error message, but one problem with this test is that it wouldn't have failed before this.

Possibly the right move is to split the difference and say that in this case we'll expect the entire line to be shown, not just the specific property that failed:

```python
def testRaisesCorrectErrorIfDatetimeContainsColons(self):
    rule = "FREQ=YEARLY;WKST=MO;UNTIL=2019-03-29T00:59:59"
    with pytest.raises(ValueError, match=re.escape(rule)):
        rrulestr(rule)
```

dateutil/test/test_rrule.py
krithikabalu: Got it, will change accordingly

dateutil/rrule.py
krithikabalu: Yea, this is better; will change accordingly

dateutil/test/test_rrule.py
krithikabalu: okay, will change accordingly

krithikabalu:
Any updates @pganssle ?

krithikabalu:
Any updates please? I hope this is still relevant

PR #1070, opened 2020-07-30
Update pyproject.toml

wiljanvanravensteijn:
restrict setuptools version to 43.0 since it is the last one that is compatible with python 2.7


pganssle:
We shouldn't need to do this. Anything that respects `build-system.requires ` should also respect `python_requires `, and `setuptools` sets `python_requires ` properly.

PR #1077, opened 2020-08-28
Fix isoparse issue where single-digit minutes or seconds was allowed at the end of a string

### LINKED ISSUE #1074: Incorrect value parsed as ISO date
**Version:**
`Python`: 3.8.2
`dateutil`: 2.8.1

**Description:**
The `dateutil.parse.isoparse` function parses an incorrect ISO string.

```python
from dateutil import parser


dt = parser.isoparse("20200728121240")
# Expected: `ValueError`
# What happens is:
print(dt)
# 2020-07-28 21:24:00
# datetime.datetime(2020, 7, 28, 21, 24)
```

**Expected Behavior:**
`ValueError` raised with a proper error message.

pganssle:
## Summary of changes

This was happening because in Python, slices are allowed to go past the end of a string, and `int(b'3'[0:2]) == 3`. We need to check that we didn't run past the end of the string when parsing these components.

Closes #1074 

### Pull Request Checklist
- [x] Changes have tests


ffe4 (approved):
LGTM but it would be great if we could make the error messages more consistent.

PR #1091, opened 2020-10-31
Support Internet RFC 8536 TZif files

eggert:
These are generated by modern-day zic, and zic -b slim
takes advantate of this format.
* ci_tools/run_tz_master_env.sh:
* dateutil/zoneinfo/rebuild.py:
Revert recent change that insisted on slim TZif files,
since the code now works on both fat and slim.
* dateutil/parser/_parser.py:
Add support for POSIX-2001 TZ strings, which can contain < and >.
Add support for RFC 8536 extensions to TZ strings, which
allow negative times in DST transitions.
* dateutil/test/test_tz.py: 'WART4WARST,J1,J365/-25' is now valid
due to RFC 8536 section 3.3.1, so add another '-'.  to make it
invalid.
* dateutil/tz/tz.py: Add support for TZif versions 2 and 3, which
have 64-bit timestamps, TZ strings for far-future timestamps,
and TZ strings involving negative transition times.

## Summary of changes

Closes 

### Pull Request Checklist
- [x] Changes have tests


eggert:
After creating this PR Github reports "Some checks were not successful", but looking at the details these all appears to be problems with the Github CI (e.g., "We are currently unable to download the log. Please try again later.") rather than actual bugs with the PR.

pganssle:
Hm.. My plan was to actually backport as much of the pure Python logic from `zoneinfo` as possible into `dateutil`, but these changes do seem fairly minimal.

That said, there are no tests for this (and I do have a fairly extensive test suite for `zoneinfo`), and I'm so behind on this package that the CI situation is a complete mess ☹.

I'll try and prioritize this soon, either by finishing the `zoneinfo` backport or adding a few tests to this and merging it.

pganssle:
Oh, and I forgot to say _thank you_ for taking the time to prepare this PR. It's very appreciated.

eggert:
> My plan was to actually backport as much of the pure Python logic from `zoneinfo` as possible into `dateutil`,

Yes, that would be nice as well. I did it by hand as I hoped these changes would be smaller and easier to audit.

At some point I suppose I should try to get `zoneinfo` fixed too, as it has some minor errors in this area.

> That said, there are no tests for this

There are tests, in that if you run the test suite on a platform where 'zic' defaults to 'zic -b slim', the changes are pretty thoroughly tested because 'zic -b slim' omits all 32-bit transitions and relies heavily on the TZ string at the end. That's how I tested the change (I also tested 'zic -b fat' of course). It would be nice to arrange to test the code with both 'zic -b fat' and 'zic -b slim' on all platforms that support 'zic -b', but I hope it's OK if I leave that detail up to you....

CraigeHales:
I copy/pasted the tz.py and _parser.py on top of existing versions used by a django project and they produce correct answer for this test case, attached. It is the US election day recurrence in the doc, run out to 2060, which is the first time after 2038 when daylight time has not ended (for future reference, daylight time ends on the first Sunday in November, and election day is the Tuesday after the first Monday.)
Without the change, the opening time for the polls is off by an hour, 7:30 rather than 6:30.
I did not do a code review.
[tztest.py.zip](https://github.com/user-attachments/files/19542700/tztest.py.zip)

niccokunzmann:
@CraigeHales I wonder: Could you create a separate pull request, if required based on this one? A zip file is more likely to be lost.

0-wiz-0:
I've just added this patch to pkgsrc - can this please be finally merged so NetBSD is properly supported again?

0-wiz-0:
And thanks for the work, @eggert !

PR #1122, opened 2021-05-04
prevent unnessary b' prefix in parse_isodate exception messages

pawl:
## Summary of changes

I started working on #383 by adding type hints to this branch: https://github.com/dateutil/dateutil/compare/master...pawl:monkeytype_merge_pyi?expand=1#diff-fba91b4117b31a0a29957adeb621fe79541fdb2424d10359e444df0d1cea2473R4

When I run mypy to see if those type hints work, I get a bunch of errors (a lot of which are probably my fault): https://gist.github.com/pawl/7ad1e9ce650b3241cc106d238d7edf04

This PR fixes one of those errors: 
>dateutil/parser/isoparser.py:167: error: On Python 3 '{}'.format(b'abc') produces "b'abc'", not 'abc'; use '{!r}'.format(b'abc') if this is desired behavior

That error is saying `isoparser().parse_isotime('2014-04-19T')` will give an error that looks like this with the `b'` prefix around the input:
```
String contains unknown ISO components: b'2014-04-19T'
````

This is happening because the `_takes_ascii` turns the string input into a bytestring, and python 3 represents bytestrings with `b'` when you pass them into `.format()`.

This PR fixes the issue by using `.decode()` on the bytestring before using it with `.format()` for the error message. 

### Pull Request Checklist
- [x] Changes have tests

### CODE REVIEW

**Comment by pawl** [comment]
@mariocj89 Thanks so much for the good feedback! I've added fixes for the things you've mentioned. Let me know if I need to make any additional tweaks.

File: dateutil/parser/isoparser.py [hunk 1]
```diff
@@ -159,7 +159,7 @@ def parse_isodate(self, datestr):
         components, pos = self._parse_isodate(datestr)
         if pos < len(datestr):
             raise ValueError('String contains unknown ISO ' +
-                             'components: {}'.format(datestr))
+                             'components: {}'.format(datestr.decode('ascii')))
```
mariocj89, line 162: Nit: I think it'd be even easier to read if we passed here: `'components: {!r}'`.

That will put `'` around the text that failed (Say for example that we have [some invisible characters](https://support.sas.com/resources/papers/proceedings/proceedings/sugi31/055-31.pdf).

mariocj89 (approved):

dateutil/parser/isoparser.py
pawl: Fixed: https://github.com/dateutil/dateutil/pull/1122/commits/f28e639ddcbaae341da5472a783381330ca67e0f

Definitely agree with removing the quotes making it more difficult to detect things like whitespace.

mariocj89:
Thanks, I've added a fixup for your tests to pass in Py2. I'll land this later today.

PR #1123, opened 2021-05-07
add missing dunder methods to rrule(set)

N-Coder:
This minor PR adds the missing __str__, __repr__, __eq__, and __hash__ methods to rrule and rruleset, which I've been missing when working with both in the context of ics files.


pganssle:
Hmm.. Sorry I haven't had much time to think about this stuff lately. I am not confident that we should be implementing `__hash__` here, since IIRC these things have state, no?

Possibly the only state involved is caches, in which case it's probably OK to add `__hash__`, though I'm not entirely sure how valuable such a thing is, particularly since there are many specific representations of what are (practically speaking) the same rule.

Is the implementation of `__str__` based on standards, or is it ad-hoc?

In the long run, if this is to be merged, it will need tests, but it may be worth starting out with an issue rather than a PR so that we can discuss the details and semantics.

N-Coder:
Let me maybe start with a quick summary of the methods I'm adding:
- `rrule.to_dict` - helper method that returns all relevant fields used for constructing the rrule
- `rrule.__repr__` - uses `to_dict` to generate the constructor call of the instance, such that `eval(repr(rrule)) == rrule)`
- `rrule.__eq__` - checks that the `to_dict`s of both objects are the same
- `rrule.__hash__` - hashes to `to_dict`s
- `rruleset.__str__` - converts a rruleset to an ics-compatible string (analogous to `rrule.__str__`), such that `rrulestr(str(rruleset)) == rruleset`
- `rruleset.__repr__` - prints the 4 contained lists to generated an `eval`able constructor call
- `rruleset.__eq__` - compares the 4 contained lists
- `rruleset.__hash__` - converts the 4 contained lists to hashable tuples and hashes them

As `rrule` is immutable, defining its [`__hash__`](https://docs.python.org/3/reference/datamodel.html#object.__hash__) method together with its `__eq__` should be the way to go. The case for `rruleset` is less clear: it doesn't allow passing the items for the 4 lists in the constructor, instead these are added incrementally by 4 corresponding (but somewhat strangely-named) methods. Still, it doesn't allow any further modifications to the 4 lists, especially no deletion is possible. Thus, I'd say that the usage of this class is mostly like with a builder - you create the instance, call the methods to initialize the object, in this case you don't call `finalize` to get the immutable instance, but still modifcations after the initialization are probably very rare. This probably makes most usages effectively immutable. The stored caches as only further "state" can be ignored, as they shouldn't be considered for equality testing, and thus also shouldn't be considered for hashing. So, `rruleset.__hash__` should (due to `rruleset`s semi-immutability) be the only candidate from the list above that needs further thoughts, all others should be good to go.

My reason to add `rruleset.__hash__` was that I was working with `vTimezone`s in ics, more precisely I was generating a dict for checking which vTimezones are used by which Events in an ics file. The `vTimezone` class is a simple frozen (and thus immutable) `attrs` data class implementing the python timezone interface. To be able to correctly represent all daylight-saving time transitions, it contains multiple `rrulesets` that model when these transitions happen (as defined by the ics standard). Thus, to be able to use `vTimezone`s as dict keys, they and especially all their members (including the `rruleset`) need to be hashable.

Still, we're somewhat in a gray zone with adding this method, it definitely would be useful, but you could also break things this way. Meaning that if you were to use a `rruleset` as dict key, to which you then later add further rules, your dict would break. Still, I don't think that this would ever happen, especially as the other and current alternative is that already using the `rruleset`s as key throws an "unhashable" `TypeError`. I guess python doesn't really prevent you from shooting in your own foot here, and we can't either - if you really wanted this functionality anyways you could always monkey-patch that method, as I'm currently doing with all of them. The cleanest way might be making `rruleset` immutable, but that would be breaking change which we should avoid.

> In the long run, if this is to be merged, it will need tests, but it may be worth starting out with an issue rather than a PR so that we can discuss the details and semantics.

Sure! I already had the code lying around, so I thought a PR might be better suited, but I'm definitely open to discussing and updating all these things. Once we reach an agreement on what should be done how, I can add tests for that your test suite. This should be easy to do, as most methods have a clear contract for their return value (as noted in my list above).

PR #1127, opened 2021-05-18
ci: Fix 3.4 test runs

mariocj89:
## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ install:
       pip install -U "pip==10.0.1"
       PIP_ARGS="-c requirements/3.3/constraints.txt"
     fi
+  - |
+    if [[ $TRAVIS_PYTHON_VERSION == "3.4" ]]; then
+      pip install -U setuptools
+    fi
   - pip install -U "tox<3.8.0" $PIP_ARGS
   - if [[ $TOXENV == "py" ]]; then ./ci_tools/retry.sh python updatezinfo.py; fi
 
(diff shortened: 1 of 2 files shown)


pganssle:
Alternative proposal: let's drop everything earlier than 3.6 from the CI run (but not tox or anything of that nature).

If any changes we make break on earlier versions Python, the only action that would trigger is we would change the `python_requires` and trove classifiers, so we only really need to test it before release.

I think we need to automate the release process anyway, so we can do it in two stages, start by dropping old versions of Python in CI, then add back in the old versions we still support (for now) as part of the release automation.

Also, I have no confidence that the Travis runs will keep working for very long, so probably don't bother with them if they are too fiddly.

mariocj89:
As spoken offline, I'll shoot a commit to move all "modern" Python's to github actions and we can later drop travis/appveyor once we drop support for those.

.travis.yml
```diff
@@ -33,6 +33,10 @@ install:
       pip install -U "pip==10.0.1"
       PIP_ARGS="-c requirements/3.3/constraints.txt"
     fi
+  - |
+    if [[ $TRAVIS_PYTHON_VERSION == "3.4" ]]; then
+      pip install -U setuptools
```
pganssle, line 38: Why doesn't `pip` handle this for us? From what I can tell `pip` didn't drop support for Python 3.4 until 19.2, but PEP 517 support went in in 19.0.

That said, I'm OK with whatever works here, this is all basically written on sand.

mariocj89: We need setuptools 52.0, let me fix pyproject.toml.

mariocj89: Nope, setuptools drop support for 3.4 in 44.0.0.

As said... if we plan to drop 3.4, I'd just land this to make the CI pass and we can look at it if we really care 🤷 .

PR #1129, opened 2021-05-19
ci: Move main interpreter checks to github actions

mariocj89:
Please review only f8b4571  and a09f9fe, other commits are part of different PRs. I'll rebase once those are landed.


pganssle:
Is it possible / easy to do 2.7 as a GHA as well? That's probably one of the more important ones until we release 3.0.

mariocj89:
> Is it possible / easy to do 2.7 as a GHA as well?

Sure, let me try that out.

pganssle:
OK, #1127 and #1128 have been merged, we should be able to move forward with this one now. Thanks so much for doing this, Mario!

mariocj89:
@pganssle let me know if  36ed255  works for you. It basically means that users might not be able to install distutil with the default `pip` `setuptools`, but to be honest I cannot figure out what is going on. It seems to always fallback to easy_install, I would have expected as you mentioned that pip will honour pyproject.toml and install with `pip` but I'm missing something.

If this is something that bothers you (you want CI to run with default `pip`), I can further look into the issue.

pganssle:
I'm still not clear on why the `pip install -U setuptools` is necessary, maybe some bug in the isolation somewhere?

In any case, it's nothing that needs to block things right now, so let's just merge.

mariocj89:
Mistery solved: https://github.com/dateutil/dateutil/pull/1131

Issue #1059, opened 2020-06-22
dateutil mishandles today's dates on NetBSD 9.0 etc.


eggert:
#462 talks about a longstanding bug in dateutil. In 1995, tzcode 95f introduced a 64-bit extension to TZif files, designed to make them work after the year 2038. dateutil's TZif parser does not use this data, causing it to mishandle timestamps after the last explicit 32-bit transition in a TZif file, which means timestamps after 2038 in (say) Los Angeles are mishandled by dateutil. I'm sure you've been meaning to fix this before 2038 rolls around.

Something new has happened, though. A year ago, [tzdb 2019b](https://mm.icann.org/pipermail/tz-announce/2019-July/000056.html) introduced the '-b slim' flag to the zic command, causing it to omit unnecessary data, the idea being to speed up the parsing of TZif files and make them smaller. Since the 32-bit data entries are always unnecessary, they're omitted. NetBSD 9.0, dated 2020-02-14, is using the '-b slim' flag to generate its TZif files, which means dateutil does not work even for today's timestamps on NetBSD. I assume other platforms will follow suit.

This change means it's time to boost the priority of fixing this bug.

To reproduce the problem, run the shell script datetime-slim-bug.sh (see below) in a writeable directory. It will create a slim TZif file 'Los_Angeles' and will use it to convert the time_t value 1592782632 into Los Angeles time. GNU 'date' handles this correctly, but datetime gets confused and thinks that Los Angeles is observing UTC. On my platform (Fedora 31 x86-64, Python 3.7.7, datetime 2.8.0) this script outputs:

```
2020-06-21 16:37:12-07:00
2020-06-21 23:37:12+00:00
```

The first line (from GNU 'date') is correct; the second line (from datetime) is wrong).

Here are the contents of datetime-slim-bug.sh:

```
#!/bin/sh
export LC_ALL=C
PWD=`pwd`

base64 -d >Los_Angeles <<'EOF'
VFppZjIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAEAAAAAAAAAVFppZjIA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB9AAAABQAAABT/////XgQawP////+epkig////
/5+7FZD/////oIYqoP////+hmveQ/////8uJGqD/////0iP0cP/////SYSYQ/////9b+dFz/////
[...]

PR #1130, opened 2021-05-20
Backport zoneinfo logic into dateutil.tz.tzfile

### LINKED ISSUE #462: tzfile reader only reads 32-bit (verson 0) zoneinfo files
The output of `zic` includes a 32-bit legacy section and a 64-bit section that encodes the transition times as 64-bit epoch times. Currently we're only reading the first one. What we *should* do is read the zoneinfo header to determine how long the first section is, seek to the end of that section and see if there's a second "version 2" section, and if so read that one instead, otherwise go back and finish reading the 32-bit version.

pganssle:
## Summary of changes

This moves `tzfile` into its own submodule containing what is effectively a backport of the pure python implementation of `zoneinfo.ZoneInfo`, but with the semantics of a `dateutil.tz.tzfile` object.

The major differences from zoneinfo are:

1. All caching logic is implemented in `tz.gettz` rather than in `tz.tzfile`.
2. `tzfile` objects are pickled by value, not by reference to the IANA key.
3. `tzfile` has equality-by-value semantics, whereas `zoneinfo` has equality-by-identity semantics.

This does change the internal implementation details of the class. It also fixes at least two bugs:

1. It adds support for V2 and V3 files, which also solves the 2038 problem (and the "slim zoneinfo" problem).
2. Prior to this change, `dateutil` did not use the `fold` attribute during gaps (only during folds), in violation of PEP 495. This fixes that, but that does change the semantics of imaginary times a bit (and thus may affect some mechanisms for detecting imaginary times).

This is mostly a code dump with compatibility adjustments. It's quite possible that we'll want to refactor, particularly with respect to the way POSIX strings are handled. Theoretically `tzfile` could fall back to a `tz.tzstr` or some other `tz.tzrange`, or we could refactor `tzstr` and `tzrange` in terms of _CalendarOffset, _DayOffset and _TZStr (or both, by refactoring the internal logic of `tzstr` and/or `tzrange`).

Closes #462, #1059

**Open Tasks**:
- [x] Add the `key` parameter for consistency with `zoneinfo.ZoneInfo`
- [x] Add support for customizing `TZPATH`
- [x] Backport relevant tests from zoneinfo
    - [X] Basic tests
    - [x] Weird zone tests
    - [x] TZStr-specific tests
    - [x] TZPath tests
    - [x] V1 support tests
- [x] Fix issue with the repr when falling back to `tzdata`
- [x] Add tests against the "slim" tzdata builds.
- [x] Format all new and changed code with `black` (at least trying to be incremental here when we can...)
- [x] Document all relevant changes

### Pull Request Checklist
- [x] Changes have tests


mariocj89:
@pganssle do you want me to review this?

pganssle:
@mariocj89 I still need to backport the tests, but you are welcome to review it. I expect in the long run we'll want to consolidate some of the logic around `tzstr` and `tzrange` and some of these custom rule representations built into the TZif parser (for the extended version 2 support that allows specifying a POSIX string), but I don't expect to overhaul anything else.

Also I believe some documentation changes will be necessary before we can merge this as well.

mariocj89:
OK, I'll take a look once all is ready.

Jorricks:
So @mariocj89 I suppose this is really the only remaining blocker for dropping support for Python2?
Given this has been open for ages, should I try to prioritize this MR, open a new MR and see if we can get somewhere within the next few months?

mariocj89:
@Jorricks this is indeed blocking most of the progress in the repo. If you have cycles to work on this we could drop Python 2 after we land and release this. @pganssle has been wanting to work on this for a long time but was not able to commit time to it.

kloczek:
+1

Jorricks:
Hey @mariocj89,
I'm attempting to start but I am missing some context. Could you help me out answering these questions :)?
1. How is this effort blocking the Python2 removal/deprecation effort?
2. If I understand correctly, the idea of this MR is to back port a copy of [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) (that is only available in Python3.9+) into this application which is compatible with the current `tzinfo` class right?
3. Is the idea to implement this feature Python3.7+ or should it be Python2 compatible? Mostly asking because I'd like to add type hints already :).

mariocj89:
> How is this effort blocking the Python2 removal/deprecation effort?

Without this change users in 2.7 will not get tz upgrades. This splits the tzdata from the package, allowing users to stay in an old version of the package but still get tzdata upgrades.

> If I understand correctly, the idea of this MR is to back port a copy of [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) (that is only available in Python3.9+) into this application which is compatible with the current tzinfo class right?

Correct

> Is the idea to implement this feature Python3.7+ or should it be Python2 compatible? Mostly asking because I'd like to add type hints already :).

Needs to work for py2, so no type hints :).

Jorricks:
> > How is this effort blocking the Python2 removal/deprecation effort?
> 
> Without this change users in 2.7 will not get tz upgrades. This splits the tzdata from the package, allowing users to stay in an old version of the package but still get tzdata upgrades.
> 
> > If I understand correctly, the idea of this MR is to back port a copy of [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) (that is only available in Python3.9+) into this application which is compatible with the current tzinfo class right?
> 
> Correct
> 
> > Is the idea to implement this feature Python3.7+ or should it be Python2 compatible? Mostly asking because I'd like to add type hints already :).
> 
> Needs to work for py2, so no type hints :).

Now it completely makes sense! Thanks.

Jorricks:
@mariocj89 to improve visibility on the progress & smoothen the process a little. Can we open a new branch containing this work so far?
We can just keep using this topic to keep track of everything that remains to be done so far I suppose.

I was diving a bit further (as I want to make sure I understand the full context before making changes) and was wondering the following:
During the building of the `python-dateutil` wheel, we package the `dateutil-zoneinfo.tar.gz`. This could(/should?) then also be replaced with a slim version of tzdata right?

mariocj89:
@pganssle do you have some time to guide @Jorricks through this change please?

sfc-gh-izinkovsky:
@pganssle are you still actively working on getting this change merged?

Jorricks:
Unfortunately I don't have the time anymore to work on this and made zero progress anyway. So I don't think anyone is working on this. 

The code is very complex and you are only doing this for python 2 users. Given that reached end of life years ago, I'm not to enthusiastic to work on this anyway.

kloczek:
Is it anything else on pending list before  release new version? 🤔

pganssle:
@kloczek This is the main feature blocking 3.0.

My plan is:

1. Set up release automation
2. Make a 2.8.3 release with what is on `master` now
3. Make a 3.0.0 release that includes this change.
4. Remove support for Python <3.7-ish?
5. Make a 4.0.0 release.

pganssle:
Current status of this branch:

I believe all features are complete, we just need to backport some of the tests from `zoneinfo` and add a test that uses "slim" builds for `tzdata`, to ensure the TZif Version 2+ support is working.

If anyone has time to review the code now, no need to wait for the backported tests.

nanonyme:
Are those 3.5 tests failing just because legacy versions of Python 3 do not stringify (through os.fspath) pathlib objects so you have to do it yourself?

a-detiste (approved):

pganssle:
Made a lot of progress on this at the sprints, but I don't think I'm going to quite get there today. I believe the main things left to do are:

1. Documentation of the changes
2. Test the tzfile with V1 data
3. Maybe adjust the repr to treat key/filename a bit better
4. Test the situation where only slim tzdata is installed

kytta:
Hey @pganssle, [you said](https://github.com/dateutil/dateutil/issues/653#issuecomment-1396351075) that this PR "is really not that hard for a motivated individual to finish". After two years since my last comment, I think I've become motivated enough :D

I've started by rebasing this branch against `master` to get rid of conflicts: https://github.com/kytta/dateutil/compare/new_tzfile. You can reset your branch to mine, if you wish.

However, I am kind of stuck on the list of "things left to do", so maybe you can comment on some things:

> 2. Test the tzfile with V1 data

I assume this means that V2 (and V3) tests are ready? If so, where can I find them? I am a bit lost looking at huge `test_tz.py` and `test_tzfile.py`... 😅 

Also, does this still mean that it would be enough to backport https://github.com/pganssle/zoneinfo/blob/666d80c27bda69541130758bbc7f1c9e035f79a0/tests/test_zoneinfo.py?

> 3. Maybe adjust the repr to treat key/filename a bit better

Do you mean `tzfile.__repr__` or something else?

> 4. Test the situation where only slim tzdata is installed

I guess I know how to do this in GitHub Actions, but do you have guidance on how to enable developers to test this, too?

---

I will take my time to properly understand the changes in this PR, so no rush with your reply :)

src/dateutil/tz/_tzfile.py
```diff
@@ -0,0 +1,1093 @@
...
+
+            dstoff = 0
+            utcoff = utcoffsets[idx]
+
+            comp_idx = trans_idx[i - 1]
+
+            if not isdsts[comp_idx]:
+                dstoff = utcoff - utcoffsets[comp_idx]
+
+            if not dstoff and idx < (typecnt - 1):
```
StanFromIreland, line 509: See https://github.com/python/cpython/pull/145885/changes#diff-7c5f7d13590c56bce99efe1cd8d9692be0953d9adf1aee20b24e4192e7d42280R341

src/dateutil/tz/tz.py
```diff
@@ -1625,35 +1082,38 @@ def nocache(name=None):
...
                                 continue
                         try:
-                            tz = tzfile(filepath)
+                            tz = tzfile(filepath, key=name)
                             break
                         except (IOError, OSError, ValueError):
                             pass
                     else:
                         tz = None
+                        if not tz:
```
StanFromIreland, line 1103: But you just set `tz = None`, so this is always true?

src/dateutil/__init__.py
```diff
@@ -6,13 +6,13 @@
...
+
 
 def __getattr__(name):
     import importlib
 
-    if name in __all__:
+    if name in __all__ or name == "zoneinfo":
         return importlib.import_module("." + name, __name__)
     raise AttributeError(
         "module {!r} has not attribute {!r}".format(__name__, name)
```
StanFromIreland, line 18: FYI, spotted an existing typo.

src/dateutil/__init__.py
```diff
@@ -21,4 +21,6 @@ def __getattr__(name):
 
 def __dir__():
     # __dir__ should include all the lazy-importable modules as well.
-    return [x for x in globals() if x not in sys.modules] + __all__
+    return (
+        [x for x in globals() if x not in sys.modules] + __all__ + ["zoneinfo,"]
```
StanFromIreland, line 25:
```suggestion
        [x for x in globals() if x not in sys.modules] + __all__ + ["zoneinfo"]
```
Did you mean?

src/dateutil/_tzdata_impl.py
```diff
@@ -0,0 +1,89 @@
...
+
+        return _nullcontext(io.StringIO(str_package_data))
+
+    def _open_binary(package, resource):
+        return io.BytesIO(pkgutil.get_data(package, resource))
+
+else:
+    import importlib.resources
+
+    if sys.version_info > (3, 9):
```
StanFromIreland, line 44:
```suggestion
    if sys.version_info >= (3, 9):
```
[It was added in 3.9, no?](https://docs.python.org/dev/library/importlib.resources.html#importlib.resources.files)

src/dateutil/_tzdata_impl.py
```diff
@@ -0,0 +1,89 @@
...
+def _load_tzdata_keys():
+    with _open_text("tzdata", "zones") as f:
+        return list(filter(None, (l.strip() for l in f)))
+
+
+class TZFileNotFound(ValueError):
+    """Indicates that a time zone file isn't present in tzdata."""
+
+    def __init__(self, msg, *args, **kwargs):
+        # TODO: Use keyword-only argument when this is Python 3-only
```
StanFromIreland, line 83: Python 2 usage has dropped severely these last few years, is it really worth all the pain? And furthermore, many don't like the `six` dependency (https://hugovk.dev/blog/2025/minus-six/).

tests/resources/data/zoneinfo_v1.json
StanFromIreland, line 1: I suggest you mark such files as generated in `.gitattributes`.

tests/resources/update_tzif.py
```diff
@@ -0,0 +1,241 @@
...
+        ],
+        cwd=cwd,
+        check=True,
+    )
+
+    # The tzdb tarball has for some reason a top-level directory called
+    # tzdb-..., which is annoying to deal with. We will delete that and
+    # unpack its contents into `output_dir`
+    for i, subdir in list(enumerate(output_dir.glob("tzdb-*"))):
+        if i > 1:
```
StanFromIreland, line 98:
```suggestion
        if i >= 1:
```
Doesn’t this allow for two currently?

tests/resources/update_tzif.py
```diff
@@ -0,0 +1,241 @@
...
+    # Ensure that this is a valid TZif file
+    contents = tzif_file.read_bytes()
+    if contents[0:4] != b"TZif":
+        raise ValueError(f"Invalid TZif data found! {tzif_file}")
+    version = int(contents[4:5])
+
+    header_start = 4 + 16
+    header_end = header_start + 24  # 6l == 24 bytes
+    if version < 2:
+        logging.warn(
```
StanFromIreland, line 120:
```suggestion
        logging.warning(
```

tests/resources/update_tzif.py
```diff
@@ -0,0 +1,241 @@
...
+    tasks = []
+    for tzif_in_file in filter(_is_tzif, input_base_path.rglob("*")):
+        tzif_v1_file = output_base_path / tzif_in_file.relative_to(
+            input_base_path
+        )
+        tasks.append(
+            asyncio.create_task(_file_to_v1(tzif_in_file, tzif_v1_file))
+        )
+
+    asyncio.gather(*tasks)
```
StanFromIreland, line 156: Missing `await?`

PR #1131, opened 2021-05-21
ci: Remove the need of upgrading pip/setuptools

mariocj89:
We were running tox without enabling isolated environment, this was
therefore resulting in tox not using `pyproject.toml`.

The commit enables isolated environment, and to make it work it also
needs to mark that the building of the tox `.package` env does not have
any dependency. This is needed as otherwise tox will attempt to install
the dependencies in `deps` when building the package as those will be
passed to PEP517, which does not support requirements files.

There you go :). @pganssle do you want me to add a news entry type misc for this kind of changes?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,12 @@ envlist = py27,
           docs
 minversion = 2.9.0
 skip_missing_interpreters = true
+isolated_build = true
+
+[testenv:.package]
+# no additional dependencies besides PEP 517 for building the package
+# Needed as we are running with an old version of tox.
+deps =
 
 [testenv]
 description = run the unit tests with pytest under {basepython}
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -32,8 +32,6 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Update pip and setuptools
-        run: python -m pip install -U pip setuptools
       - name: Install dependencies
         run: python -m pip install -U tox
       - name: Install zic (Windows)


pganssle:
> There you go :). @pganssle do you want me to add a news entry type misc for this kind of changes?

Historically I've had changelog entries for everything because when I was doing releases I would double check everything and make sure we didn't forget a changelog somewhere. If the list of entries in `changelog.d` had one entry for every PR that was merged and/or every issue closed, I knew I hadn't forgotten anything.

That said, I have a lot of overly persnickety procedures I've been doing with this project which has made it harder to get things done, leading to a great stagnation in my ability to work on it, so I'm thinking we should do whatever is sustainable.

The main things that I think _need_ a changelog entry are things where the change could plausibly have an effect on any user. So, changes to `setup.py` or `setup.cfg` need a changelog entry because people install from source. Changes to the tests need an entry because people run the test suite. Changes to the CI don't really need a changelog entry, though, because the CI is only run by us.

mariocj89:
Cool, I won't add it for this PR then and will follow that general rule. Good to land this then?

tox.ini
```diff
@@ -11,6 +11,11 @@ envlist = py27,
           docs
 minversion = 2.9.0
 skip_missing_interpreters = true
+isolated_build = true
+
+[testenv:.package]
+# no additional dependencies besides PEP 517 for building the package
+deps =
```
pganssle, line 18: This is only necessary because we are pinning to `tox<3.8.0`, right?

I modified `setup.py` to throw an exception if `pytest` or `six` is installed while building and on `3.7.0` it will fail, but on `3.21.1` it succeeds.

Considering that the downside of "the package is built with some additional stuff installed" is very low, maybe instead we should pin `tox` only on the `tz` job and leave this out?

mariocj89: Yup, I can remove that once https://github.com/dateutil/dateutil/pull/1132 is landed.

mariocj89: Right... but the tz environment still needs to run with tox<3.8.0, so it still needs that hack.

mariocj89: I've added it back

mariocj89: 🙄 and it fails py34 with the new version of tox. I'm gonna leave the requirement and we can remove all this crap once we drop 3.4 altogehter.

pganssle: OK, SGTM.

PR #1132, opened 2021-05-21
ci: Move docs, build and tz tox env validation to github actions

mariocj89:
Actions are split across multiple jobs to simplify setup and better
represent their intent.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -59,3 +59,58 @@ jobs:
           file: ./.tox/coverage.xml
           name: ${{ matrix.os }}:${{ matrix.python-version }}
           fail_ci_if_error: true
+
+  docs:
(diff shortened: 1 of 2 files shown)


mariocj89:
I need to look at the tz cache mechanism to understand why this is failing. Will check after holidays.

pganssle:
I think you are running up against #908, you can get it working again by pinning `tox` to `<3.8.0` in the `tz` environment for now.

Bernat and I tracked it down to some mega-commit where he refactored all of `tox`'s internals (my go-to example of why small atomic commits are useful). I think it's something to do with how environment variables or CWD are handled. I think #914 was the beginnings of a fix that I never wrapped up for some reason.

mariocj89:
@pganssle indeed! This is passing now.

pganssle (approved):

PR #1133, opened 2021-05-21
release: Move to `python-build`

mariocj89:
PEP517 is deprecated, move to `python-build`. This will additionally
validate the build dependencies, enhancing the CI.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,5 +5,5 @@ freezegun ; python_version != '3.3'
 hypothesis >= 3.30
 coverage
 mock ; python_version < '3.0'
-pep517 >= 0.5.0
+build >= 0.3.0 ; python_version >= '3.6'
 attrs!=21.1.0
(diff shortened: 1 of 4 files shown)


mariocj89:
What was expected to be easy is not, for some weird reason `python-build` seems not to work properly within tox. It does work fin within a venv and virtualenv though. I'll look into this once I come back from holidays, unless @gaborbernat knows something about this already :P.

requirements-dev.txt
```diff
@@ -5,5 +5,5 @@ freezegun ; python_version != '3.3'
 hypothesis >= 3.30
 coverage
 mock ; python_version < '3.0'
-pep517 >= 0.5.0
+build >= 0.3.0 ; python_version >= '3.6'
```
gaborbernat, line 8: I don't think you need the version qualifier here

mariocj89: This runs on Python 3.3, which is not supported by python-build I believe.

gaborbernat: Oh 😊 technically virtualenv also does not support that xD

pganssle (approved):
Might be worth adding a changelog entry for this one, since it might cause changes to the output wheels for some unknown reason (though admittedly at this point the "misc" section is omitted from the actual changelog, I'm wondering if we may want to change that before the next release.

PR #1158, opened 2021-07-31
Fix dateutil/dateutil#1157

### LINKED ISSUE #1157: relativedelta.relativedelta returns False when checked for inequality with None
```python
    print("timeUntilRun", timeUntilRun)
    print("type(timeUntilRun)", type(timeUntilRun))
    print("timeUntilRun != None", timeUntilRun != None)
    print("timeUntilRun is not None", timeUntilRun is not None)
    print("timeUntilRun == None", timeUntilRun == None)
    print("timeUntilRun is None", timeUntilRun is  None)
```

Output:
```
timeUntilRun relativedelta(hours=+1, minutes=+54, seconds=+4, microseconds=+560945)
type(timeUntilRun) <class 'dateutil.relativedelta.relativedelta'>
timeUntilRun != None False
timeUntilRun is not None True
timeUntilRun == None False
timeUntilRun is None False
```

Yes, you read that right. `timeUntilRun` is neither equal to None nor not equal to None.

FriskySaga:
## Summary of changes

Inequality checks for RelativeDelta objects with mismatched types will return more intuitive results, i.e., doesn't return False when comparing for inequality with None.

Closes #1157 

### Pull Request Checklist
- [x] Changes have tests


FriskySaga:
Kudos to @pganssle for his help with fixing this bug. I honestly probably wouldn't have thought to try that, and I still don't quite fully understand how that fixes the issue.

mariocj89 (commented):
I'm not sure this works for Python 2. Have you checked if there are tests in the test suite validating this?

IIRC the right thing to do here would be to do `return not self == other` for py2 compatible code. I need to look this up why it is this way though.

FriskySaga:
> I'm not sure this works for Python 2. Have you checked if there are tests in the test suite validating this?

Seems like [test_relativedelta.py was ran on Python 2.7 with the latest Ubuntu](https://github.com/dateutil/dateutil/pull/1158/checks?check_run_id=3211756863#step:8:78), so I believe the code changes work for Python 2.7.

dateutil/test/test_relativedelta.py
```diff
@@ -314,8 +314,8 @@ def testComparison(self):
         self.assertNotEqual(d1, d3)
 
     def testInequalityTypeMismatch(self):
-        # Different type
-        self.assertFalse(relativedelta(year=1) == 19)
+        self.assertFalse(relativedelta(year=1) == None)
```
pganssle, line 317: Please add a new test instead of changing an existing one. The existing test should still be valid.

FriskySaga: Addressed

FriskySaga:
Sorry to re-request the review late, I didn't know that there was a button to re-request reviewers and that I ought to use that button. First time making an open-source contribution involving code, and my second time making a GitHub pull request (we use Bitbucket at work).

FriskySaga:
My apologies for the delay once again, I didn't realize my branch had merge conflicts. They should be fixed now. Please advise if there's a better git workflow you would suggest... I rebased my five commits into one commit before merging with `dateutil/dateutil/master`

PR #1168, opened 2021-08-29
Bugfix: Plus/Minus logic failed in the month way when the last day of the month was involved.

### LINKED ISSUE #1167: last day of the month is wrong??
From documentation, it seems to correctly get last day of the month..
**How does it handle months with different numbers of days? Notice that adding one month will never cross the month boundary.**

```
>>> date(2003,1,27)+relativedelta(months=+1)
datetime.date(2003, 2, 27)
>>> date(2003,1,31)+relativedelta(months=+1)
datetime.date(2003, 2, 28)
>>> date(2003,1,31)+relativedelta(months=+2)
datetime.date(2003, 3, 31)
```
I tried, 
```
from datetime import date
print(date(2021, 6, 30) + relativedelta(months=-1))
```
which gives 2021-05-30. i should have been 2021-05-31.

Mifrill:
## Summary of changes

Fixed issue where compare with `relativedelta` would erroneously return the wrong last day of the month if the passed argument is the last day of the month.

Closes #1167

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -15,6 +15,7 @@ switch, and thus all their contributions are dual-licensed.
 - Adrien Cossa <cossa@MASKED>
 - Alec Nikolas Reiter <alecreiter@MASKED>
 - Alec Reiter <areiter@MASKED>
+- Aleksei Strizhak <alexei.mifrill.strizhak@MASKED> (gh: @Mifrill)
 - Alex Chamberlain (gh: @alexchamberlain) **D**
 - Alex Verdyan <verdyan@MASKED>
 - Alex Willmer <alex@moreati.org.uk> (gh: @moreati) **R**
--- a/changelog.d/1167.bugfix.rst
+++ b/changelog.d/1167.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed issue where relativedelta would erroneously returns wrong last day of the month
+if passed argument is a last day of the month. Reported by @hawkEye-01 (gh issue #1167).
+Fixed by @Mifrill (gh pr #1168)
--- a/src/dateutil/relativedelta.py
+++ b/src/dateutil/relativedelta.py
@@ -48,7 +48,7 @@ class relativedelta(object):
             the corresponding arithmetic operation on the original datetime value
             with the information in the relativedelta.
 
-        weekday: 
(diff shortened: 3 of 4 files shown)


pganssle:
I'm kind of surprised that this PR didn't break any existing tests, because the behavior described in #1167 is actually working as intended.

I am not at a computer, so I can try and give a more thorough explanation on that ticket later, but basically the "fall back to last day of month" is only supposed to happen if "same day on a different month" resolves to a date that doesn't exist, so 3/31 + 1 month = 4/30 and 5/31 - 1 month = 4/30, but 4/30 + 1 month = 5/30.

Maybe this PR can be repurposed to add tests enforcing the existing behavior?

Mifrill:
Hi, @pganssle 
> actually working as intended

well, then some of added tests are incorrect, for example:
```python
self.assertEqual(date(2021, 2, 28) + relativedelta(months=1), date(2021, 3, 31))
```
is it should be?:
```python
self.assertEqual(date(2021, 2, 28) + relativedelta(months=1), date(2021, 3, 28))
```
but it's really the end of the month, and the end of the next month would be 31, not 28, isn't it?
Same for `minus`, for example:
```python
self.assertEqual(date(2021, 2, 28) - relativedelta(months=1), date(2021, 1, 31))
```
is it should be?:
```python
self.assertEqual(date(2021, 2, 28) - relativedelta(months=1), date(2021, 1, 28))
```
but the last day of Jan is `31` not `28`, so it is a bit confusing.

pganssle:
Right, this PR and the tests accompanying it are not a desirable change, but the fact that you didn't have to change any existing tests in order to get CI to pass highlights a problem with our existing tests.

We can either close this PR and address that problem at a separate time or if you wanted to, you could change the tests you've added to reflect the actual desired behavior around these things.

The tests themselves will also need to be implemented differently - use pytest style `assert` statements and parameterize the tests instead of multiple asserts per test.

Mifrill:
@pganssle okay, I see the point. I think it's better to close this PR and open a new one with `doc` type with a pytest style tests suite and some notes for: https://github.com/dateutil/dateutil/blob/master/docs/relativedelta.rst

PR #1176, opened 2021-11-18
Implement ChainParser

### LINKED ISSUE #1153: Provide with a ChainParser that allows users to pass a list of formats or parser to try in order
Quite often users want to customize in some way what the default parser in this library does as they might be getting unexpected results or they have str that are not parseable by `dateutil.parser`.

 `dateutil.parser` is quite an overkill when the user knows the list of formats that they are willing to accept. This issue is to provide a dateutil.parser.

This new ChainParser will just take callables and invoke them in the specified order.

Proposed interface (to be updated as we iterate over it):

```python
parser = dateutil.parser.ChainParser(
    lambda dtstr: datetime.datetime.strptime(dtstr, "%d %B, %Y"),
    dateutil.parser.isoparse,
    dateutil.parser.parse
)
```

Something I'd like to add to improve the interface is to also take `str` as a possible "parser" and just use it in `strptime`. I think requesting users to create a lambda for every format is unnecessary. Example:

```python
parser = dateutil.parser.ChainParser(
    "%d %B, %Y",
    dateutil.parser.isoparse,
    dateutil.parser.parse
)
```

Q: Do we want to provide in any way information about what parser was used as a way to let users discover what format was used?

mariocj89:
## Summary of changes
This will provide users with a new parser type that allows to parse a
string using a list of parsers or string formats to be used in strptime.

This class can be used when users have an idea of potential formats that
the string might have been formatted with and still allows to be
combined with other dateutils parsers.

Closes #1153

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/src/dateutil/parser/chain_parser.py
+++ b/src/dateutil/parser/chain_parser.py
@@ -1,0 +1,70 @@
+import datetime as dt
+
+
+class _FixedStrptime(object):
+    """Parses datetime from strings by using a format that will be fed to strptime"""
+    def __init__(self, format_):
+        self._format = format_
+
+    def __call__(self, dt_str):
+        return dt.datetime.strptime(dt_str, self._format)
+
+    def __repr__(self):
+        return "StrptimeParser(%r)" % (self._format)
+
+
+class ChainParser(object):
+    """Parser that is built from combining multiple parsers
+
+    All provided parsers should have a call method that takes a string as input
+    and raise `ValueError` if they cannot parse the datetime.
+
+    If the parser provided is a string, a parser will be constructed which
+    uses :func:`datetime.datetime.strptime` with the provided format.
+
+    This is useful for situations where you know that your dates are in one of a
+    small number of formats, or where you want to specifically prioritize some
+    formats over others. For example, if you want to interpret a 6-digit number as
+    ``YYMMDD``, then ``DDMMYY``, then fall back to `dateutil.parser.parse` for any
+    strings that don't match either of those, you could construct the following
+    ``ChainParser``:
+
+    .. doctest:: chainparser
+
+        >>> parser = ChainParser("%y%m%d",  "%d%m%y", dateutil.parser.parse)
+        >>> print(parser("120101"))
+        2012-01-01 00:00:00
+        >>> print(parser("010199"))
+        1999-01-01 00:00:00
+
+
+    """
+    def __init__(self, *parsers):
+        self._parsers = [
+            _FixedStrptime(parser) if isinstance(parser, str) else parser
+            for parser in parsers
+        ]
+        for parser in self._parsers:
+            if not callable(parser):
+                raise ValueError(
+                    "Invalid parser {0!r}, the parser is not callable."
+                    .format(parser)
+                )
+
+    def __call__(self, dt_str):
+        for parser in self._parsers:
+            try:
+                return parser(dt_str)
+            except ValueError:
+                pass
+        raise ValueError(
+            "Unable to parse string {0!r} with the available parsers"
+            .format(dt_str)
+        )
+
+    def __repr__(self):
+        chain_parser_args = [
+            repr(p._format) if isinstance(p, _FixedStrptime) else repr(p)
+            for p in self._parsers
+        ]
+        return "ChainParser(%s)" % ", ".join(chain_parser_args)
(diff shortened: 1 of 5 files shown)


tests/test_chain_parser.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+
+import datetime as dt
+
+
+def custom_parser(dt_str):
+    return dt.datetime(1989, 4, 24)
+
+
+def test_parser_repr():
```
pganssle, line 12: I think we maybe don't want to enshrine the exact repr in tests, particularly when it involves the repr of a subclass.

We could maybe mark this as a "smoke test" with `pytest.mark.smoke`, to indicate it's not guaranteed?

Another option might be if we change it so that the repr of `ChainParser("%Y%m%d", "%Y-%m-%d")` is `ChainParser('%Y%m%d', '%Y-%m-%d')`, which seems more consistent with the idea that, if possible, you should be able to copy-paste a repr into a REPL to re-create an equivalent object.

tests/test_chain_parser.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    assert repr(parse) == "ChainParser(StrptimeParser('%Y%m%d'), StrptimeParser('%Y-%m-%d'))"
+
+
+def test_dateutil_parsers():
+    parse = ChainParser(
+        dateutil.parser.isoparse,
+        dateutil.parser.parse,
+    )
+    assert parse("2019-12-04T00:00:00") == dt.datetime(2019, 12, 4)
```
pganssle, line 25: I think maybe we want these to be subtests or parameterized tests.

tests/test_chain_parser.py
```diff
@@ -0,0 +1,46 @@
...
+    assert parse("2019/12/04T00:00:00") == dt.datetime(2019, 12, 4)
+
+
+def test_custom_parser():
+    parse = ChainParser(
+        "%Y%m%d",
+        custom_parser,
+    )
+    assert parse("20191204") == dt.datetime(2019, 12, 4)
```
pganssle, line 34: Same here — subtests or parameterized tests rather than multiple asserts.

src/dateutil/parser/chain_parser.py
```diff
@@ -0,0 +1,71 @@
...
+class ChainParser(object):
+    """Parser that is built from combining multiple parsers
+
+    All provided parsers should have a call method that takes a string as input
+    and raise `ValueError` if they cannot parse the datetime.
+
+    If the parser provided is a string, a parser will be constructed which
+    uses :func:`datetime.datetime.strptime` with the provided format.
+
+    The following example will attempt some custom parsing formats and if that
```
pganssle, line 25: Rewording of this paragraph:

```
The following example first attempts to parse using two custom parsing formats,
then falls back to :func:`dateutil.parser.isoparse`, and finally to a
``dateutil``'s fuzzy parser:
```

src/dateutil/parser/chain_parser.py
```diff
@@ -0,0 +1,71 @@
...
+        return dt.datetime.strptime(dt_str, self._format)
+
+    def __repr__(self):
+        return "StrptimeParser(%r)" % (self._format)
+
+
+class ChainParser(object):
+    """Parser that is built from combining multiple parsers
+
+    All provided parsers should have a call method that takes a string as input
```
pganssle, line 19: It is probably worth an additional paragraph about why you would want to use this:

```
This is useful for situations where you know that your dates are in one of a
small number of formats, or where you want to specifically prioritize some
formats over others. For example, if you want to interpret a 6-digit number as
``YYMMDD``, then ``DDMMYY``, then fall back to `dateutil.parser.parse` for any
strings that don't match either of those, you could construct the following
``ChainParser``:

.. doctest:: chainparser

    >>> parser = ChainParser("%y%m%d",  "%d%m%y", dateutil.parser.parse)
    >>> print(parser("120101"))
    2012-01-01 00:00:00
    >>> print(parser("010199"))
    1999-01-01 00:00:00
```

If we're including the example, we should probably put it right before the other example rather than before explaining the inputs.

mariocj89: I'm thinking on deleting the other example and just leave this one

mariocj89:
Seems the gh actions do not run with 3.6 anymore 😢 . Do you want me to drop that?

Issue #251, opened 2016-04-14
ValueError on YYDDMM string


pganssle:
Using the current master branch version, running:

```python
from dateutil.parser import parse

parse('201508')
```

Raised `ValueError` constructing the date. Originally reported as a comment on #242, and I suspect the original poster wanted it to be parsed as YYYYMM, but the parser _should_ parse this to `datetime(2008, 8, 15)` (YYDDMM), since an assumption of the parser is that no one puts the year in the middle, this date should probably by rights be considered to be YYDDMM string.

That said, while we could fix this now if anyone cares, I'm thinking that as part of the 2.6.0 release, we'll want to give end users much finer control over how ambiguous strings are parsed, and I think that doing that will make this bug disappear naturally. I also suspect that if we fixed this bug, a bunch of people would show up complaining that the behavior of the parser changed because they were counting on it throwing an error in this case.

pganssle:
@jbrockmendel Since the parser needs a lot of work,, I think I'm going to push it out the later releases so we're not blocking the release of 2.7.0. This one seems like it *might* be easy to tackle, any thoughts?

jbrockmendel:
I think pushing this out makes sense.  In my usage, I encounter %Y%m _much_ more often than %d%m%y.  I'd rather take a little longer to think through/discuss the desired behavior.

PR #1179, opened 2021-12-01
docs: YYYYMM is not supported

jpmckinney:
Remove "YYYYMM" from the docs, as it is not actually supported (https://github.com/dateutil/dateutil/issues/251).

To keep it simple, I do not need to be added to AUTHORS.md.

I don't think a docs fix requires a news fragment.


mariocj89 (approved):
Thanks @jpmckinney !

PR #1214, opened 2022-05-15
Remove weird, unused count parameter

NicoHood:


pganssle (approved):

pganssle (changes requested):
Oh wait, I forgot this needs a changelog entry for sure.

It's technically a breaking change, though it will only break people whose code is probably mildly broken anyway I guess?

NicoHood:
Yes, it should be documented. Feel free to do so.

Issue #1348, opened 2024-03-01
Version 2.9.0 is not compatible with aws-cli


m4vazquez:
Hello, I have this error on aws-cli execution:
+ aws --version
Traceback (most recent call last):
  File "/usr/local/bin/aws", line 19, in <module>
    import awscli.clidriver
  File "/usr/local/lib/python3.5/site-packages/awscli/clidriver.py", line 17, in <module>
    import botocore.session
  File "/usr/local/lib/python3.5/site-packages/botocore/session.py", line 29, in <module>
    import botocore.configloader
  File "/usr/local/lib/python3.5/site-packages/botocore/configloader.py", line 19, in <module>
    from botocore.compat import six
  File "/usr/local/lib/python3.5/site-packages/botocore/compat.py", line 26, in <module>
    from dateutil.tz import tzlocal
  File "/usr/local/lib/python3.5/site-packages/dateutil/__init__.py", line 5, in <module>
    from ._version import version as __version__
  File "/usr/local/lib/python3.5/site-packages/dateutil/_version.py", line 10
    version: str
           ^
SyntaxError: invalid syntax
----------------------------------------------

I fixed uninstalling 2.9.0 and install 2.8.2 and it's works but on the pip install aws-cli call any version <3.0.0 so I have to make changes on more that two hundred of piplelines.

mattmauriello:
we are having the same issues. another package install updates dateutil to 2.9.0 and we get the exact same stack trace.
running aws-cli/1.18.147 Python/2.7.18

we got it to work again with
pip2 install --force-reinstall -v 'python-dateutil==2.8.2'

seems to be simply because the _version.py file uses type checking which doesn't seem to be supported in python2.7, not until 3.6

tipabu:
Duplicate of https://github.com/dateutil/dateutil/issues/1344 -- looks like the short-term plan is to pin an older version of `setuptools_scm` (which was responsible for generating `_version.py`) and get another release out; longer term, see https://github.com/dateutil/dateutil/issues/1347

Until the short-term fix is out, rolling back to `python-dateutil==2.8.2` is about all downstream consumers *can* do.

pganssle:
Also, I think there's no benefit to upgrading to 2.9.0 for any Python versions other than 3.7 unless you are using the internal `zoneinfo` database (which you hopefully are not, or you are patching in a newer version anyway) — that would only come up on Windows for the most part.

m4vazquez:
[...]

Issue #1347, opened 2024-03-01
Remove setuptools_scm


pganssle:
For the 3.0.0 release, I think we should remove `setuptools_scm`. So far this hasn't really saved us much effort, and on multiple occasions it's caused trouble. The upsides of manually maintaining a `__version__` string (or equivalent):

1. Version information is in the repo itself, so a snapshot of any given version is self-contained.
2. Removes a build-time dependency
3. The code in the repo will match what is actually distributed more closely.

Downsides are:

1. We lose the nice property where each revision gets its own version string
2. We might forget to bump the version (might be something we can check with GA?)

I think `setuptools_scm` was also doing stuff about maintaining the `MANIFEST.in` that I didn't realize, and I'm not sure I'm entirely comfortable with.

mattmauriello:
@pganssle , thank you for looking in to this. So that we can make plans internally, when do you expect a patch version would be published? we have a fair amount of patching and deploy to do to work around this, but if you expect a patch available shortly, we may be able to wait for that.

pganssle:
@mattmauriello Like an hour? Maybe 2?

mattmauriello:
@pganssle , we see the new package version, and it's resolving our issues where we've tested so far. Really appreciate your work and the quick turn around here, it was a life saver.

wimglenn:
You can see my comment in the [packaging.python.org guide](https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1646696925) for several more downsides on a `__version__` string. What happened with python-dateutil 2.9.0 is a great case study on the sort of trouble which this convention can cause.

big +1 to removing `setuptools_scm`

tacaswell:
Even if you switch tooling, please do not stop providing `__version__` as we rely on it in in Matpltolib https://github.com/matplotlib/matplotlib/blob/bdbbb2dc50d9cf901e78a5bbcea8f458b757650d/lib/matplotlib/__init__.py#L244-L260.

pganssle:
> Even if you switch tooling, please do not stop providing `__version__` as we rely on it in in Matpltolib

Well, you could always just check to see if it's a version that has stopped providing it... oh, wait 😛

Anyway, to be clear, removing `__version__` is not on the table. I imagine it will be hard coded in some fashion.

PR #1349, opened 2024-03-01
Unblock setuptools_scm

alerque:
This should un-bork what #1346 broke for folks that are actually in current instead of EOL Python envirenments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,8 @@ requires = [
     "setuptools; python_version != '3.3'",
     "setuptools<40.0; python_version == '3.3'",
     "wheel",
-    "setuptools_scm<8.0"
+    "setuptools_scm; python_version >= '3.6'",
+    "setuptools_scm<8.0; python_version < '3.6'"
 ]
 build-backend = "setuptools.build_meta"
 


pyproject.toml
```diff
@@ -3,7 +3,8 @@ requires = [
     "setuptools; python_version != '3.3'",
     "setuptools<40.0; python_version == '3.3'",
     "wheel",
-    "setuptools_scm<8.0"
+    "setuptools_scm; python_version >= '3.0'",
+    "setuptools_scm<8.0; python_version < '3.0'"
```
edgarrmondragon, line 7:
```suggestion
    "setuptools_scm; python_version >= '3.6'",
    "setuptools_scm<8.0; python_version < '3.6'"
```

https://github.com/dateutil/dateutil/issues/1348

pganssle:
Unfortunately this doesn't work, because the issue is that we build the tarball and the wheel once, using some arbitrary Python version, targeting all Python versions.

If we wanted to continue using `setuptools_scm` while we're still supporting old Python versions, the right way to do it would be to detect the `setuptools_scm` version in `setup.py` and set a `version_file_template` in versions of `setuptools_scm` that support that (I believe `write_to` would also need to be changed to `version_file` in that case).

Unfortunately (and ironically), `setuptools_scm` doesn't expose a `__version__` attribute, and we can't pull in any dependencies on `importlib.metadata` or its backports.

Instead, for future releases I plan to remove `setuptools_scm` entirely (see #1347), and for now we can pin to a version that does what we're expecting.

tucked:
Couldn't we use `importlib_metadata` instead of the version template?
Or better yet.... drop support for EOL pythons (downstreams can continue using the older versions)?

pyproject.toml
edgarrmondragon: I see what @pganssle said in https://github.com/dateutil/dateutil/pull/1349#issuecomment-1974059666. He's correct that this won't work:

* If you build on Python >= 3.6, you get a wheel built with the latest `setuptools_scm`
* If you build on Python < 3.6, you get a wheel built with `setuptools_scm` < 8.0

You can only pick one, and in either situation you'll have unhappy users :)

alerque:
> Unfortunately this doesn't work, because the issue is that we build the tarball and the wheel once, using some arbitrary Python version, targeting all Python versions.

If the issue is what ends up in the wheel, then setup your wheel builder to use the version you want to use, don't block everybody else from building with current tooling. The solution in this PR should work for the source dist tarball for anybody building it.

> Instead, for future releases I plan to remove `setuptools_scm` entirely (see #1347), and for now we can pin to a version that does what we're expecting.

That's up to you. In general using vcs/scm versioning modules is a good thing for distros because it allows them to build devel packages from Git HEAD with sensible version data. It also usually gives an extra channel that works for builds (being able to use Git archive sources instead of source dist tarballs while still getting the expected version info). I package hundreds of Python libraries for Arch and usually having setuptools_scm (or equivalent for other Python build systems) is usually just fine. I don't think you necessary *need* to give up on those advantages.

pganssle:
If you would like to build and deploy something different from the standard version of `dateutil` that is your decision, but for now the official version does not include type hints in `_version.py`, and I'm not interested in introducing differences in the output depending on what version of Python is used to generate releases, which is what this PR does.

Thank you for your contribution.

tucked:
> building with current tooling.

If you build this PR with current tooling, you end up with a package that claims to support py2 but doesn't support py2.

alerque:
I'm not trying to "deploy" or "generate releases". I'm trying to build and install against system-provided dependencies. In other words anybody not operating on a per-project silo basis such as venvs or whatever is going to have trouble with this not being addressed.

PR #1388, opened 2024-09-07
parse TZID param for RDATE

### LINKED ISSUE #872: Add support for all the parameters of RDATE
This is related to #410 and #859, we currently only support a limited subset of possible values for `RDATE`, but we should support all the parameters in `RDATE`, the grammar for which is set out in [RFC 5545](https://tools.ietf.org/html/rfc5545#page-120):

```
       rdate      = "RDATE" rdtparam ":" rdtval *("," rdtval) CRLF

       rdtparam   = *(
                  ;
                  ; The following are OPTIONAL,
                  ; but MUST NOT occur more than once.
                  ;
                  (";" "VALUE" "=" ("DATE-TIME" / "DATE" / "PERIOD")) /
                  (";" tzidparam) /
                  ;
                  ; The following is OPTIONAL,
                  ; and MAY occur more than once.
                  ;
                  (";" other-param)
                  ;
                  )

       rdtval     = date-time / date / period
       ;Value MUST match value type
```

This is nearly identical to the `EXDATE` grammar, so you can start with a direct port of #859, the only difference is that `RDATE` should support a `PERIOD` value. I do not know exactly what that means and if it's something we even support in the `rrule` class itself, so I'd say start with the port of #859, and when that's done see how easy it will be to add `PERIOD` support and we'll take it from there.

jeriox:
## Summary of changes

parse TZID param for RDATE

Closes #872 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/src/dateutil/rrule.py
+++ b/src/dateutil/rrule.py
@@ -1675,10 +1675,11 @@ class _rrulestr(object):
                         raise ValueError("unsupported RRULE parm: "+parm)
                     rrulevals.append(value)
                 elif name == "RDATE":
-                    for parm in parms:
-                        if parm != "VALUE=DATE-TIME":
-                            raise ValueError("unsupported RDATE parm: "+parm)
-                    rdatevals.append(value)
+                    rdatevals.extend(
+                        self._parse_date_value(value, parms,
+                                               TZID_NAMES, ignoretz,
+                                               tzids, tzinfos)
+                    )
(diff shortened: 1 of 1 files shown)


jeriox:
@pganssle would be happy about a review and some hints on how to make the other checks happy

jeriox:
@pganssle did you have a chance to look at this?

PR #1406, opened 2025-01-03
Allow a missing dictionary to be returned from dateutil.parser::parse

### LINKED ISSUE #1405: No way to detect missing values in the date from dateutil.parser::parse
You can set a default value for the parts of the date, but if the default value is a valid option then it can't guarantee that value as missing. For example, there is no current way to determine if "January 2" has a year or not. If there is a way, please let me know, I just haven't seen anything that can expose it.

DRobinson4105:
The missing dictionary maps each component of the date/time string to a boolean indicating whether it was explicitly provided (False) or missing (True).

Resolves #1405

### DIFF
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -570,7 +570,7 @@ class parser(object):
         self.info = info or parserinfo()
 
     def parse(self, timestr, default=None,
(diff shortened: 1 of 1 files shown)


DRobinson4105:
@schinckel @jpmckinney @pquentin @nicoe The only failing check seems to be out of my control. If there is anything I need to change, please let me know.

PR #1417, opened 2025-02-24
Fix validation workflow

rowan-walsh:
## Summary of changes

- Rearrange test matrix: removed exclude section and moved special cases (older versions) as include entries.
- Switch to macos-13 runners for 3.5-3.7 since macos-12 is no longer available.
- Fix issue with latest codecov action not working on intel macos runners.
- Remove ubuntu-20.04 runners since these are being deprecated soon: https://github.com/actions/runner-images/issues/11101. Instead, test older python versions in python:x.y containers.

### Pull Request Checklist
- [x] Changes have tests (N/A: CI changes)

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -16,10 +16,6 @@ jobs:
     strategy:
       matrix:
         python-version: [
-          "2.7",
-          "3.5",
-          "3.6",
-          "3.7",
           "3.8",
           "3.9",
           "3.10",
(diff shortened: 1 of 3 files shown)


.github/workflows/validate.yml
```diff
@@ -104,6 +94,9 @@ jobs:
           file: ./.tox/coverage.xml
           name: ${{ matrix.os }}:${{ matrix.python-version }}
           fail_ci_if_error: false
+          # codecov/codecov-action@v3 needs v0.7.3 to work on (intel) macos-13
+          # See https://github.com/codecov/codecov-action/issues/1549
+          version: ${{ matrix.os == 'macos-13' && 'v0.7.3' || 'latest' }}
```
niccokunzmann, line 99: I do not understand this statement.
What does it do?

.github/workflows/validate.yml
```diff
@@ -30,49 +26,43 @@ jobs:
...
+          - os: "ubuntu-latest"
+            python-version: "2.7"
+            use-container: true
+          - os: "ubuntu-latest"
+            python-version: "3.5"
+            use-container: true
+          - os: "ubuntu-latest"
+            python-version: "3.6"
+            use-container: true
+          - os: "ubuntu-latest"
```
niccokunzmann, line 53: It is really nice that you update the workflow.
In production, dateutil is used with all  current versions.
Could you update the matrix to also check

3.8 - 3.13?

niccokunzmann (commented):
Hi, I am not a maintainer but I took some time to review. Thank you for the patch!

.github/workflows/validate.yml
hugovk: It's the equivalent of:

```python
version = matrix.os == "macos-13" and "v042" or "latest"
```
and:
```python
if matrix.os == 'macos-13':
   version = 'v0.7.3'
else:
   version = 'latest'
```

.github/workflows/validate.yml
rowan-walsh: I may be misunderstanding you, but python versions 3.8 - 3.13 are already included in the regular matrix definition (expanded against the 3 'latest' os's). Older python versions are listed explicitly because they need special handling.
You can see the full list of os/python versions that this expands to in the check results for this PR.

pganssle:
Wow, thanks for this! I had been putting off the inevitable bitrot cleanup until I was able to make my final charge at #1130; I am now even more inspired to get that final charge done very soon in order to finish before further bitrot sets in 😅

jjerphan:
Hi, this PR introduces https://github.com/dateutil/dateutil/issues/1419.

https://github.com/dateutil/dateutil/pull/1420 proposes a fix.

PR #1418, opened 2025-02-25
CRLF -> LF for 1417.misc.rst

holzman:
## Summary of changes

Change CRLF to LF from rst from changelog.  This is necessary to avoid the autoformat kicking in when doing
things like

```pip install git+https://github.com/dateutil/dateutil@9eaa5de584f9f374c6e4943069925cc53522ad61```

After making this change, the above seemed to work for me on both Linux and Windows.


holzman:
Closing since this is identical to #1420.

PR #1512, opened 2026-05-30
docs(parser): warn that non-datetime strings have undefined parse behaviour

### LINKED ISSUE #1270: Be more explicit in the documentation that passing non-datetime strings has undefined behavior
The only really sustainable contract for `dateutil.parser.parse` as it currently exists is that if you pass it a string representing a `datetime` it will parse it, and if you pass it some other string, it may parse to *something* or it may fail. This is *implicit* in [the documentation](https://dateutil.readthedocs.io/en/stable/parser.html#module-dateutil.parser), but we should probably put a `.. caution` or `.. note` note in there, because it's a pretty common misconception.

sridhar-3009:
## Summary

Fixes #1270.

`parser.parse()` makes no guarantees about strings that don't represent a date/time — it may return something, raise `ParserError`, or silently produce a wrong date. This was only implicit in the docs.

Add a `.. warning::` admonition directly on the `timestr` parameter so users see this at a glance.

### DIFF
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -577,6 +577,14 @@ class parser(object):
         :param timestr:
             Any date/time string using the supported formats.
 
+            .. warning::
+
+                Passing a string that does not represent a date or time has
+                **undefined behaviour** — the parser may return an arbitrary
+                result, raise :exc:`ParserError`, or silently produce an
+                incorrect date.  Only strings that actually represent a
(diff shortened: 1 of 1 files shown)


sridhar-3009:
Note: the  CI failure shows  as unreachable — this is a network-level issue in the CI environment (the link was already present before this PR). The GNU.org TZ Variable page is legitimately reachable externally; it just times out from the runner's network. This failure is unrelated to the docstring warning I added.

PR #1518, opened 2026-06-01
fix: fuzzy_with_tokens does not return unused numerical tokens

### LINKED ISSUE #1063: fuzzy_with_tokens does not return unused numerical tokens
Case 1: rel #326
If the `_parse_numeric_token` method exits without having used the token, the token does not get handled any further but also does not get added to the `skipped_idxs` list.

Case 2: rel #1062
When parsing a string with multiple `HH:MM` times, they get overwritten each time a new time is encountered, without the overwritten values being added to the token list. For example:
```pycon
>>> parse("from 13:45 to 17:15", fuzzy_with_tokens=True)
(datetime.datetime(2020, 6, 27, 17, 15), ('from ', ' to '))
```

While case 1 might be solved by fixing the underlying bug with how colons are handled, there might be more edge cases which could trigger this issue. I also do not think that case 2 would be the expected behavior. There should probably be a check whether a value has already been set [here](https://github.com/dateutil/dateutil/blob/fc9b1625ebc729f01e449879b6b140abd12ae621/dateutil/parser/_parser.py#L944), which could perhaps be an issue in other places as well.

YousafZahid1:
Closes #1063

## Summary of changes

This PR fixes an issue where `fuzzy_with_tokens` was not returning unused numerical tokens. The problem arose in two cases: when `_parse_numeric_token` exited without using a token, and when parsing strings with multiple times, where previous times were overwritten without being added to the token list.

### Pull Request Checklist
- [X] Changes have tests

Closes #1521

### DIFF
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -873,6 +873,12 @@ class parser(object):
             return res, None
 
     def _parse_numeric_token(self, tokens, idx, info, ymd, res, fuzzy):
+        if fuzzy:
+            # Store unused numerical tokens in skipped_tokens
+            skipped_tokens = getattr(self, 'skipped_tokens', None)
+            if skipped_tokens is None:
+                skipped_tokens = self.skipped_tokens = []
+            skipped_tokens.append(tokens[idx])
         # Token is a number
         value_repr = tokens[idx]
         try:


pganssle:
None of those check boxes seems accurate?

PR #1522, opened 2026-06-16
Keep comma as a separator before a time field in _timelex (fixes #1409)

### LINKED ISSUE #1409: ParserError when parsing a date like `2025-1-10,0:57:17.0,+1:0`
## A minimal, reproducible example
```python
import dateutil
date_str = "2025-1-10,0:57:17.0,+1:0"
dateutil.parser.parse(date_str)
>>> dateutil.parser._parser.ParserError: Unknown string format: 2025-1-10,0:57:17.0,+1:0
```

## Platform and versions:
dateutil: 2.9.0.post0
Python: 3.11.9

## Description
When parsing a date `2025-1-10,0:57:17.0,+1:0`, a ParserError is raised.
However, when using a similar format, but with only one digit for the day, the parsing succeeds:
```python
import dateutil
date_str = "2025-1-1,0:57:17.0,+1:0"  # Notice the date being 2025-1-1 as opposed to 2025-1-10 as in the previous example
dateutil.parser.parse(date_str)
>>> datetime.datetime(2025, 1, 1, 0, 57, 17, tzinfo=tzoffset(None, 3600))
```

I tracked down the issue to be at dateutil.parser._parser:143 being in `_timelex.get_token`, the line `elif nextchar == '.' or (nextchar == ',' and len(token) >= 2):` evaluates to True because token=10 and so the len is 2. I could not find why this if check exists in the first place.

If any extra information is required, do let me know.

gaoflow:
## Summary

`parser.parse("2025-1-10,0:57:17")` raises `ParserError: Unknown string format`, but the identical string with a one-digit day parses fine:

```python
>>> parse("2025-1-1,0:57:17")     # one-digit day
datetime.datetime(2025, 1, 1, 0, 57, 17)
>>> parse("2025-1-10,0:57:17")    # two-digit day
dateutil.parser._parser.ParserError: Unknown string format: 2025-1-10,0:57:17
```

Accepting the format with a one-digit day while rejecting it with a two-digit day is an internal inconsistency, not a deliberate stance on the format. Reported in #1409 (with a trailing UTC offset, which is also fixed).

## Root cause

In `_timelex` (`src/dateutil/parser/_parser.py`), a comma is treated as a decimal point whenever the number it extends already has two or more digits:

```python
elif nextchar == '.' or (nextchar == ',' and len(token) >= 2):
```

This exists to support the Python logging `HH:MM:SS,fff` format (the existing `"2003-09-25 10:49:41,502"` test case). But it fires too eagerly: in `2025-1-10,0:57:17` the day `10` already has two digits, so `10,0` is merged into the float `10.0` and the comma that actually separated the date from the time is swallowed, corrupting the token stream. With a one-digit day (`1,0`) the two-digit guard does not trip, which is why that case works.

## Fix

When a comma might extend a numeric token, peek past the digit run that follows it. If those digits lead into a `:`, the comma is separating a date from a following time field, so it is emitted as its own token rather than folded into the number. The logger fractional-seconds case is unaffected, because there the comma is followed only by digits and then end-of-token, never a `:`.

The peeked characters are pushed back onto the tokenizer's `charstack` unchanged, so tokenization proceeds exactly as if the lookahead had not happened.

## Tests

Added one case to `PARSER_TEST_CASES`, next to the existing logger-format case:

```python
("2025-1-10,0:57:17", datetime(2025, 1, 10, 0, 57, 17), "comma separator, two-digit day (GH#1409)"),
```

It fails on `main` (raises `ParserError`) and passes with this change. The full `tests/test_parser.py` suite is unchanged otherwise (231 passed, 14 xfailed), including the `"...41,502"` logger case, and I swept a range of comma/decimal edge inputs (`12:34:56,789`, trailing comma, `10,5:30`, etc.) for regressions.

I will add the `changelog.d/<this PR number>.bugfix.rst` entry once the PR number is assigned, per CONTRIBUTING.

Disclosure: I developed this fix with AI assistance, under my direction. I have reviewed and verified every line, the tokenizer push-back ordering, and the regression sweep myself, and I am happy to adjust the approach.

### DIFF
--- a/changelog.d/1522.bugfix.rst
+++ b/changelog.d/1522.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed a tokenizer bug where a comma separating a date from a time was merged
+into the preceding number as a decimal point when the number had two or more
+digits, so strings such as ``2025-1-10,0:57:17`` raised ``ParserError`` even
+though the same string with a one-digit day parsed. Reported by @shifqu
+(gh issue #1409). Fixed by @gaoflow (gh pr #1522).
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -140,9 +140,19 @@ class _timelex(object):
                 # numbers until we find something that doesn't fit.
                 if self.isnum(nextchar):
                     token += nextchar
-                elif nextchar == '.' or (nextchar == ',' and len(token) >= 2):
+                elif nextchar == '.':
                     token += nextchar
                     state = '0.'
+                elif (nextchar == ',' and len(token) >= 2 and
+                      not self._comma_is_separator()):
+                    token += nextchar
+                    state = '0.'
+                elif nextchar == ',' and len(token) >= 2:
+                    # ``_comma_is_separator`` already pushed the peeked chars
+                    # back onto the charstack; re-queue the comma in front of
+                    # them so it is tokenized as a standalone separator.
+                    self.charstack.insert(0, nextchar)
+                    break  # emit token
                 else:
                     self.charstack.append(nextchar)
                     break  # emit token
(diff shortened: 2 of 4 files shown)


binggao1230:
I pushed `9a53ddb` to address the CI failures on this PR.

Changes:
- Applied the `darker` / `isort` formatting required by the `darker` job.
- Updated the `test_ymd_hms_micro` parametrization to pass a tuple instead of a generator, which fixes pytest 9.x under `filterwarnings = error`.

Local checks run:
- `darker --check --diff --isort --revision origin/master .`
- `darker --check --diff --isort --revision origin/master...HEAD .`
- `PYTHONPATH=src pytest tests/test_isoparser.py tests/test_parser.py -q` (`799 passed, 15 xfailed`)
- `tox -e py -- tests/test_isoparser.py tests/test_parser.py` (`799 passed, 15 xfailed`)
- `git diff --check`

The new remote run is green across GitHub Actions, AppVeyor, and ReadTheDocs.

Issue #1519, opened 2026-06-01
Proposal: speed up parser.parse() (ISO-8601 fast-path + tokenizer hot-path) — gauging appetite before a PR


tfoutrein:
Hi — before opening a large PR I'd like to check your appetite and preferred shape. This is strictly *zero behavioral change* perf work.

For context: I'm doing this as part of a meetup talk on self-improvement harnesses — the thesis being that you can point one at a real, widely-used library and get genuine, conformance-preserving speedups. `dateutil` is the case study (a hot transitive dependency, and the per-element fallback in `pandas.to_datetime`), and I'd genuinely like the improvements to land, not just live on a slide.

I have a fork (branched from current `master`) with a few commits touching **only** `src/dateutil/parser/_parser.py`:

- a **fast-path for ISO-8601 inputs** at the top of `parse()` — today `parse()` never dispatches to `isoparse()` (that's a separate module since #489/#424), so well-formed ISO strings pay the full general-parser cost. A precompiled regex recognises complete ISO forms and builds the `datetime` directly; anything non-ISO or out-of-range falls back to the existing slow path unchanged;
- a **bulk-scan tokenizer** in `_timelex.split` (replacing the per-char `read(1)` loop);
- unrolled `_result.__init__` / `__len__`, an exception-free per-token `float()` guard, and a leaner numeric path.

No `:=` and no f-strings were introduced (kept py2.7-friendly on purpose).

## Correctness (the priority, given how widely the lib is used)

I ran the **full test suite on both `master` and the fork** with identical tests: the failure sets are **identical** (43 failures, all in `test_tz.py` / `test_tz_prop.py` — environmental, local tzdata I/O), **1988 passed**, **0 failures in `parser`/`isoparser`**, **0 introduced regressions**. I also ran a differential harness over **2546 inputs** (`parse(s)` datetime equality + exception parity) and the Hypothesis property tests (green).

## Performance (workload-dependent — both ends, honestly)

| Workload | Result | Note |
|---|---|---|
| Deliberately diversified corpus | **×1.64 (−39%)** | lower bound — over-represents rare manual formats that need the general parser |
| Representative ISO-dominant corpus (ETL / logs / pandas per-row) | **×6.42 (−84%)**, 534 → 83 ms | fast-path bypasses ~90% of the ISO inputs in that corpus |

[...]

PR #1523, opened 2026-06-18
Add a fast path for complete ISO-8601 timestamps in parser.parse()

tfoutrein:
## Summary

The vast majority of real-world inputs to `parser.parse()` are complete ISO-8601 timestamps (`YYYY-MM-DD[(T| )HH:MM[:SS[.ffffff]]][Z|+-HH[:]MM]`) — logs, ETL pipelines, `pandas.to_datetime`'s per-element fallback, etc. For those forms this adds a fast path at the top of `parse()` that builds the `datetime` directly from a single regex match, instead of running the full pipeline (the `_timelex` tokenizer + `_parse` + `resolve_ymd` + `_build_naive`/`_build_tzaware`).

This follows up on #1519 (where I asked whether perf contributions would be welcome).

## Strict equivalence

The fast path is deliberately conservative — it only engages when the result is provably identical to a full `parse()`:

- only the **default** `parserinfo` (`DEFAULTPARSER`) and **default** kwargs (`default=None`, `ignoretz=False`, `tzinfos=None`, no `fuzzy`/`fuzzy_with_tokens`, no `dayfirst`/`yearfirst`);
- a 4-digit year + 2-digit month/day makes YMD resolution unambiguous under the defaults;
- fraction truncated/padded to 6 digits exactly like `_parsems`; offset handling mirrors `validate()`/`_build_tzaware` (`Z`/offset 0 → `tz.UTC`, else `tz.tzoffset`);
- **anything else falls back to the slow path** — non-matching strings, and out-of-range components (month 13, day 32, year 0, ...) which raise inside the naive `datetime(...)` construction and are caught to defer to the slow path, so the **exact same `ParserError` (or `ValueError`) is raised**.

Pure Python, no new dependency; py2-safe (on Python 2 the `type(timestr) is str` guard simply declines, so behaviour there is unchanged).

## Benchmarks

`parse()` median, drift-immune A/B vs current `master`, by input shape:

| input | speedup |
|---|---|
| ISO-8601 + timezone | ~8× |
| ISO-8601 (date/time) | ~5× |
| US-style / verbose / numeric (slow path) | ~1.0× (unchanged) |

The gain is concentrated on ISO inputs; non-ISO formats are untouched.

## Correctness

- Full `test_parser.py` / `test_isoparser.py` / property suites pass.
- Added parametrized tests covering the fast-path forms (date-only, time, fractional seconds, `Z`, `±HH:MM`, `±HHMM`, space separator) and the out-of-range fall-back.
- Verified iso-functional against the pre-change parser by a differential over a 2546-entry corpus **and** an adversarial fuzz of ~6800 generated ISO / near-ISO / malformed strings — 0 divergence in result or exception (type + message).

---

**Note (CI):** this branch also includes a one-line fix to `tests/test_isoparser.py` (a generator expression passed to `@pytest.mark.parametrize` → list comprehension). Under pytest 9.1 that generator raises `PytestRemovedIn10Warning`, which `filterwarnings = error` turns into a collection error, currently failing the whole test job (and, via the matrix fail-fast, the rest of CI) on every PR. Happy to split it into its own PR if you prefer.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -113,6 +113,7 @@ switch, and thus all their contributions are dual-licensed.
 - Thierry Bastian <thierryb@MASKED>
 - Thomas A Caswell <tcaswell@MASKED> (gh: @tacaswell) **R**
 - Thomas Achtemichuk <tom@MASKED>
+- Thomas Foutrein (gh: @tfoutrein) **D**
 - Thomas Grainger <tagrain@gmail.com> (gh: @graingert) **D**
 - Thomas Kluyver <takowl@MASKED> (gh: @takluyver)
 - Tim Gates <tim.gates@iress.com> (gh: timgates42)
--- a/changelog.d/1523.feature.rst
+++ b/changelog.d/1523.feature.rst
@@ -1,0 +1,1 @@
+Added a fast path for complete ISO-8601 timestamps in ``parser.parse()``: for those inputs the ``datetime`` is built directly (roughly 5-8x faster), while every other form falls back to the full parser with identical results. (gh pr #1523)
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -114,8 +114,11 @@ def _try_iso_fastpath(timestr):
         second = int(ss) if ss is not None else 0
         frac = m.group(7)
         if frac is not None:
-            # Same as _parsems: pad/truncate to 6 digits.
-            microsecond = int(frac.ljust(6, "0")[:6])
+            # Same as _parsems: pad/truncate to 6 digits. The fill char is
+            # str("0"), not "0": under Python 2.7 the fast path can receive a
+            # bytes timestr, and bytes.ljust rejects the unicode "0" produced
+            # by unicode_literals. str("0") is a native str on both 2 and 3.
+            microsecond = int(frac.ljust(6, str("0"))[:6])
         else:
             microsecond = 0
 
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -52,6 +52,109 @@ from .. import tz
 __all__ = ["parse", "parserinfo", "ParserError"]
 
 
+# Fast path for fully-specified ISO-8601 timestamps at the top of
+# parser.parse(). The vast majority of real-world inputs are complete ISO
+# dates (YYYY-MM-DD[(T| )HH:MM[:SS[.ffffff]]][Z|+-HH[:]MM]); for those forms we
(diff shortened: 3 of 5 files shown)


tfoutrein:
A heads-up on the CI status for this PR.

On top of the fast path, the branch carries two small CI-enablers:

- **`tests/test_isoparser.py`**: a generator expression passed to `@pytest.mark.parametrize` is converted to a list comprehension. Under pytest 9.1 the generator raises `PytestRemovedIn10Warning`, which `filterwarnings = error` promotes to a *collection* error — that currently fails the whole test job (and, via the matrix `fail-fast`, cancels the rest of CI) on every PR. With this change collection succeeds and the suite runs (2048 tests).
- A Python 2.7 fix for the new fast path: it can receive a `bytes` `timestr`, where `str.ljust(6, "0")` rejects the unicode `"0"` fill char coming from `unicode_literals` (`TypeError: ljust() argument 2 must be char, not unicode`). The fill char is now `str("0")`, a native `str` on both 2 and 3 — same result, no behaviour change.

After those, the two remaining red checks appear pre-existing / unrelated to this change:

- **`other (docs)`** — the Sphinx linkcheck fails on `https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html` with `Network is unreachable`. This also fails on master (e.g. run `26113946184`), so it looks like an external linkcheck/infra issue rather than something introduced here.
- **`test (3.8, ubuntu-latest)`** — `tests/property/test_tz_prop.py::test_gettz_returns_local` fails with `assert 'LMT' == 'EST'` on a 1901 datetime. That's a tzdata-sensitive Hypothesis property test on `tz.gettz`, unrelated to the parser fast path.

Happy to split the two CI-enablers into a separate PR if you'd prefer them apart from the feature.

tfoutrein:
One more on CI, to complete the picture (my earlier note only covered the two GitHub Actions checks): the AppVeyor check is also a pre-existing/transient infra failure, not from this change.

It's a single job (`PYTHON_VERSION=34`, x64) that fails during **dependency installation**, before any test runs:

```
ERROR: Could not find a version that satisfies the requirement coverage (from -r requirements-dev.txt) (from versions: none)
ERROR: No matching distribution found for coverage
```

`(from versions: none)` means pip found *no* versions of `coverage` at all (not "none compatible with 3.4") — i.e. the package index was momentarily unreachable, a transient hiccup rather than a code or dependency problem. The same Python 3.4 job passed on the two immediately preceding pushes of this PR (the changelog fragment and the pytest-9.1 collection fix); only the most recent push hit it. This diff touches neither `requirements-dev.txt` nor the install step, so a re-run should clear it.

PR #1528, opened 2026-06-30
fix: handle space after month abbreviation dot (e.g. Jan. 5.2014)

Someshwar-prox:
## Summary

Fixes issue where dates like `Jan. 5.2014` (space after month abbreviation dot) fail with `Unknown string format`.

## Root Cause

The tokenizer produces `5.2014` as a single numeric token when there is a space between the month abbreviation dot and the number. The parser then fails in the YYMMDD/HHMMSS branch because `int(5.)` raises ValueError.

## Fix

Added handling in `_parse_numeric_token` for numeric tokens in `day.year` format. These are now split into separate day and year components.

## Testing

- All 231 existing tests pass
- Verified: `Jan. 5.2014`, `Jan.5.2014`, `Jan 5 2014`, `5 Jan 2014` all parse correctly

### DIFF
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -904,6 +904,12 @@ class parser(object):
                 ymd.append(s[:2])
                 ymd.append(s[2:4])
                 ymd.append(s[4:])
+            elif '.' in s and len_li == 6 and s.count('.') == 1:
+                # day.year (e.g., "5.2014")
+                part_before_dot, part_after_dot = s.split('.')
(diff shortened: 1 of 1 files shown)


Someshwar-prox:
Fixed the issue where Jan. 5.2014 (with space after month abbreviation dot) was failing to parse.

Changes:
- Added handling in _parse_numeric_token for numeric tokens in day.year format
- Splits tokens like 5.2014 into separate day and year components
- All 231 existing tests pass with no regressions

PR #1540, opened 2026-07-23
Make relativedelta.__hash__ consistent with __eq__ for weekday

Osamaali313:
### Summary

`relativedelta.__eq__` treats a weekday whose `n` is in `(None, 0, 1)` as equal (`MO == MO(1)`, since the docstring notes *"Not specifying it is the same as specifying +1"*), but `__hash__` hashed the raw `weekday` object. So two `relativedelta`s that compare equal can hash differently, which violates the equal-objects-equal-hash invariant and makes set/dict membership silently wrong:

```python
from dateutil.relativedelta import relativedelta, MO

a = relativedelta(weekday=MO)
b = relativedelta(weekday=MO(1))

a == b              # True
hash(a) == hash(b)  # False  <- bug
b in {a}            # False  <- bug
```

### Fix

Normalize the `(None, 0, 1)` weekday equivalence class in `__hash__` so equal objects hash equally. Weekdays with other explicit counts (e.g. `MO(2)`, `MO(-1)`) keep the full `weekday` object, so distinct values still hash distinctly. Added a regression test (`testHashableWeekdayEquivalence`).

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -90,6 +90,7 @@ switch, and thus all their contributions are dual-licensed.
 - Nicolas Évrard (gh: @nicoe) **D**
 - Nick Smith <nick.smith@MASKED>
 - Orson Adams <orson.network@MASKED> (gh: @parsethis) **D**
+- Osamaali313 (gh: @Osamaali313)
 - Paul Brown (gh: @pawl) **D**
 - Paul Dickson (gh @prdickson) **D**
 - Paul Ganssle <paul@ganssle.io> (gh: @pganssle) **R**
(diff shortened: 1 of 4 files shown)


Osamaali313:
Heads-up on the red CI here: it's not this change. `master` itself is currently failing at test **collection** because `tests/test_isoparser.py::test_ymd_hms_micro` passes a generator expression to `parametrize`, which pytest 8.4 turns into an error under the repo's `filterwarnings = error`, aborting the whole `tests/` run before it reaches `test_relativedelta.py`.

I opened #1558 to fix that (generator → list). Once it lands I'll rebase this PR and CI should go green. The `__hash__`/`__eq__` change here is unrelated and unaffected.

PR #1542, opened 2026-07-29
fix: validate timezone offset components in parser.parse()

### LINKED ISSUE #1508: parser.parse() silently accepts invalid timezone offsets (minutes >= 60, hours >= 24); inconsistent with isoparse()
### Description

`dateutil.parser.parse()` silently accepts timezone offsets where the **minutes component is >= 60** and where the **total offset is >= 24 hours**, producing datetimes that either have incorrect offsets or are "toxic" (parse OK but every subsequent method call throws `ValueError`).

This is inconsistent with `dateutil.parser.isoparse()` (and Python stdlib `datetime.fromisoformat()`), which correctly reject these.

### Reproducer (python-dateutil 2.9.0.post0, latest on PyPI)

```python
from dateutil.parser import parse, isoparse

# 1. Silent timezone spoof — minutes >= 60 wraps into the hour
dt = parse("2024-01-15T12:00:00+0060")
print(dt.isoformat())       # 2024-01-15T12:00:00+01:00  ← wrong: input said +00:60
isoparse("2024-01-15T12:00:00+00:60")
# ValueError: Invalid minutes in time zone offset  ← isoparse correctly rejects

# More examples — parser silently normalizes:
parse("2024-01-15T12:00:00+0560")  # → +06:00  (off by 1 hour from "+05:??")
parse("2024-01-15T12:00:00+0099")  # → +01:39

# 2. Toxic datetime: parses OK but is unusable
dt = parse("2024-01-15T12:00:00+2400")     # no error here
print(dt)                                  # ValueError: offset must be strictly between -24h and +24h
# Same for: dt.timestamp(), dt.isoformat(), dt.utcoffset(), ...

# 3. Wrong exception type
from dateutil.parser import ParserError
try:
    parse("2024-01-15T12:00:00+9900").timestamp()
except ParserError:
    pass        # never reached
except ValueError:
    pass        # actually caught here — bypasses ParserError-only handlers
```

### Root cause

`dateutil/parser/_parser.py` extracts the offset like:

```python
if len_li == 4:                                   # e.g. "0060"
    hour_offset = int(l[i + 1][:2])
    min_offset  = int(l[i + 1][2:])
res.tzoffset = signal * (hour_offset * 3600 + min_offset * 60)
```

There is no check that `0 <= min_offset < 60` or that `0 <= hour_offset < 24`. The corresponding code in `dateutil/parser/isoparser.py` does perform these checks:

```python
if minutes > 59:
    raise ValueError("Invalid minutes in time zone offset")
```

so the bug is local to the non-ISO parser.

### Why this matters

- Applications using `parse()` to normalize user-supplied timestamps end up with the wrong UTC offset, silently. A scheduler or expiration check using these timestamps can be off by an hour (or more).
- Inputs in the `>= 24h` range produce datetime objects whose every method call raises — a downstream crash far from the parse site, when the exception handler is set up to catch `ParserError`.
- `parse()` and `isoparse()` disagree on the same input, which is a footgun.

### Suggested fix

In `_parser.py`, after computing `hour_offset` / `min_offset`, validate before assembling:

```python
if not (0 <= min_offset < 60):
    raise ValueError(f"Invalid minute offset {min_offset} in timezone")
if not (0 <= hour_offset < 24):
    raise ValueError(f"Invalid hour offset {hour_offset} in timezone")
```

(and ideally re-wrap as `ParserError` to keep the documented API contract).

### Environment

- python-dateutil: 2.9.0.post0
- Python: 3.9+

cyforkk:
## Description

parser.parse() silently accepted invalid timezone offsets where
minutes >= 60 or hours >= 24, producing incorrect or 'toxic'
datetime objects. This fix adds validation to match isoparse()
behavior.

## Changes
- Added validation in  to check:
  - 
  - 
- Raises  for invalid offsets

Fixes dateutil/dateutil#1508

### DIFF
--- a/src/dateutil/parser/_parser.py
+++ b/src/dateutil/parser/_parser.py
@@ -829,6 +829,20 @@ class parser(object):
                     else:
                         raise ValueError(timestr)
 
+                    # Validate timezone offset components to prevent
+                    # silent acceptance of invalid offsets (e.g. minutes >= 60
+                    # or hours >= 24), matching isoparse() behavior.
+                    if not (0 <= min_offset < 60):
+                        raise ValueError(
+                            f"minute offset in timezone must be in 0..59, "
+                            f"got {min_offset}"
+                        )
+                    if not (0 <= hour_offset < 24):
+                        raise ValueError(
+                            f"hour offset in timezone must be in 0..23, "
+                            f"got {hour_offset}"
+                        )
+
                     res.tzoffset = signal * (hour_offset * 3600 + min_offset * 60)
 
                     # Look for a timezone name between parenthesis
--- a/tests/test_isoparser.py
+++ b/tests/test_isoparser.py
@@ -120,8 +120,8 @@ def test_ymd_hms(dt, date_fmt, time_fmt, tzoffset):
 DATETIMES = [datetime(2017, 11, 27, 6, 14, 30, 123456)]
 @pytest.mark.parametrize('dt', tuple(DATETIMES))
 @pytest.mark.parametrize('date_fmt', YMD_FMTS)
-@pytest.mark.parametrize('time_fmt', (x + sep + '%f' for x in HMS_FMTS
-                                      for sep in '.,'))
+@pytest.mark.parametrize('time_fmt', [x + sep + '%f' for x in HMS_FMTS
+                                      for sep in '.,'])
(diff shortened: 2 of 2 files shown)


cyforkk:
Closing this PR because the current implementation breaks existing tests. The validation logic is too strict and causes many test failures. A different approach is needed that doesn't break backward compatibility.

